PR #2, opened 2026-07-06
Add per-agent behavior packs: an opt-in per-agent behavior layer

alexrao2000:
## The point of this PR: a per-agent, opt-in behavior layer

AgentOS runs every agent on one shared runtime (the runner + worker below the ACI port). That is right for genuinely universal concerns (kill switch, tracing, sandboxing), but it leaves **opinionated, per-agent behavior with no home**: a touch like the Slack placeholder is a single global config, and there is no per-agent hook at all. So an owner who wants some behavior for *their* agent can only force it on *every* agent on the install, or go without.

This PR adds the missing layer: **behavior packs** — a per-agent, opt-in, declarative config an owner enables for their own deployments, resolved at bind time, with no effect on anyone else's agent. It is designed to grow: new pack types slot into the same mechanism.

**The example packs are not the point.** `tips`, `greeting`, and `help` are three seed pack types included to exercise the mechanism end to end (one cosmetic, two deterministic/pre-model). The reviewable substance is the packs mechanism — storage, per-agent resolution, and the declarative-data safety model.

## We mapped the whole template battery set

Per the request to cover *every* all-agent feature from the templates: `docs/behavior-packs.md` now contains a table mapping all ~15 batteries to this mechanism. The key finding is that **most are not packs** — and forcing them to be would break the design:

- **Packs** (declarative data): tips, greeting, help. (`runtime-settings` schema is a candidate, deferred.)
- **Already native to AgentOS**: kill switch (`killswitch.py` + kernel gate + control endpoint), logging, env config, activity/observability (Langfuse). Re-adding would duplicate platform features.
- **Code, not data** (would run battery code in the worker, breaking isolation): http-backoff, vision, polling-worker → belong in the runner/plugin; integrations → the plugin's `.mcp.json`.
- **Need a Block Kit reply model AgentOS doesn't have**: nav (back/up), Block Kit rendering.
- **Domain logic, not a battery**: proactive notifier.

The throughline: the moment a battery needs to run agent-authored logic, it stops being a pack and must live in the sandboxed runner (or already lives in the platform). Packs cover only the declarative, deterministic slice — which is what keeps them safe to resolve and apply outside the sandbox.

## Design: declarative data, not code

A pack is JSON on the agent's row. The platform owns the logic that consumes it; only the data varies. Pack content never executes, so enabling a pack can never run an agent's code and the sandbox-isolation guarantee holds without any pack crossing into the runner.

## Scope (substrate only)

- **apps/api** — nullable `agents.behavior_packs` JSONB column (migration `0005`), `BehaviorPacksConfig` schema, accepted on `POST /agents`, read/written via `GET|PUT /agents/{id}/behavior-packs` (mirrors the budget control endpoints). NULL reads as all-off.
- **apps/worker/behaviorpacks.py** — `sample_tip`, `match_greeting`, `match_help`; pure stdlib, fully unit-tested. The matchers share one bare-utterance core; "hi show me the report" falls through to the model.
- **apps/worker/binding.py** (not the sacred kernel) — resolver selects `behavior_packs`, carries it on `ResolvedDeployment`, exposes `packs_for()`.

## Deliberately NOT in this PR: the kernel wiring

The packs fire from inside `apps/worker/kernel.py`, the F1 "sacred" module — any change there needs the escalated adversarial review per `apps/worker/CLAUDE.md`, and a greeting/help short-circuit brushes against kernel rule 3 ("never keyword-guess intent"). The exact integration points are written up in `docs/behavior-packs.md`.

## Note: this branch also carries the CI fix (#3)

The first commit here is the standalone CI fix also proposed as #3 (the `Start dev stack` step pulled a private image and failed before pytest). It is included so this PR's Python check can actually run and go green. Merge #3 first and rebase, or merge this PR directly — either way the change is identical.

## Verification

Against the real compose Postgres + MinIO (not mocks): `ruff check` clean, `mypy` clean (93 files), and **38 pytest passed** — the pack unit tests (incl. help + shared-matcher), the API round-trip, `crud`, `openapi_drift`, `migrations`, and the worker binding resolver.

Ref CUR-1586

### DIFF
--- a/apps/api/alembic/versions/0005_behavior_packs.py
+++ b/apps/api/alembic/versions/0005_behavior_packs.py
@@ -1,0 +1,31 @@
+"""add per-agent behavior_packs column
+
+Revision ID: 0005
+Revises: 0004
+Create Date: 2026-07-06
+"""
+
+from collections.abc import Sequence
+
+import sqlalchemy as sa
+from alembic import op
+from sqlalchemy.dialects import postgresql
+
+revision: str = "0005"
+down_revision: str | None = "0004"
+branch_labels: str | Sequence[str] | None = None
+depends_on: str | Sequence[str] | None = None
+
+SCHEMA = "agentos"
+
+
+def upgrade() -> None:
+    op.add_column(
+        "agents",
+        sa.Column("behavior_packs", postgresql.JSONB(), nullable=True),
+        schema=SCHEMA,
+    )
+
+
+def downgrade() -> None:
+    op.drop_column("agents", "behavior_packs", schema=SCHEMA)
--- a/apps/api/openapi.json
+++ b/apps/api/openapi.json
@@ -3,6 +3,16 @@
     "schemas": {
       "AgentCreate": {
         "properties": {
+          "behavior_packs": {
+            "anyOf": [
+              {
+                "$ref": "#/components/schemas/BehaviorPacksConfig"
+              },
+              {
+                "type": "null"
+              }
+            ]
+          },
           "name": {
             "title": "Name",
             "type": "string"
@@ -32,6 +42,18 @@
       },
       "AgentOut": {
         "properties": {
+          "behavior_packs": {
+            "anyOf": [
+              {
+                "additionalProperties": true,
+                "type": "object"
+              },
+              {
+                "type": "null"
+              }
+            ],
+            "title": "Behavior Packs"
+          },
           "created_at": {
             "format": "date-time",
             "title": "Created At",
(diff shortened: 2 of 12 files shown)


alexrao2000:
The red **Python** check here is a pre-existing CI/infra bug, not this diff: the `Start dev stack` step's bare `docker compose up -d` pulls the private `ghcr.io/curie-eng/agentos-api` image and fails `unauthorized` before pytest runs (it hits every PR and main). My tests never execute. Fixed independently in #3; once that merges, re-running this check should let the Python job run for real. Ref CUR-1586.

TheConnMan (approved):
LGTM! This is half of a feature I had in mind (I've linked the issues). Thanks for implementing!

PR #174, opened 2026-07-09
Generate compose.release.yaml from compose.dev.yaml at publish time

### LINKED ISSUE #88: Generate compose.release.yaml from compose.dev.yaml at publish time
## Problem

`compose.release.yaml` (added in #87 / #41) is a hand-maintained, self-contained copy of `compose.dev.yaml`. It duplicates the whole stack and only differs by three transforms: the worker uses the published `agentos-worker-local` overlay image (not a `build:`), the otel collector config is inlined via a compose `configs:` block, and the `agentos-*` image tags are pinned to the release version at publish time.

Keeping a full duplicate means every `compose.dev.yaml` change must be mirrored by hand or the release stack silently drifts. This already bit once: #42 added the console UI service to `compose.dev.yaml` while #41 was in flight, and the release copy had to be updated to match during rebase.

## Change

Delete the committed `compose.release.yaml` and instead **generate** it from `compose.dev.yaml` in the release workflow's publish step, applying the three transforms programmatically:
1. Replace the `agentos-worker` `build:` block with `image: ghcr.io/curie-eng/agentos-worker-local:<version>`.
2. Inline `otel/collector-config.yaml` into a compose `configs:` block (with the `$$` escaping for the collector's `${env:...}`) and drop the host mount.
3. Pin every `ghcr.io/curie-eng/agentos-*:latest` tag to the release version (already done via sed on the committed copy).

The resolver keeps fetching the published `compose.release.yaml` release asset unchanged — only the source of that asset moves from a committed file to a generated one. One source of truth (`compose.dev.yaml`), no drift.

## Why

The current committed-duplicate approach is the "sum of lanes is not the product" fragility realized: two files that must stay in lockstep but nothing enforces it. Generation makes `compose.dev.yaml` the single source and the release variant a mechanical derivation.

TheConnMan:
## What

`compose.release.yaml` was a hand-maintained, self-contained copy of `compose.dev.yaml` that differed only by three mechanical transforms, so it silently drifted (it had already lost the ollama local-model services and the worker's local-model env vars). This makes `compose.dev.yaml` the single source of truth and derives the release variant at publish time.

- **`compose/generate_release_compose.py`** applies the three transforms to `compose.dev.yaml`:
  1. worker `build:` overlay -> pinned `ghcr.io/curie-eng/agentos-worker-local:<version>` image
  2. `otel/collector-config.yaml` inlined as a top-level `configs:` block (with `${env:...}` escaped to `$${env:...}`), dropping the host mount
  3. every `ghcr.io/curie-eng/agentos-*:latest` tag pinned to the release version

  It preserves comments and YAML anchors (text transforms, not parse-and-redump) and raises loudly if an expected anchor is missing.
- Deletes the committed `compose.release.yaml` duplicate.
- Release workflow generates and stages `dist/compose.release.yaml` (same uploaded release asset, new source); the CI `compose` job generates the file before its `docker compose config` validations.
- Drops the Rust test that read the committed file; its core/full invariant now holds via the generator's pytest and CI's docker-compose service-count checks.

The resolver / release-asset fetch path is unchanged.

Closes #88


compose/generate_release_compose.py
TheConnMan, line 1: Are there utilities which do this instead of maintaining it ourselves?

PR #178, opened 2026-07-09
Convert dispatcher + worker config to pydantic-settings

### LINKED ISSUE #74: Refactor: dispatcher and worker hand-roll config parsing where pydantic-settings is the house pattern
**Impact: medium · Effort: S per service.**

**Evidence:** `apps/api/src/agentos_api/config.py:10-13` uses `pydantic_settings.BaseSettings`, but the dispatcher and worker reimplement env->typed-config by hand: `apps/dispatcher/src/agentos_dispatcher/config.py:82-100` (`_set_str/_int/_float` + a manual `from_env` at `:61-79`) and `apps/worker/src/agentos_worker/config.py:188-200` (`_s/_i/_b` + a 30-line `from_env` at `:156-186`).

**Fix:** convert dispatcher + worker to `BaseSettings`. `pydantic-settings>=2.7` is already declared by `agentos-api` and present in `uv.lock`, so it is in the workspace lock; dispatcher/worker add it to their own `pyproject.toml`. **Scope note (leave as-is):** `packages/aci-protocol/.../session.py` and `runner/.../config.py` also hand-roll `from_env`, but `aci-protocol` deliberately depends on `pydantic` only (frozen-contract footprint) and the runner round-trips via a bespoke `to_env` — do not convert those.

**Success criterion:** dispatcher/worker config modules are `BaseSettings` subclasses with no `_set_*`/`_s`/`_i`/`_b`/`from_env` helpers; tests still green.

alexrao2000:
Closes #74. The dispatcher and worker hand-rolled env→typed-config (`_set_*`/`_s`/`_i`/`_b` + manual `from_env`) where `apps/api` already uses `pydantic-settings`. Both are now `BaseSettings` subclasses with the helpers removed.

## Behavior-preserving
- **Every env var name + default unchanged.** Plain-name fields (`VALKEY_*`, `DATABASE_URL`, `S3_*`, `SLACK_*`, `LANGFUSE_*`) via case-insensitive name matching; `AGENTOS_*` fields via explicit `validation_alias`. Reconstructed the old `from_env` mapping and confirmed no rename/drop/default-change.
- **Bool parsing replicated exactly, per service** (dispatcher truthy set includes `on`; worker does not) via a `BeforeValidator`, with real-bool kwargs still passing through.
- Numeric `Field(gt=…)` constraints, the `consumer_name` default factory, and the `valkey_socket_timeout_s` derived invariant preserved.

## One subtlety handled
`populate_by_name=True` is needed for field-name kwarg construction in tests — but in pydantic-settings it also makes the env source read the bare uppercased field name as a fallback for aliased fields (so `API_KEY`/`CREDENTIALS`/`MODEL` would newly be read). Closed that with a small `_AliasOnlyEnvSource(EnvSettingsSource)` (via `settings_customise_sources`) that drops the field-name key for aliased fields while leaving kwarg population intact. Regression tests assert a stray `API_KEY`/`CREDENTIALS`/`STREAM` is ignored while the `AGENTOS_*` alias is read.

## Scope
Only dispatcher + worker config (+ their `run.py` call sites and config tests). `packages/aci-protocol/…/session.py` and `runner/…/config.py` deliberately left as-is. No sacred-kernel change.

## Tests
New `apps/worker/tests/test_config.py` + `apps/dispatcher/tests/test_config.py` (alias-only env behavior, kwarg construction, plain-name fields). Existing config/socket-timeout tests pass. `ruff` + `mypy` clean. (10 pre-existing infra failures — unmigrated DB / no API on :8000 — unrelated, verified on base.)

### DIFF
--- a/apps/dispatcher/CLAUDE.md
+++ b/apps/dispatcher/CLAUDE.md
@@ -35,7 +35,7 @@ summary.
 
 ## Config surface
 
-`DispatcherConfig.from_env(os.environ)` parses `SLACK_APP_TOKEN`,
+`DispatcherConfig()` (a `pydantic_settings.BaseSettings`) reads `SLACK_APP_TOKEN`,
 `SLACK_BOT_TOKEN`, `VALKEY_*`, `AGENTOS_STREAM` (must match the worker's
 stream name), `AGENTOS_DEDUPE_PREFIX`/`_TTL_SECONDS`, `AGENTOS_PLACEHOLDER_TEXT`,
 and the backoff tunables. Full table in `apps/dispatcher/README.md`.
--- a/apps/dispatcher/README.md
+++ b/apps/dispatcher/README.md
@@ -57,7 +57,7 @@ connection and exits the loop without reconnecting. The Socket Mode adapter
 
 ## Config surface (env vars)
 
-Parsed by `DispatcherConfig.from_env(os.environ)`.
+Read from the environment by `DispatcherConfig()` (a `pydantic_settings.BaseSettings`).
 
 | env var | default | meaning |
 |---|---|---|
--- a/apps/dispatcher/pyproject.toml
+++ b/apps/dispatcher/pyproject.toml
@@ -7,6 +7,7 @@ dependencies = [
     "slack-bolt>=1.21",
     "redis>=5.2",
     "pydantic>=2.10",
+    "pydantic-settings>=2.7",
 ]
 
 [build-system]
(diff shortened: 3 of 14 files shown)


TheConnMan (commented):
Two things to nail down before merge, since config drift is a silent prod break: the worker's env-var surface appears to have widened in the conversion (double-check every var name/default matches the pre-pydantic reads exactly), and the per-service bool divergence is currently untested. A couple of parity tests on defaults/overrides would close it.

alexrao2000:
> Two things to nail down before merge, since config drift is a silent prod break: the worker's env-var surface appears to have widened in the conversion (double-check every var name/default matches the pre-pydantic reads exactly), and the per-service bool divergence is currently untested. A couple of parity tests on defaults/overrides would close it.

Addressed, preparing to merge

PR #179, opened 2026-07-09
Runner: fast-fail on a rejected model credential

### LINKED ISSUE #121: runner hangs ~2min on a rejected model credential instead of failing fast (auth errors are retried)
A bad/placeholder OpenRouter credential (e.g. the literal `sk-or-...`) makes `skill message` "think" for ~2 minutes and then time out, with no clear error. Root cause: the credential is routed to OpenRouter (`sk-or-` prefix matches), OpenRouter rejects it (401), and the SDK/runner retries with backoff until the wall instead of surfacing the auth failure immediately.

Repro: `AGENTOS_CREDENTIALS=sk-or-... agentos skill up --model z-ai/glm-5.2` then `agentos skill message "..."`. Runner env showed `AGENTOS_CREDENTIALS` = the literal 9-char placeholder; the turn hung ~2min then timed out. A real ~73-char `sk-or-v1-` key returns in seconds.

**Fix (either/both):**
- Do NOT retry provider auth failures (HTTP 401/403) -- surface them immediately as a clear "model credential rejected by provider" error, distinct from a generic timeout.
- Optionally validate the credential shape at `skill/local/cluster up` (a `sk-or-` value far shorter than a real key, or the literal placeholder) and fail fast with a hint, rather than baking a doomed key into the runner.

This is a top onboarding footgun: the failure mode (silent 2min hang) gives no signal that the key is the problem. Surfaced during the v0.2.0 no-clone walkthrough (a runbook placeholder was pasted verbatim).

alexrao2000:
Closes #121. A bad/placeholder model credential made a turn hang ~2 min then time out with no clear error — a top onboarding footgun (a runbook `sk-or-...` placeholder pasted verbatim). Root cause: the provider returns 401, and the SDK/CLI retries with backoff to the wall while the runner waited for a terminal result.

## Fix (runner only)
A provider auth rejection surfaces via the SDK as `AssistantMessage.error == "authentication_failed"` (not a Python exception). At the top of `_drive_turn`'s receive loop, detect it, `interrupt()` the session to stop the retry loop, and emit a **distinct terminal** `ErrorEvent(classification="model-credential-rejected")` + `Final(CLASSIFIED_FAILURE)` with a clear `"model credential rejected by provider (check AGENTOS_CREDENTIALS)"` message.

- **Non-retryable downstream:** the worker's `RETRYABLE_CLASSIFICATIONS` (`rate-limit`, `runner-error`) excludes `model-credential-rejected`, so it escalates to a human instead of retrying — no more ~2 min hang.
- **Transient errors unchanged:** `rate_limit`/`server_error` and raised `ClaudeSDKError`/`ProcessError` keep their existing (retryable) paths — the fast-fail is auth-only.
- **Not billed, no leak:** returns before budget accounting; the message/log carry only a hint, never the credential value.

## Scope
Only `runner/src/agentos_runner/session.py` (+ tests). The optional CLI credential-shape validation is out of scope (Rust lane). No sacred-kernel change.

## Tests
`test_auth_rejection_fails_fast_not_retried` (auth message → exactly `[error, final]`, `model-credential-rejected`, session interrupted, later messages never consumed) and `test_transient_model_error_is_not_fast_failed` (rate_limit flows through, `interrupts == 0`). `runner/tests` green; `ruff` + `mypy` clean.

### DIFF
--- a/runner/src/agentos_runner/session.py
+++ b/runner/src/agentos_runner/session.py
@@ -242,8 +242,13 @@ class SessionRunner:
             if _is_auth_rejection(message):
                 # A rejected model credential is terminal: stop the live session
                 # so the SDK/CLI does not keep retrying with backoff to the wall,
-                # then surface a distinct, immediate classified failure.
-                await self._session.interrupt()
+                # then surface a distinct, immediate classified failure. Suppress
+                # a failing interrupt (a wedged transport -- the very state a bad
+                # credential can cause) so it cannot propagate to the generic
+                # retryable ``runner-error`` handler and defeat the fast-fail; the
+                # terminal ``model-credential-rejected`` error is emitted regardless.
+                with contextlib.suppress(Exception):
+                    await self._session.interrupt()
                 for line in self._auth_halt_lines():
                     yield line
                 return
--- a/runner/src/agentos_runner/session.py
+++ b/runner/src/agentos_runner/session.py
@@ -34,7 +34,7 @@ from aci_protocol import (
     ToolNote,
     to_ndjson_line,
 )
-from claude_agent_sdk import ResultMessage
+from claude_agent_sdk import AssistantMessage, ResultMessage
 
 from .adapter import ModelSession
 from .budget import BUDGET_CLASSIFICATION, BudgetTracker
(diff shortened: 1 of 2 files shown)


TheConnMan (commented):
Two small things before merge: wrap the auth-path `await self._session.interrupt()` in `contextlib.suppress(Exception)` so a wedged transport can't drop through to the *retryable* generic `runner-error` handler and defeat the fast-fail. Also worth one manual run against the #121 repro to confirm a live OpenRouter 401 actually surfaces as `error == "authentication_failed"` — the whole mechanism hinges on that exact string.

alexrao2000:
> Two small things before merge: wrap the auth-path `await self._session.interrupt()` in `contextlib.suppress(Exception)` so a wedged transport can't drop through to the _retryable_ generic `runner-error` handler and defeat the fast-fail. Also worth one manual run against the #121 repro to confirm a live OpenRouter 401 actually surfaces as `error == "authentication_failed"` — the whole mechanism hinges on that exact string.

Addressed

PR #180, opened 2026-07-09
Reject zip symlink entries on bundle extraction; dedupe the worker extractor

### LINKED ISSUE #73: Refactor: safe archive extraction is copy-pasted 3x (and the zip path misses the symlink guard)
**Impact: high (security surface) · Effort: S (intra-package) / M (cross-service).**

**Evidence:** the "detect zip/tar, guard traversal, extract, unwrap bundle root" logic exists three times:
- `apps/api/src/agentos_api/bundles.py:28-87`
- `apps/worker/src/agentos_worker/bundle_store.py:74-93`
- `apps/worker/src/agentos_worker/eval/stream.py:163-178`

The worker's two copies are byte-for-byte identical. A real latent asymmetry: the tar path uses `tarfile.extractall(filter="data")` (py3.12+, which also rejects symlinks/special files) while the **zip path uses only a manual `..` check and does NOT reject symlink entries**.

**Fix:** deduplicate to one implementation. Intra-`agentos_worker` is a zero-dep change (`eval/stream.py` imports from `bundle_store`); folding in the API copy means adding `plugin-format` as a worker dep (its natural home). Consolidating lets the zip-symlink gap be closed once.

**Success criterion:** exactly one `safe_extract`/`bundle_root` exists in the repo; `grep -rn "\.\. in Path(name).parts"` (excluding tests) returns one hit; the single implementation rejects symlink entries in both the zip and tar paths.

jw3329:
Closes #73

## What
- **Security:** reject zip **symlink** entries on extraction, in both the API upload path (`bundles.extract`) and the worker extractor. A symlink entry like `config -> ../../etc/passwd` passed the name-only `..` check but escaped the extraction dir when followed. The tar path was already safe via `tarfile` `filter="data"`.
- **Dedupe:** collapse the two byte-identical worker copies into one — `eval.stream` now imports `safe_extract` from `bundle_store` (zero-dep, both inside `agentos_worker`).

## Validation
ruff + mypy clean; new unit tests for symlink/`..` rejection and normal extraction on both paths; full `test_bundles.py` (incl. real MinIO/Postgres round-trip) green.

## Scope note for reviewer
The gap is closed in all three original sites and the two worker copies are now one. The issue's literal "exactly one `safe_extract` in the repo" is **not** fully met: the API keeps its own `extract` (different signature, part of `extract_and_validate`), so the traversal-guard `grep` now returns 2 hits (API + worker), down from 3. Fully unifying API + worker needs a shared home; the issue suggests `plugin-format`, which is a **frozen contract** package — flagging that relocation as a follow-up decision rather than touching a frozen package here.

### DIFF
--- a/apps/api/src/agentos_api/bundles.py
+++ b/apps/api/src/agentos_api/bundles.py
@@ -7,10 +7,12 @@ router; this module is pure intake logic.
 """
 
 import io
+import stat
 import tarfile
 import tempfile
 import zipfile
 from pathlib import Path
+from typing import Literal
 
 from plugin_format import ValidationResult, validate_bundle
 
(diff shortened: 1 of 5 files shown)


TheConnMan:
Recommend closing this in favor of #181, which subsumes it: #181 fully unifies to a single `safe_extract`/`bundle_root` (this PR intentionally left two implementations) and ships a stronger malicious-archive test suite covering the same zip-symlink fix. Both close #73, so only one should land.

jw3329:
Closing in favor of #181 (merged), which subsumes this. #181 unifies to a single `safe_extract`/`bundle_root` in `plugin_format.archive` — the shared-home decision I explicitly flagged and deferred here — and lands the same zip-symlink rejection (`S_ISLNK`) plus a dedicated malicious-archive test suite. Nothing is dropped by closing. Thanks @TheConnMan.

Issue #17, opened 2026-07-07
Observability polish: unit misnomer, empty traces, log noise, missing-OTLP warning


TheConnMan:
Batch of small, independent observability fixes:

- [ ] **`latency_p95_seconds` field misnomer.** The metrics field (`apps/api/src/agentos_api/metrics.py`, `schemas.py`) is named `latency_p95_seconds` but actually carries milliseconds; any consumer reading it as seconds is off by 1000x. Rename the field and schema to the real unit, or convert at the boundary.
- [ ] **Empty-trace UX.** Stop writing observation-less trace shells for 0/0 eval replays; keep the honest empty state.
- [ ] **Worker idle-log noise.** The blocking `XREADGROUP` timeout logs a WARNING per poll cycle on both consumers, flooding logs at idle. Demote the routine blocking-read timeout to DEBUG; keep WARNING for real failures.
- [ ] **Missing-OTLP boot warning.** Warn at worker boot when the substrate is Docker and no OTLP endpoint is set, so silent no-telemetry configs are caught early.

PR #183, opened 2026-07-09
Observability polish: honest empty evals, quieter idle eval log, missing-OTLP warning

jw3329:
Part of #17 (batch of small observability fixes). Three of the four items land here; two are deliberately deferred (below).

## Done here (worker lane)
- **Empty-trace shells:** `recorder.record()` short-circuits for a 0/0 run (empty suite, or a failed-to-load bundle reported with `results=[]`) instead of POSTing an observation-less batch — keep the honest empty state.
- **Idle-log noise (eval consumer):** split the eval read-loop `except` so a routine blocking-read timeout logs at DEBUG (it fires every poll cycle on an idle worker) while a real `RedisConnectionError` still logs WARNING.
- **Missing-OTLP boot warning:** the worker warns at boot when `AGENTOS_SANDBOX_SUBSTRATE=docker` and `OTEL_EXPORTER_OTLP_ENDPOINT` is unset, so a silent no-telemetry config is caught early.

## Deferred, with reasons
- **`latency_p95_seconds` misnomer:** the honest fix (rename, or convert to seconds) also requires changing the UI consumer (`apps/ui/src/lib/format.ts`, `RealMetrics.tsx`), which is another lane's surface. Should land as one coordinated API+UI change, not a solo worker-lane push.
- **Idle-log noise on the *runs* consumer (`consumer.py`):** `consumer.py` is the sacred single-owner kernel module (`apps/worker/CLAUDE.md`). The same DEBUG/WARNING split belongs there, applied by that owner, not as a side effect of this ticket.

Leaving #17 open for those two.

## Validation
ruff + mypy clean; new unit tests for the empty-run skip (mock transport) and the OTLP boot warning (both present/absent). The `test_stream.py` failures are pre-existing on `main` (need Langfuse / a runner not booted here) and unchanged by this diff.

### DIFF
--- a/apps/worker/src/agentos_worker/eval/recorder.py
+++ b/apps/worker/src/agentos_worker/eval/recorder.py
@@ -46,6 +46,12 @@ class LangfuseEvalRecorder:
 
     async def record(self, run: EvalRunResult) -> list[str]:
         """Ingest one trace + score per case. Returns the created trace ids."""
+        if not run.results:
+            # A run with no case results (a 0/0 replay, an empty suite, or a
+            # failed-to-load bundle reported with results=[]) has nothing to
+            # observe. Posting it would create observation-less trace shells;
+            # keep the honest empty state and write nothing.
+            return []
         now = _now_iso()
         batch: list[dict[str, Any]] = []
         trace_ids: list[str] = []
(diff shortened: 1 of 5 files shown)


TheConnMan:
Recommend closing this in favor of #171, which is a superset of these worker changes (this PR's own body already defers the `consumer.py` hunk to #171). Closing here avoids merge churn on the same five files.

jw3329:
Closing in favor of #171, the superset of these worker observability changes (this PR already deferred the `consumer.py` hunk to #171). Avoids merge churn on the shared files. Thanks @TheConnMan.

PR #184, opened 2026-07-09
API: POST /evals/trigger — run a platform eval on demand

### LINKED ISSUE #10: agentos eval --remote: trigger a platform eval run on demand
Today platform eval fan-out is GitHub-push-triggered only. Add a CLI command that triggers a platform eval run on demand for the current agent/version by enqueueing on the `agentos:evals` stream with the frozen payload shape.

Depends on #8 (one frozen eval-case format) landing first to be worth building.

alexrao2000:
Closes #10 (API-endpoint side). Platform eval fan-out was GitHub-push-triggered only; this adds an authenticated endpoint to trigger a run on demand.

## What
`POST /evals/trigger` (evals router, under the shared `require_api_key`). Body: `{ agent_id, version_id?, suite?, target_url? }`. It enqueues the **same `EvalJobRequest`** onto `agentos:evals` that `gitflow.process_push` uses — minus the push-only (`dev + bundle_built`) gate.

- **Resolution** (via crud, no hand-rolled SQL): explicit `version_id` is ownership-checked against the agent; otherwise the agent's **active dev deployment** → its version. `sha = version.commit_sha or deployment.commit_sha`; `suite` defaults to `eval_default_suite`.
- **Clean errors, no stray enqueue:** unknown agent → 404, no active dev deployment → 404, unknown/foreign version → 404, missing sha → 400. Enqueue happens only after resolution succeeds.
- Returns the stream id.

## Decoupled from #8
The endpoint never imports/parses the eval-**case** models (`EvalCase`/`EvalSuite`/`Grader`) — the worker reads cases from the bundle at run time. The trigger contract is `EvalJobRequest` (suite by name + bundle by ref), so #8's case-format work can't change this endpoint. (#10's stated "depends on #8" is product-value, not a code coupling.)

## Tests
`test_evals_trigger_integration.py` (7, against real Postgres+Valkey): payload round-trips through `EvalJobRequest`; requires api key; unknown agent / no dev deployment / unknown version → 404 + no enqueue; suite default vs explicit; explicit version resolves its sha/bundle_ref. `openapi.json` regenerated (drift test passes). `ruff` + `mypy` clean. Rebased onto current `main`.

### DIFF
--- a/apps/api/openapi.json
+++ b/apps/api/openapi.json
@@ -567,6 +567,103 @@
         "title": "EvalReportResult",
         "type": "object"
       },
+      "EvalTriggerRequest": {
+        "description": "Ask for an on-demand platform eval run for an agent (issue #10).\n\nEnqueues the same EvalJobRequest the git-push fan-out uses, minus the\npush-only gate. With no version_id the agent's active dev deployment is\nevaluated; suite falls back to Settings.eval_default_suite when omitted.",
+        "properties": {
+          "agent_id": {
+            "format": "uuid",
+            "title": "Agent Id",
+            "type": "string"
+          },
+          "suite": {
+            "anyOf": [
+              {
+                "type": "string"
+              },
+              {
+                "type": "null"
+              }
+            ],
+            "title": "Suite"
+          },
+          "target_url": {
+            "anyOf": [
+              {
+                "type": "string"
+              },
+              {
+                "type": "null"
+              }
+            ],
+            "title": "Target Url"
+          },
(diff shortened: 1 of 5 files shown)


TheConnMan (commented):
One gap: `trigger_eval` enqueues without checking `version.bundle_ref` is populated, so a version whose bundle was never built silently enqueues an unrunnable eval (the worker reads eval cases from the bundle). Suggest mirroring the sha guard and returning 400 when `bundle_ref is None`.

PR #300, opened 2026-07-11
dx: one-command smoke test + surface the e2e path in the README

alexrao2000:
## Why

Verifying the system works end to end was hard to find: the runner-image build was buried in a parenthetical in the Quickstart, and `cli/scripts/e2e.sh` (the offline round-trip that actually exercises the ACI loop) wasn't mentioned in the README at all.

## What

- **`scripts/smoke.sh`** — one command that builds the runner image and runs `cli/scripts/e2e.sh` (init a bundle → boot a real runner container, fake model → send a message → stream the NDJSON reply → run evals → tear down). No Slack, no cluster, no credential.
  ```bash
  ./scripts/smoke.sh
  ```
- **README "See it work (60-second smoke test)"** callout at the top of the Quickstart, pointing to `scripts/smoke.sh` and to the full local loop (`agentos local up` → `local deploy` → `local message`) for the dispatcher→queue→worker→reply path, plus how to get a real-model reply.

Docs + a wrapper script only — no code changes. `bash -n scripts/smoke.sh` clean; the two commands it wraps are the repo's existing, documented e2e entry points.

### DIFF
--- a/README.md
+++ b/README.md
@@ -154,6 +154,33 @@ channel binding, and troubleshooting) see
 
 ## Quickstart
 
+### See it work (60-second smoke test)
+
+One command builds the runner image and round-trips a real turn through a runner
+container — no Slack, no cluster, no credential:
+
+```bash
+./scripts/smoke.sh
+```
+
+Green output means the ACI loop works end to end (init → runner → message →
+streamed reply → evals). To drive the **full** dispatcher → queue → worker →
+reply path locally, use the local loop (the *One-command middle mode* block
+below):
+
+```bash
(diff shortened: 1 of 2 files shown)


alexrao2000:
Superseded. main already has QUICKSTART.md (the 1-minute front door) using the CLI subcommands directly, so the README callout here is redundant. And per the direction that dev flows should be `agentos <command>` rather than loose scripts, the right shape is a native `agentos e2e` subcommand (replacing the smoke.sh wrapper and the cli/scripts/e2e.sh round-trip), which I'm opening separately.

PR #301, opened 2026-07-11
cli: 'agentos build' + 'agentos e2e' (dev flows as commands, not scripts)

alexrao2000:
## Why

Dev flows in this CLI are `agentos <command>`, not loose shell scripts. The offline end-to-end smoke test still lived as `cli/scripts/e2e.sh`, invoked by hand. This promotes it to a first-class subcommand: the whole runner-only, fully-offline round-trip is now `agentos e2e`.

## What

Adds a top-level `E2e` command (a peer of `init`/`skill`/`local`/`cluster`) whose handler runs the same round-trip `e2e.sh` does today, but **natively by reusing the existing handlers** (`scaffold`/`start`/`send`/`eval`/`stop`) rather than shelling out to the `agentos` binary:

1. Scaffold a throwaway `deal-desk` bundle into a temp dir.
2. `skill up --fake-model` on it (fully offline).
3. Send a message and stream the NDJSON reply.
4. Run the fake-model eval cases (final text `"all done"`, byte-mirroring the suite `e2e.sh` writes).
5. `skill down` + remove the temp dir on exit — even on failure — unless `--keep`.

Flags: `--image <ref>` (defaults to the same image `skill up` resolves via `artifacts::resolve_image`), `--port` (default `7245`), and `--keep` (skip teardown for debugging). Clear step headers are printed like the script.

### Constraints honored

- **Never runs `docker build`.** The command assumes the runner image already exists — built for dev with `docker build -f runner/Dockerfile -t agentos-runner .`, or pulled from GHCR on first run for a release binary, matching `skill up`'s image resolution exactly.
- No changes to the worker kernel or any frozen contract.
- `cli/scripts/e2e.sh` is untouched and CI is unchanged; the docs note it can later become a thin `agentos e2e` wrapper.

### Docs

Updated `cli/README.md` (top-level verb table + the Verify section) and `QUICKSTART.md` to present `agentos e2e` as the one-command offline round-trip.

## Note on teardown

Teardown runs around the round-trip via a captured `Result`, so the common failures (missing image, unhealthy runner, connection refused) clean up the container and temp dir. The two reused handlers that call `std::process::exit(1)` internally (a classified-failure reply, a failing eval case) will skip Rust cleanup — but the fake model always yields `"all done"`, so the offline path never hits them; a genuinely broken runner leaving the container is effectively `--keep` for inspection.

## Verify

`cargo fmt --check`, `cargo clippy --all-targets -- -D warnings`, and `cargo test` all pass (added two arg-parse tests mirroring the existing `main.rs` parse tests). `cargo build` + `agentos e2e --help` wires up correctly.

A full `agentos e2e` run needs the `agentos-runner` image, which is **not present** in this environment (and building it is out of scope here), so I verified build + `--help` + graceful-failure teardown (no leftover container or temp dir when the image is absent) rather than a green round-trip.

### DIFF
--- a/QUICKSTART.md
+++ b/QUICKSTART.md
@@ -33,6 +33,12 @@ That is the full loop. `agentos skill up` starts the runner container,
 `agentos skill down` stops it. Edit `skills/weather-bot/SKILL.md` in the bundle
 and re-run steps 2 and 3 to see your change answer.
 
+Want the whole offline round-trip in one command? `agentos e2e` scaffolds a
+throwaway bundle, runs `skill up --fake-model` -> `skill message` ->
+`skill eval` -> `skill down`, and cleans up after itself — the scripted
+equivalent as a first-class command. It only needs the runner image present
+(the same one `skill up` uses); it never builds an image.
+
 ## Level up: a real model
 
 The fake model returns scripted replies. To get a genuine answer, bring your own
--- a/cli/README.md
+++ b/cli/README.md
@@ -312,9 +312,25 @@ agentos e2e
 
 It scaffolds a throwaway bundle, runs `skill up --fake-model` -> `skill message`
 -> `skill eval` -> `skill down`, and cleans up on exit. Requires an
-`agentos-runner` image to be present (`docker build -f runner/Dockerfile -t
-agentos-runner .` from the repo root for dev; a release binary pulls its pinned
-GHCR ref on first run). It never runs `docker build`.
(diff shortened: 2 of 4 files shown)


alexrao2000:
Split up. The 'agentos build' half you wanted is now its own focused PR. Dropping the 'agentos e2e' smoke-test command — it was my extrapolation, not requested, and real e2e testing is better done manually. The e2e subcommand remains in this branch's history if it's ever wanted.

PR #357, opened 2026-07-13
cluster deploy self-plumbs the API port-forward

### LINKED ISSUE #352: `cluster deploy` does not port-forward like `cluster message` does — first deploy fails "platform API unreachable"
## Friction — `cluster deploy` doesn't port-forward, but `cluster message` does
In the run-2 cluster dogfood, `agentos cluster message` **auto-opens a port-forward** to the in-cluster API, but `agentos cluster deploy` **does not** — so a first `cluster deploy` fails with:

```
platform API at http://localhost:8000 is unreachable
```

until you manually run `kubectl -n agentos port-forward svc/agentos-api 8000:8000` yourself. The asymmetry is a surprise: the two verbs hit the same in-cluster API, so a user who ran `deploy` first (the natural order — deploy before message) hits a connection-refused on the tool's own happy path.

## Fix
Make `cluster deploy` establish the same port-forward (or the same API-reachability setup) that `cluster message` already does, so the deploy-then-message order works with no manual `kubectl port-forward`.

## Acceptance criteria
- `agentos cluster deploy` against a healthy cluster succeeds with **no** manual port-forward, matching `cluster message`'s behavior.
- `deploy` then `message` works end-to-end out of the box.

## Related
- Field evidence: cold-start run-2 cluster tier. Same first-run friction family as #341 (deploy's default slack channel).

TheConnMan:
`agentos cluster deploy` did not port-forward to the in-cluster platform API the way `agentos cluster message` does, so a first deploy (the natural deploy-then-message order) failed with `platform API at http://localhost:8000 is unreachable` until the operator ran `kubectl port-forward` by hand.

Now, omitting `--api-url` auto-establishes a kubectl port-forward to the in-cluster `agentos-api` service (matching `cluster message`) and releases it when the deploy returns, so a first deploy against a healthy cluster needs no manual port-forward.

- The self-forward is gated on the flag being **omitted** (`Option<String>`), not on its value, so an explicit `--api-url` / `AGENTOS_API_URL` (including `http://localhost:8000`, as the e2e deploy leg passes) is always dialed as given.
- Adds `--namespace`/`--release` to target the port-forward, mirroring `cluster message`.
- `start_port_forward` refuses up front when the local port is already in use and fails loud if kubectl exits early, so a stale or concurrent forward cannot silently divert traffic (hardens `cluster message` too).

Verified live on k8scratch: explicit `--api-url` dials directly with no forward; omitted `--api-url` takes the self-forward path; the port-collision guard fires deterministically. Full CLI suite (306) plus fmt/clippy green.

Closes #352

### DIFF
--- a/cli/README.md
+++ b/cli/README.md
@@ -159,7 +159,7 @@ Wraps the umbrella Helm chart and the deployed release, the way `linkerd` or
 | `agentos cluster status` | Report release health, pod readiness, and access URLs (read-only). |
 | `agentos cluster comms --slack` | Connect or disconnect a real Slack workspace with a thin `helm upgrade --reuse-values`; env-backed tokens are masked in dry-run output. |
 | `agentos cluster message "..."` | Drive the deployed release end to end with zero Slack: self plumbs kubectl port forwards, points the deployed worker at a local Slack stub (`helm upgrade --reuse-values`), enqueues, and prints the reply. |
-| `agentos cluster deploy` | Package the bundle as tar.gz and push it to the platform API (`--api-url`, default `http://localhost:8000`). Auth via `--api-key` or `AGENTOS_API_KEY`. |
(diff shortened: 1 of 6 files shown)


TheConnMan:
Closing in favor of #359. Review surfaced that `cluster deploy` only speaks the HTTP API (never Valkey), so it needs no tunnel at all: the UI NodePort already reverse-proxies `/api/` to the in-cluster API and its nodePort is pinned. #359 redoes this by auto-discovering the UI proxy URL instead of self-plumbing a port-forward.

Issue #244, opened 2026-07-10
Approval gates: durable Approval record + awaiting-approval status & suspend/resume lifecycle


alexrao2000:
Part of #22 (approval gates epic). Interface contract: `docs/interfaces/approval/INTERFACE.md`.

## Scope
The durable core of the approval primitive, CLI-verifiable without Slack:

- Durable `Approval` record (Postgres) with **resolve-once claim semantics** (conditional `UPDATE`) and optional SLA/expiry.
- `awaiting-approval` session status in the ACI protocol (`packages/aci-protocol`) plus kernel/worker lifecycle: suspend on pending, resume on resolve. This is the **first live use of the dormant suspend/resume path** (ADR-0003).

## Why grouped
The record and the status/lifecycle are one atomic capability — a durable pending state that survives restart. Nothing below can be built without it.

## Acceptance
Create an approval request from a skill, restart the worker/kernel, and confirm the session is still suspended and the record intact; resolve it and watch the session resume.

Issue #22, opened 2026-07-07
Epic: approval gates and human-in-the-loop workflows


TheConnMan:
Real business agents pause for a human decision: a discount needs sign-off, an invoice needs review, a remediation needs a yes before it runs. Today builders hand-roll all of it (Slack buttons, pending state, idempotent claiming, actor attribution), and the platform offers nothing: the runner hardcodes `permission_mode="bypassPermissions"` (`runner/src/agentos_runner/adapter.py`), there is no `canUseTool` callback, and `SessionStatus` has no awaiting-approval value. The suspend/resume path (ADR-0003) is built but dormant; approvals are exactly the hours-to-days pause it was designed for.

Two gate types on one shared primitive:
- **Policy gates** (business-level): the deterministic engine or skill decides something needs approval and invokes an approval-request primitive.
- **Permission gates** (tool-level): configuration marks a tool as approval-required and the runner intercepts the call (`side_effects.py` is the seam; today reactive, this makes it proactive).

High-level scope, to be broken into issues when picked up:

- [ ] Durable `Approval` record (Postgres) with resolve-once claim semantics (conditional UPDATE) and optional SLA/expiry
- [ ] `awaiting-approval` session status in the ACI protocol plus kernel/worker lifecycle (suspend on pending, resume on resolve; first live use of the dormant suspend path)
- [ ] Runner tool-permission gate: `canUseTool` replacing the hardcoded bypass, driven by per-agent approval config
- [ ] Worker-owned Block Kit approval cards with channel routing and a click-to-resolve flow through the dispatcher (losers of the claim race get "already resolved by X")
- [ ] Authorizer as an interface, enforced server-side at click time, self-approval blocked; first implementation is channel membership, with user-group, explicit user-list, and platform-RBAC implementations later
- [ ] Approval-policy config split: gate points and route names in the bundle manifest (versioned, evaluable); workspace-specific route bindings as per-agent deployment config (see PR #2, behavior packs)
- [ ] Platform audit log with authorizer snapshots (who resolved, and why they counted)

Design principle: gates are policy-triggered, never phase-hardcoded. The platform never forces an approval step; the agent's policy decides when one fires.

Related: PR #2 (behavior packs substrate) is the storage/plumbing foundation for route bindings.

[...]

PR #369, opened 2026-07-13
Add awaiting-approval to the frozen ACI contract

jw3329:
First slice of the approval-gates epic (#22, ADR-0010): the frozen `aci-protocol`
change every later lane compiles against, landed on its own per the frozen-contract
rule before dependent work proceeds.

## Changes
- `SessionStatus` gains `awaiting-approval`, a non-terminal pause: the session
  suspends until the approval resolves (ADR-0003), then resumes.
- `Final` gains two optional, additive fields: `session_id` (the SDK session id the
  worker rehydrates from on resume, the gap `apps/worker/CLAUDE.md` warns about) and
  `approval_request` (the gated tool call, so the worker builds the durable record
  without reconstructing it from a preceding `tool_note`).
- Bump `PROTOCOL_VERSION` 0.1.0 -> 0.2.0 and regenerate the committed JSON Schema,
  Rust crate, and TypeScript (no hand-edits to generated artifacts).
- Runner stamps `session_id` on the terminal finals (`translate.py`).
- Promote ADR-0010 Proposed -> Accepted. **Flagged for review**: this is the point we
  commit to the decision; push back here if that is premature.

## Verification
- `uv run pytest packages/aci-protocol/tests runner/tests -q`: 154 passed, 3 skipped
  (credential-gated live tests). New coverage: awaiting-approval wire value, a paused
  `Final` round-trip through the union, and `session_id` stamping.
- JSON Schema / Rust / TypeScript regenerated with no drift (schema-compat gate green);
  generated TypeScript compiles under `tsc`; the UI typechecks against the bumped types.
- `ruff` + `mypy` clean.
- `cargo check` on the generated Rust was not run locally (no Rust toolchain on the dev
  machine); the Rust source is verified against the models by the drift test, and CI's
  `cargo test` covers the compile.

Part of #244. Does not close it: this is only the contract slice. The durable Approval
record, the `canUseTool` gate, and the Slack resolve flow follow in subsequent PRs.

### DIFF
--- a/apps/ui/src/fixtures/traces.ts
+++ b/apps/ui/src/fixtures/traces.ts
@@ -50,14 +50,14 @@ export function traceSpans(t: TraceSummary): TraceSpan[] {
       detail: "deal-desk · skill.md",
       offset: "+12ms",
       kind: "skill",
-      event: { type: "tool_note", version: "0.1.0", text: "matched skill deal-desk", tool: "router" },
+      event: { type: "tool_note", version: "0.2.0", text: "matched skill deal-desk", tool: "router" },
     },
     {
       label: "Tool call",
       detail: `salesforce.get_deal(id: ${dealId})`,
       offset: "+840ms",
       kind: "tool",
-      event: { type: "tool_note", version: "0.1.0", text: `salesforce.get_deal(id: ${dealId})`, tool: "salesforce.get_deal" },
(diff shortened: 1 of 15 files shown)


TheConnMan:
Closing as superseded. The awaiting-approval status and the approval_summary / approval_route / approval_gate_kind / approval_granted_tool fields all landed on main independently (packages/aci-protocol events.py plus the generated Rust/TS and conformance), so this contract change is fully absorbed. Its stacked child #373 is being closed for the same reason.

Issue #246, opened 2026-07-10
Approval gates: Block Kit approval cards + server-side authorizer & click-to-resolve


alexrao2000:
Part of #22 (approval gates epic). Unblocked now that #111 (real Slack in compose) has landed.

## Scope
The human-facing resolve flow and who is allowed to resolve:

- Worker-owned Block Kit approval **cards** with channel routing and a click-to-resolve flow through the dispatcher (losers of the claim race get "already resolved by X").
- **Authorizer as an interface**, enforced server-side at click time, self-approval blocked. First implementation is channel membership; user-group, explicit user-list, and platform-RBAC implementations come later.

## Acceptance
Route a card to a channel; an authorized member clicks Approve and the session resumes; a non-member gets the ephemeral "you are not an approver" rejection; a second clicker gets "already resolved by X".

Issue #245, opened 2026-07-10
Approval gates: runner tool-permission gate (canUseTool) replacing the hardcoded bypass


alexrao2000:
Part of #22 (approval gates epic). Interface contract: `docs/interfaces/approval/INTERFACE.md`.

## Scope
Tool-level (permission) gates in the runner:

- Replace the hardcoded `permission_mode="bypassPermissions"` (`runner/src/agentos_runner/adapter.py`) with a `canUseTool` callback.
- Drive it from per-agent approval config: configuration marks a tool as approval-required and the runner intercepts the call proactively (`side_effects.py` is the seam — today reactive, this makes it proactive).

## Acceptance
Configure a tool as approval-required; confirm the runner pauses at the tool call and emits an approval request rather than executing.

PR #373, opened 2026-07-13
Durable approval record + suspend/resume lifecycle

jw3329:
Second slice of the approval-gates epic (#22, ADR-0010): the durable core that
pauses a session on an approval gate and resumes it on resolution. CLI-verifiable
without Slack.

**Stacked on #369** (the `awaiting-approval` contract change) — review that first;
this branch targets it, so the diff here is only the durable-core changes.

## API
- Durable `Approval` record (Postgres) + migration `0008` with **resolve-once
  compare-and-set**: a resolve flips `pending -> approved/rejected` exactly once
  (the click-race loser gets 409), self-approval is refused (403).
- `routers/approvals.py` (get / list / resolve) and an `ApprovalNotifier` that
  publishes a resume wake-up on `agentos:approval-events`, mirroring the kill switch.

## Worker
- **Kernel (sacred module)** handles an `awaiting-approval` final: captures the SDK
  session id + gated tool call off the final, persists a pending record, suspends
  the session, marks the event done. Persist-before-suspend-before-ack keeps a
  crash mid-pause resolvable.
- `ApprovalStore`: raw SQL over the shared engine (like `binding.py`) to write the
  pending row and mark it resumed.
- `ApprovalResumer`: subscribes to resolutions and enqueues a **synthetic turn**
  onto the runs stream, so the existing `claim -> SuspendedThreadError -> resume`
  path rehydrates and delivers the outcome. A periodic reconcile sweep catches a
  resolve whose wake-up reached no worker. Suspended-route TTL raised to a week for
  multi-day pauses; the durable row stays the source of truth.

## Verification
- API suite (200 passed): resolve CAS, self-approval 403, already-resolved 409,
  and the real-Valkey wake-up publish; against the disposable migrated Postgres.
- Worker suite (295 passed): the kernel suspend-on-approval lifecycle test (real
  Valkey + fake runner) and the resumer enqueue against real Valkey.
- ruff + mypy clean; committed OpenAPI regenerated.

## Review asks
- **This touches the sacred kernel** (`kernel.py`): per `apps/worker/CLAUDE.md` it
  needs the adversarial spec-vs-impl + side-effects-detective review before merge.
  The one edge flagged in-code for that pass is the finish-race-vs-suspend
  interaction (a follow-up opening a fresh turn on the idle sandbox at the instant
  the awaiting-approval final lands).
- The worker writes the `approvals` table via raw SQL (mirroring `binding.py`)
  rather than through the API, to avoid a new API-up dependency on the suspend path.

Part of #244. The runner `canUseTool` gate that actually emits `awaiting-approval`
(#245) and the Slack card + authorizer (#246) follow; the full restart -> resolve
-> resume E2E rides the epic acceptance demo once those land.

### DIFF
--- a/apps/api/alembic/versions/0008_approvals.py
+++ b/apps/api/alembic/versions/0008_approvals.py
@@ -1,0 +1,66 @@
+"""add approvals (durable human-in-the-loop approval records, #22 / ADR-0010)
+
+Revision ID: 0008
+Revises: 0007
+Create Date: 2026-07-13
+"""
+
+from collections.abc import Sequence
+
+import sqlalchemy as sa
+from alembic import op
+from sqlalchemy.dialects import postgresql
+
+revision: str = "0008"
+down_revision: str | None = "0007"
+branch_labels: str | Sequence[str] | None = None
+depends_on: str | Sequence[str] | None = None
+
+SCHEMA = "agentos"
+
+
+def upgrade() -> None:
+    op.create_table(
+        "approvals",
+        sa.Column("id", postgresql.UUID(as_uuid=True), primary_key=True),
+        sa.Column("agent_id", postgresql.UUID(as_uuid=True), nullable=False),
+        sa.Column("conversation_id", sa.String(), nullable=False),
+        sa.Column("session_id", sa.String(), nullable=True),
+        sa.Column("channel", sa.String(), nullable=False),
+        sa.Column("reply_placeholder", sa.String(), nullable=False),
+        sa.Column("reply_endpoint", sa.String(), nullable=True),
+        sa.Column("tool", sa.String(), nullable=False),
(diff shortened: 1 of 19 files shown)


TheConnMan:
Closing as superseded. The durable Approval record and awaiting-approval suspend/resume lifecycle landed independently via cdaa5887 (same title, opened a day later) and became the base of the merged approval epic: #411/#412/#418 (reconciler + expiry), #430/#437 (one-shot allowance, ADR-0035), #544 (gate convergence). Main's Approval schema is a strict superset of this branch, the awaiting-approval contract is already on main, and every capability here exists on main in a more-evolved form. There is no remaining delta to salvage, and a rebase would only reintroduce a divergent schema that conflicts with migrations 0009-0016.

Issue #429, opened 2026-07-15
Authed-MCP bundles can't promote past skill tier: no per-agent secret delivery to worker-spawned sandbox (ADR-0009)


TheConnMan:
## What

A bundle that authenticates an off-the-shelf MCP server with a secret (e.g. the
`examples/github-issues` bundle, which needs `GITHUB_PERSONAL_ACCESS_TOKEN`)
works at the **skill** tier but cannot be promoted to the **local/cluster**
platform tiers as-is: the secret never reaches the worker-spawned sandbox.

Evidence (v0.4.0-rc.1 cold-start E2E): after `agentos local message ... "triage"`,
the agent's `list_issues` call returned `Authentication Failed: Bad credentials`.
The GitHub token was present at the skill tier (forwarded by name via the CLI
`--secret` flag) but not at the local tier.

## Root cause

- Skill tier: `cli/src/docker.rs` `--secret NAME` forwards a named env var into
  the runner container. This is a **CLI-only** feature.
- Local/cluster tier: the **worker** (not the CLI) spawns the sandbox
  (`apps/worker/.../sandbox/docker.py`). It forwards only `binding.boot_env`
  (budget/session/agent-id/plugin-dir/approval-gate/memory+history refs), the
  model credential `AGENTOS_CREDENTIALS` by name, and the two ambient SDK creds.
  There is no field for arbitrary bundle/connector secrets.
- `AgentCreate`/`AgentUpdate` have no `secrets`/`connector_secrets` field;
  `local deploy` / `local up` have no `--secret`. `agentos secrets set` writes a
  host-side 0600 vault the containerized worker never reads.

This is exactly ADR-0009 ("per-agent secrets and connector credentials"), whose
status is Proposed -- the platform-side resolution is unbuilt.

## Why it matters

The parity promise is "one bundle behaves identically across skill -> local ->
cluster." Any authed-MCP bundle (the primary demo shape) breaks that promise
today: it runs at skill tier and fails at the platform tiers. The only workaround
is baking the secret as a literal into the bundle's `.mcp.json` env -- the
anti-pattern ADR-0009 explicitly rejects for production (secret lives in the
artifact).

## Direction (per ADR-0009)

Per-agent secret plane the worker reads when spawning the sandbox: a `secrets`
field on the agent record resolved from a K8s Secret (cluster) / the local vault
(compose), forwarded by name into the sandbox env alongside `AGENTOS_CREDENTIALS`.
[...]

PR #438, opened 2026-07-15
local tier: per-agent connector secret delivery (#429)

alexrao2000:
Restores **skill → local parity** for authed-MCP bundles. Today the CLI forwards `--secret` at the skill tier, but a bundle promoted to `local` fails with `Bad credentials` because the worker never delivers the connector secret to the sandbox it spawns (#429). This wires the per-agent secret plane end to end for the local tier (ADR-0009 / PR #435).

## Path: `deploy --secret` → agent row → worker → sandbox
- **CLI** — `agentos local deploy --secret NAME` (and `cluster deploy --secret`) resolve each value from the caller's env or the host secret vault (`agentos secrets set`) and PATCH it onto the agent. The value travels in the request body, never in argv.
- **API** — nullable `secrets` JSONB column on the agent (migration `0014`); `AgentCreate`/`AgentUpdate` accept a `{NAME: value}` map (env-var-name validated). **`AgentOut` returns names only — values never leave the API.** Regenerated `openapi.json`.
- **Worker** — `binding` resolves the agent's `secrets` and `boot_env` injects them **by name** into the sandbox env (the docker substrate forwards them as `-e KEY=VALUE`); a reserved boot-env key is never overwritten. `secrets_for()` lets the **eval lane** forward them too, so an authed bundle authenticates during eval exactly as on a bound run.

## Test
- API: `test_agent_secrets_round_trip_exposes_names_only` (values never serialize), non-env-var name → 422; `apps/api/tests` green (bar the pre-existing `test_control_integration` external-metrics ConnectError).
- Worker: `test_resolves_connector_secrets_into_boot_env` (SQL resolve → boot_env → `secrets_for`); eval-lane unit tests updated. `mypy`/`ruff` clean.
- CLI: `cargo fmt`/`clippy -D warnings` clean; command manifest regenerated; `cargo test` green.
- The worker binding + eval integration tests that need a live Postgres/Redis are environmental locally (stale dev DB / no Redis) and run on CI's fresh migrated stack.

## Manual e2e (unblocks the #429 repro)
```
agentos local up
agentos secrets set GITHUB_PERSONAL_ACCESS_TOKEN         # once
cd examples/github-issues && agentos local deploy --secret GITHUB_PERSONAL_ACCESS_TOKEN
agentos local message <channel> "List open issues in curie-eng/agentos, grouped by label."
```

Depends conceptually on the plugin-format `secrets` policy field (#436) but is base-independent. Cluster-tier per-agent K8s Secret + egress is the next PR.

Ref #429

### DIFF
--- a/apps/api/alembic/versions/0014_agent_connector_auth.py
+++ b/apps/api/alembic/versions/0014_agent_connector_auth.py
@@ -1,0 +1,31 @@
+"""add agents.secrets (per-agent connector secrets, ADR-0009 / #429)
+
+Revision ID: 0014
+Revises: 0013
+Create Date: 2026-07-15
+"""
+
+from collections.abc import Sequence
+
+import sqlalchemy as sa
+from alembic import op
+from sqlalchemy.dialects import postgresql
+
+revision: str = "0014"
+down_revision: str | None = "0013"
+branch_labels: str | Sequence[str] | None = None
+depends_on: str | Sequence[str] | None = None
+
+SCHEMA = "agentos"
+
+
+def upgrade() -> None:
+    op.add_column(
+        "agents",
+        sa.Column("secrets", postgresql.JSONB(), nullable=True),
+        schema=SCHEMA,
+    )
+
+
+def downgrade() -> None:
+    op.drop_column("agents", "secrets", schema=SCHEMA)
--- a/apps/api/openapi.json
+++ b/apps/api/openapi.json
@@ -67,6 +67,20 @@
             ],
             "title": "Repo Full Name"
           },
+          "secrets": {
+            "anyOf": [
+              {
+                "additionalProperties": {
+                  "type": "string"
+                },
+                "type": "object"
+              },
+              {
+                "type": "null"
+              }
+            ],
+            "title": "Secrets"
+          },
           "slack_channel": {
             "title": "Slack Channel",
             "type": "string"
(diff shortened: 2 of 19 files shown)


TheConnMan:
Before this lands: it also adds `--secret` to `cluster deploy` (`cli/src/main.rs` `ClusterAction::Deploy`), but cluster *delivery* is deferred to #440. On the cluster tier those values flow through the substrate-agnostic binding into `boot_env`, and `sandbox/k8s.py` (~L136) renders every boot_env entry as a plaintext `{name, value}` on the `SandboxClaim` CR except `AGENTOS_CREDENTIALS` — so `cluster deploy --secret` writes the token in plaintext into etcd (`get sandboxclaims` RBAC), the exact thing #439's per-agent Secret exists to prevent.

Suggest gating `--secret` off `cluster deploy` until #440 lands, or adding connector-secret keys to the k8s CR strip-list (same treatment as `AGENTOS_CREDENTIALS`). The local path looks good and fixes the authed-MCP break.

alexrao2000:
Good catch — fixed in b7028f8. Went with **both** of your suggestions, since the CLI gate alone doesn't cover the direct-API-PATCH path into `agents.secrets`:

1. **k8s CR strip (the real safety net).** `binding.boot_env` (and the eval lane) now mark the injected connector-secret keys in `AGENTOS_CONNECTOR_SECRET_KEYS`, and `sandbox/k8s.py` strips both that marker **and every key it names** from the claim env — same treatment as `AGENTOS_CREDENTIALS`. So no connector-secret value can land plaintext on the `SandboxClaim` CR regardless of how `agents.secrets` got populated. On k8s the value is simply **not delivered** (that's #440's `secretKeyRef` job) rather than leaked; the docker substrate still forwards it by value. New test: `test_connector_secrets_are_never_written_to_the_claim`.
2. **Gated `--secret` off `cluster deploy`.** Kept on `local deploy` only; the cluster verb carries a note pointing at #440. Command manifest regenerated.

The marker is also what #440 will key its per-agent-Secret `secretKeyRef` delivery off of, so this sets that up rather than just blocking. Local path unchanged.

Issue #512, opened 2026-07-16
Epic: closed-typed telemetry schema and close the approval-gate observability gap


TheConnMan:
## Context

Review of xAI's open-source Grok Build harness surfaced a telemetry discipline worth adopting: a **closed typed attribute schema** with an **export-time validator**, so an unlisted attribute key cannot be attached by construction and any record carrying one is dropped at export. Fail closed.

Grok models everything as counters/histograms plus structured log events over OTLP and emits no traces. **We deliberately do not copy that shape.** ADR-0038 decision 2 makes Langfuse the single authoritative backend and records that the absence of a Prometheus path for runner data is intentional (`apps/ui/CLAUDE.md:28-33`, `apps/ui/README.md:87-94`). Importing Grok's metrics-only model would reopen that decision and require a superseding ADR. This epic does not.

What is portable is backend-agnostic:

1. **Closed typed attribute schema over the existing gen_ai span stream.** `runner/src/agentos_runner/otel.py` emits `agent.run` / `llm.generation` / `execute_tool` spans with `gen_ai.*` attributes. Nothing today gates which attribute keys may be attached. A closed enum plus an export-time validator gives us the same drift-gate discipline the committed CLI JSON schema already has, and a privacy floor that matters because runners handle real credentials.
2. **Schema versioning** (`schema.version=v1`; additive changes do not bump, renames do), mirroring ADR-0036's ACI semver reader policy.
3. **Close the approval-gate observability gap ADR-0038 named as open.** Quoting ADR-0038: "did an approval get requested" is not in this span stream ... approval-gate observability is a gap this helper cannot serve from Langfuse today; it is called out as an open item, not promised. Emitting the approval decision (allow / deny / expired, per ADR-0034 and ADR-0035, issues #430 and #412) into the existing span stream closes that gap on the existing backend, with no federation and no second store.

## Roadmap — each stone ships independently and closes one named gap

The load-bearing decisions get their own short ADR (Stone 0) rather than being re-derived per stone; Stones 1-5 each implement one piece of it. Stone 6 is stretch and does not block closing this epic.

- [x] Stone 0 — Propose ADR-0076: closed-typed telemetry attribute schema #886
- [x] Stone 1 — Closed attribute-key enum + `schema.version` for runner OTel spans #887
- [x] Stone 2 — Export-time validator: fail closed on non-schema keys / unscrubbed secrets #888
- [x] Stone 3 — Emit approval-gate decisions (allow / deny / expired) into the span stream #889
[...]

PR #589, opened 2026-07-17
Redact secrets from runner stdout logs and gen_ai span attributes

### LINKED ISSUE #518: Secret-redaction pass over runner stdout and gen_ai spans
From the Grok Build review. **Worth-it-later**. Defense in depth.

Grok runs an output sanitizer over logs, telemetry, and tool output (`crates/codegen/xai-grok-secrets/src/sanitizer.rs:10-92`): a regex redactor covering `sk-` / `xai-` / `sk_`, AWS `AKIA`/`ASIA`, GitHub PATs, GitLab/Slack tokens, Google keys, PEM blocks, `Bearer`, bare JWTs, sensitive URL query params, and generic `secret=` / `password=`. It also collapses `$HOME` and username path segments (:187).

We forward **real credentials** into the sandbox and self-emit OTel. A redaction pass over runner stdout and gen_ai spans **before they cross the runner boundary** is cheap, model-agnostic, and shrinks credential-leak blast radius. This is belt-and-braces on top of the boundary, not a replacement for it.

Steal their tripwire test pattern too: adding a redaction regex forces adding a redaction pass, so the two cannot drift.

Pairs with the export-time validator in #512 (that drops records with unscrubbed secrets); this ticket is the scrub itself.

- [ ] Secret-redaction pass over runner stdout and gen_ai span attributes at the runner boundary
- [ ] Tripwire test: a new regex without a redaction pass fails CI

Relates to #512.

Source: review of https://github.com/xai-org/grok-build

alexrao2000:
Defense-in-depth (#518): the runner forwards a model credential and per-connector secrets into the sandbox, and an exception message or a provider error body can echo one back onto **stdout** (which the worker captures) or into a **gen_ai span attribute** (which ships to Langfuse).

New `redact` module is the single source of truth: one compiled pattern list (Anthropic `sk-ant-oat`/`sk-ant-`, OpenRouter `sk-or-`, OpenAI-style `sk-`, xAI, AWS `AKIA`/`ASIA`, GitHub, Slack tokens, `Bearer`, JWT, PEM private keys, URL credential params) and one `redact()` applied by both:

- a **root-handler logging filter** installed at boot — scrubs every `logger.*` line, most importantly the interpolated exception text at the credential-resolution / session-start / turn-failure sites; and
- the **one span-attribute choke point** `_set_str_attr`, through which every string span attribute the runner emits (`langfuse.trace.name`, `session/user id`, `gen_ai.request.model`, `model`, `gen_ai.tool.name`) already flows.

It is **values-only** — no log field or span attribute key is added or renamed — so it stays compatible with the frozen ACI env contract, the `check` CLI's committed JSON shape, and #512's forthcoming typed-telemetry validator (whose fail-closed drop runs *after* this scrub).

A **tripwire** test fails if a pattern is added to the list without `redact()` applying it (each pattern ships a synthetic sample that must be redacted); behavioral tests prove a secret is scrubbed from an interpolated log line and from an exported span attribute.

Closes #518
Ref CUR-1639

### DIFF
--- a/runner/src/agentos_runner/__main__.py
+++ b/runner/src/agentos_runner/__main__.py
@@ -17,7 +17,6 @@ import anyio
 from aiohttp import web
 
 from .adapter import ClaudeAgentSession, ModelSession, build_options
-from .redact import install_log_redaction
 from .approval import (
     ApprovalGate,
     build_approval_server,
@@ -38,6 +37,7 @@ from .hooks import load_bundle_hooks
 from .memory import MemoryRecord, MemoryStore, format_memory_preamble, resolve_memory
 from .otel import RunTracer, build_tracer_provider
 from .plugin import load_bundle_system_prompt, load_plugins
+from .redact import install_log_redaction
 from .sdk_auth import UnsupportedCredentialError, resolve_sdk_env
 from .server import create_app
 from .session import SessionRunner
--- a/runner/src/agentos_runner/__main__.py
+++ b/runner/src/agentos_runner/__main__.py
@@ -17,6 +17,7 @@ import anyio
 from aiohttp import web
 
 from .adapter import ClaudeAgentSession, ModelSession, build_options
+from .redact import install_log_redaction
 from .approval import (
     ApprovalGate,
     build_approval_server,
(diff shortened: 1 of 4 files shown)


alexrao2000:
Superseded by #640 (merged), which implements #518 at the same two boundaries. This branch is a duplicate of already-merged work — closing rather than rebasing (a rebase would either collapse to an empty diff or revert #640's reviewed implementation).

PR #657, opened 2026-07-17
Widen the weather eval to accept spelled-out degrees

### LINKED ISSUE #618: examples/weather has no real eval: the regex grades formatting and rejects a correct answer
`examples/weather/evals/cases.json` is the seed suite for the weather example and one of the files `init` teaches by. #527 replaced its vacuous `contains: "weather"` with a regex, but the replacement does not grade whether the agent did the work -- it grades output formatting, and it rejects a correct answer.

## Problem

The suite's only case is:

```json
{ "id": "reports-a-temperature",
  "input": "What's the weather in San Francisco today? Give the high in Fahrenheit.",
  "grader": { "kind": "regex", "expected": "\\d+\\s*°" } }
```

The graders apply no normalization beyond case-folding (`apps/worker/src/agentos_worker/eval/models.py:64-75`, mirrored `cli/src/evals.rs:46-68`), so the pattern needs a literal U+00B0 codepoint. Consequences:

- **A correct answer fails.** An agent that fetches a real forecast and replies "The high in San Francisco today is 68 degrees Fahrenheit" produces no `°` glyph and goes red. `re.IGNORECASE` maps case, not glyph-to-word. This is a false negative on the right behavior -- a worse defect class than the false positive the case's own `note` documents, because it punishes the agent for being correct.
- **A broken agent still passes.** An agent that calls no tool and invents "72°" is green. The `note` admits this and points at the ADR-0022 Phase 1 tool-call grader.

Net: the case cannot distinguish a real forecast from a hallucinated one, and actively prefers the hallucination that happens to use the glyph. "Weather in SF today" is a moving target with no stable string to anchor on, so no member of the frozen `exact | contains | regex` taxonomy can express the real assertion. This is not a bad regex to be tuned -- it is a case whose correctness is semantic, being graded by a string matcher.

## Why this matters beyond one file

`init` seeds new bundles from the examples, so the pattern propagates. ADR-0022 states the standard: a green eval should mean "the agent did the work". The weather example currently teaches "a glyph appeared".

## Dependency

The honest grader for this case is ADR-0042's semantic `GraderKind.verifier`, which explicitly supersedes ADR-0022's rejection of an LLM judge inside the frozen enum and divides by case type: deterministic graders assert what a trajectory makes checkable; the verifier judges correctness where no string match exists. Weather is squarely the latter.

**ADR-0042 is still `Proposed`, and its prerequisite is ADR-0022 Phase 1 (trajectory forwarding + the tool-call grader), which has no issue of its own.** Neither the verifier grader nor Phase 1 is built. The eval-case format is frozen (ADR-0019) and hand-mirrored in Rust, so a grader kind invented locally for this example would be contract drift.

## Scope

Rewrite the weather suite so its green means the agent actually fetched a forecast, using the semantic verifier once it exists. Blocked until then.

Interim option, if the false negative should stop hurting before the verifier lands: widen the pattern to accept the spelled-out unit (`\d+\s*(°|deg|degrees)`). That fixes only the false negative and leaves the hallucination hole open -- it is a stopgap, not this issue's acceptance.

## Acceptance

- The weather case's green requires evidence the agent fetched a forecast, not a glyph in the final text.
- A correct answer in plain English ("68 degrees Fahrenheit") passes.
- An agent that invents a plausible temperature without fetching one fails.
- The case carries no `note` deferring its own falsifiability.

Ref ADR-0042, ADR-0022, #527, #26

### LINKED ISSUE #620: Weather eval rejects a correct plain-English answer: accept spelled-out degrees (stopgap for #618)
Stopgap for the false-negative half of #618, which is blocked on ADR-0042's semantic verifier grader. This ticket does not close #618.

## Problem

`examples/weather/evals/cases.json` grades with `regex: "\\d+\\s*°"`. The graders apply no normalization beyond case-folding (`apps/worker/src/agentos_worker/eval/models.py:64-75`, mirrored `cli/src/evals.rs:46-68`), so the pattern requires a literal U+00B0 codepoint.

An agent that fetches a real forecast and answers "The high in San Francisco today is 68 degrees Fahrenheit" goes **red**. `re.IGNORECASE` maps case, not glyph-to-word. The eval punishes the agent for being correct, and the case's `note` does not mention it (it documents only the opposite failure).

## Scope

Widen the pattern to accept the spelled-out unit as well as the glyph, and correct the `note` to state both limits honestly.

This fixes only the false negative. The hallucination hole -- an agent that invents "72°" without fetching anything still passes -- is #618's scope and stays open. Do not attempt to close it here; no member of the frozen `exact | contains | regex` taxonomy can express it.

## Acceptance

- "68 degrees Fahrenheit", "68 deg F", and "68°" all pass.
- A refusal with no figure ("I could not confirm a current forecast") still fails.
- The `note` states both residual limits: the invented-temperature false positive (waiting on ADR-0042's verifier) and the fact that the case grades a figure's presence, not its provenance.
- If #619's positive-control gate has landed, add the spelled-out exemplar to it so the fix is enforced.

Ref #618, #527, #619

alexrao2000:
`examples/weather/evals/cases.json` graded with `regex: "\d+\s*°"`, and the graders normalize only case (`apps/worker/.../eval/models.py`, mirrored in `cli/src/evals.rs`) - no glyph-to-word mapping. So an agent that fetched a real forecast and answered "The high in San Francisco today is 68 degrees Fahrenheit" went **red**. The eval punished the agent for being correct, and the note did not mention it.

## Change
- Widen the pattern to `\d+\s*(°|deg)`. `deg` is a prefix of `degree`/`degrees` and the grader case-folds, so **"68 degrees Fahrenheit", "68 deg F", and "68°" all pass**; a refusal with no figure ("I could not confirm a current forecast") still fails closed.
- Correct the `note` to state both residual limits, neither expressible in the frozen `exact|contains|regex` taxonomy:
  1. **False positive** - a fake-model agent that invents a plausible temperature still passes; catching that needs the tool-call grader (ADR-0022 Phase 1) and the semantic verifier (ADR-0042, #618).
  2. **False negative** - a correct answer carrying no numeric-plus-degree-token figure (spelled-out "sixty-eight Fahrenheit", or a bare "68F") would still be graded red.

This is the stopgap for the false-negative half of **#618**; it does not close #618's hallucination hole.

## Verify
- Updated `loads_the_committed_weather_example` and added `weather_grader_accepts_spelled_out_degrees_and_still_rejects_a_refusal`, which drives the AC strings through the committed grader. `cargo test --lib evals` -> 17 passed.
- `cargo fmt --check` and `cargo clippy -D warnings` clean.
- Confirmed the Python grader engine matches the same strings (cross-engine parity).

Closes #620

### DIFF
--- a/cli/src/evals.rs
+++ b/cli/src/evals.rs
@@ -374,7 +374,23 @@ mod tests {
         // and the loader ignores the documentation-only `note` key on the case.
         assert_eq!(case.id, "reports-a-temperature");
         assert_eq!(case.grader.kind, GraderKind::Regex);
-        assert_eq!(case.grader.expected, "\\d+\\s*°");
+        assert_eq!(case.grader.expected, "\\d+\\s*(°|deg)");
+    }
+
+    #[test]
+    fn weather_grader_accepts_spelled_out_degrees_and_still_rejects_a_refusal() {
+        // #620: the glyph-only pattern graded a correct plain-English answer red.
+        // Drive the AC strings through the committed grader as `agentos skill eval`
+        // does, so the widened pattern is verified where the runtime grades.
+        let body = include_str!("../../examples/weather/evals/cases.json");
+        let (_dir, path) = write(body);
+        let grader = &load_suite(&path).unwrap().cases[0].grader;
+
+        assert!(grader.grade("The high in San Francisco today is 68 degrees Fahrenheit"));
+        assert!(grader.grade("High near 68 deg F"));
+        assert!(grader.grade("Today's high is 68°"));
+        // A refusal with no temperature figure must still fail closed.
+        assert!(!grader.grade("I could not confirm a current forecast"));
     }
 
     #[test]
--- a/examples/weather/evals/cases.json
+++ b/examples/weather/evals/cases.json
@@ -6,10 +6,10 @@
       "input": "What's the weather in San Francisco today? Give the high in Fahrenheit.",
       "grader": {
         "kind": "regex",
-        "expected": "\\d+\\s*°",
+        "expected": "\\d+\\s*(°|deg)",
         "case_sensitive": false
       },
-      "note": "Falsifiable: an agent that cannot fetch a matching forecast follows the SKILL rule and refuses ('I could not confirm a current forecast') with no temperature figure, which fails this grader -- unlike the old `contains: weather`, which the input itself guaranteed. LIMIT: a fake-model agent that INVENTS a plausible temperature still passes; proving the agent actually fetched a source needs the tool-call grader (ADR-0022 Phase 1)."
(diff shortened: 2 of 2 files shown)


alexrao2000:
Superseded: #659 landed the identical fix on main (same `\d+\s*(°|deg)` pattern, the updated loader assertion, and an equivalent grading test) and closed #620 while this was open. Closing as a duplicate; no delta to salvage.

PR #884, opened 2026-07-23
Speed up the e2e parity ladder: cache Rust + split the binary build + split the ladder

alexrao2000:
**Do not merge yet — for review.** Addresses "why does the parity ladder take so long?"

## Diagnosis (from a recent CI run)
The ladder job itself is **138s**, but it `needs: rust` (**350s**) just to receive the compiled `agentos` binary → ~**8 min** end-to-end. The `rust` job is slow because it has **no cargo caching** (full dep-tree recompile every run), compiles the tree **twice** (debug for tests + release for the artifact), and the ladder waits for fmt+clippy+test even though it only needs the binary.

## Three changes
1. **Cache cargo** in the `rust` job (`Swatinem/rust-cache@v2`) → warm runs recompile only what changed (~350s → est ~120–180s).
2. **Split the release binary into a dedicated cached `rust-build` job** that the ladder / falsifiability / cluster jobs depend on → they start as soon as the binary *compiles*, in parallel with fmt/clippy/test → ladder end-to-end est **~8 min → ~4 min**.
3. **Split the `local-release` rung into a parallel `e2e-ladder-release` job** so it overlaps skill+local.

## Can't break other PRs (same audit as #843)
- `actionlint` clean; every `needs:` resolves.
- The required-check **`Rust (fmt + clippy + test)`** keeps its exact name and still gates merges on test/clippy failure.
- The two new jobs (`rust-build`, `e2e-ladder-release`) are **not** in the ruleset's required set (Compose/Contracts/Python/Rust/UI), and nothing required `needs:` them → they can't block a merge.
- Binary artifact uploaded once (in `rust-build`), downloaded by all four consumers under the same name.

## Honest caveats / decisions for you
- **Not yet run on real CI** — a workflow edit only proves out on the first run. #1 speeds only *warm* runs (first run populates the cache).
- **#2 gating change:** the release-build *compile* check moves from the required `rust` job to the non-required `rust-build`. If a release-only compile break should stay merge-blocking, add **`Build the agentos release binary`** to the ruleset's required checks.
- **#3 is marginal:** a fresh runner shares no Docker daemon with `e2e-ladder`, so `e2e-ladder-release` rebuilds the same `:ci-local` images — extra build compute for a small wall-clock overlap, since the combined job was already short. **Recommend keeping #1+#2 and dropping #3** unless you want the overlap.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -119,6 +119,15 @@ jobs:
           persist-credentials: false
       - name: Toolchain
         run: rustup component add rustfmt clippy
+      # Cache the cargo registry + build artifacts keyed on the lockfiles + rustc
+      # so a run recompiles only what changed, not the whole dependency tree from
+      # scratch (the dominant cost of this job).
+      - name: Cargo cache
+        uses: Swatinem/rust-cache@v2
+        with:
+          workspaces: |
+            cli
+            packages/aci-protocol/generated/rust
       # The release-coupled versions (cli/Cargo.toml, Chart.yaml version +
       # appVersion) must agree on every PR, so a bump that forgets one field
       # fails here rather than shipping a chart that pulls the wrong image (#489).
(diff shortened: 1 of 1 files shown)


alexrao2000:
Split into three separate PRs for independent review/merge: #885 (cache cargo), #893 (split the release binary into a cached job), #894 (split the local-release ladder rung, stacked on #893). Closing this combined PR.

PR #894, opened 2026-07-23
Split the local-release ladder rung into a parallel CI job

alexrao2000:
**3 of 3** splitting the ladder-speedup work (was combined in #884). **Stacked on #893** (base `task/ci-split-release-build`) because the new job `needs: rust-build`, which #893 creates — review/merge #893 first (GitHub retargets this to `main` on merge).

Moves the `local-release` rung out of the combined `e2e-ladder` job into a parallel `e2e-ladder-release` job so it overlaps skill+local instead of running after them.

**Honest recommendation — this is the marginal one.** A fresh runner shares no Docker daemon with `e2e-ladder`, so `e2e-ladder-release` rebuilds the same `:ci-local` base images: extra build compute for a small wall-clock overlap, and the combined job was already short (~138s). The real wins are #885 (cache) and #893 (decouple from the 350s quality job). **Fine to close this unmerged** if the trade isn't worth it — it's isolated here.

Verified: `actionlint` clean; new job `needs: rust-build`; the `local-release` rung runs in exactly one job. Not yet run on real CI.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -493,14 +493,14 @@ jobs:
   # default (as docker/setup-buildx-action does) breaks that FROM resolution.
   # See the build steps' own comment for the full history (#697/#791).
   e2e-ladder:
-    name: E2E parity ladder (skill + local + local-release, fake model)
+    name: E2E parity ladder (skill + local, fake model)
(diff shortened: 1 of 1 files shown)


alexrao2000:

Issue #697, opened 2026-07-19
Share the GHA layer cache with the e2e-ladder job's image builds


TheConnMan:
## What

The `e2e-ladder` CI job added in #690 builds three images with plain `docker build` and no cache: the runner, the api image, and the worker base. The pre-existing `images` job already builds the identical `runner/Dockerfile`, `apps/api/Dockerfile`, and `apps/worker/Dockerfile` on the same commit with `cache-from`/`cache-to: type=gha`. Add `docker/setup-buildx-action` to the ladder job and switch its builds to `docker buildx build --load --cache-from type=gha --cache-to type=gha`, so most of those layers come back as cache hits.

## Why

The ladder job cold-rebuilds three images on every run, including the apt and dependency layers, moments after another job in the same workflow has already built two of them and written their layers to the GHA cache. This is the slowest job in the workflow and it runs on every PR.

## Notes

Deliberately not done in #690. That PR added the job and had just verified it green end to end, including the private-registry and docker-in-docker paths that were the risky part. Swapping the build mechanism underneath a freshly verified gate for a performance gain was the wrong trade at that moment. Nothing here needs cross-job image artifacts; it only shares the layer cache store, which is designed for this.

PR #904, opened 2026-07-23
Skip the whole e2e parity ladder on PRs that can't affect it

alexrao2000:
Follow-up to the ladder-speedup work (#885/#893 merged). Addresses "the ladder is still super long — what else can we do?"

## Why
With the Rust cache + rust-build decouple live, the ladder's remaining cost is its **own ~2min**, and ~55% of that is image builds rebuilt from scratch each run (runner image ≈ 42s). Today the skill+local rung pays that on **every PR** — docs, UI-only, gtools, non-ladder CI — none of which can affect what it tests.

## What
Generalize the cluster-rung path gate from #843: the `changes` job's filter (`charts/ cli/ apps/ runner/ packages/aci-protocol/ .github/workflows/ci.yaml`) is exactly the ladder-relevant path set. Rename its output `cluster` → `ladder` and gate **both** the skill+local rung and the cluster rung on it. An unrelated PR now **skips every ladder rung** (neutral check, not red); it still runs on every `push` to main and on any PR touching the tested surface.

## Safety
- No ladder rung is a **required** check (the ruleset requires only Compose/Contracts/Python/Rust/UI), so skipping can never block a merge — same reasoning as #843.
- `actionlint` clean; the `cluster`→`ladder` rename is fully consistent (no stale references); both rungs `needs: [rust-build, changes]`, `if: needs.changes.outputs.ladder == 'true'`.

## Note
This attacks a different axis than the split (#894, which parallelizes local-release): here we stop paying the ladder's runtime *at all* when nothing it exercises changed. **#894's `e2e-ladder-release` should gain the same `if: needs.changes.outputs.ladder` gate** when it merges — I'll update that branch. And the biggest remaining *runtime* lever (for PRs that DO run it) is a docker layer cache for the image builds, done without repointing the default builder (the #697/#803 landmine) — a separate, careful follow-up.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -504,7 +504,12 @@ jobs:
   e2e-ladder:
     name: E2E parity ladder (skill + local + local-release, fake model)
     runs-on: ubuntu-latest
-    needs: rust-build
+    needs: [rust-build, changes]
+    # Skip on PRs that can't affect what the ladder tests (docs, UI-only, gtools,
+    # non-ladder CI). Not a required check, so skipping never blocks a merge; it
+    # just stops paying the ~2min image-build + rung cost when nothing it
+    # exercises changed. Always runs on push to main (the changes job forces it).
+    if: needs.changes.outputs.ladder == 'true'
     steps:
       - uses: actions/checkout@v7
         with:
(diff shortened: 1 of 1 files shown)


alexrao2000:

PR #905, opened 2026-07-23
Layer-cache the e2e ladder image builds without repointing the default builder

alexrao2000:
The other half of "the ladder is too slow." [#904](https://github.com/curie-eng/agentos/pull/904) skips it when irrelevant; **this cuts the runtime for the runs that do execute.**

## The problem
The ladder rebuilds its images from scratch every run. The **runner image alone is ~58s** (apt+nodejs, `npm i -g claude-code`, a python venv/pip block). That's ~55% of the ladder's own runtime, paid every time.

## Why the obvious fix (#697) broke (#803)
#697 added the GHA cache with a bare `docker/setup-buildx-action` — whose default `use: true` makes the docker-container builder the **current** one. The compose-triggered worker-local overlay (`FROM :ci-local`) then ran on it and couldn't resolve the locally-built tags → "failed to resolve source metadata" on every run, until it was reverted to plain `docker build`. The revert note asked for *"a mechanism that does not repoint the default builder."*

## The fix
The one line #697 lacked: **`use: false`.** It creates the container builder for the cache export but leaves the plain `docker` builder current, so the compose `FROM` still resolves. Each cached build targets the named builder explicitly (`builder:` input) with `load: true` to export into the same daemon store the overlay reads. Same SHA-pinned actions + `type=gha` cache as the already-green `images` job (which proves the token wiring in this CI). The overlay build itself **stays plain `docker build` on purpose** — it's the `FROM` consumer.

Applied to both rungs (`e2e-ladder`, `e2e-ladder-cluster`); shared cache scopes across both.

## Validated locally before shipping (kind/Docker box)
Not shipped blind — the #803 mechanic was reproduced and disproven, and the speedup measured on the **real** runner Dockerfile:
- The named builder **never becomes default**; a plain `docker build` whose `FROM` is the named-builder-built `:ci-local` tag **resolves** (the exact #803 scenario).
- With the builder's internal cache **wiped** (fresh-runner simulation), a warm build restores layers from the **external cache alone** — the real runner image went **58s → 15s**, 9 layers `CACHED` (the apt / npm / pip blocks). The residual 15s is just `--load` exporting the image, which is unavoidable.

## Safety
No ladder rung is a required check (ruleset requires only Compose/Contracts/Python/Rust/UI), so even a regression here can't block merges — and if `use: false` ever regressed, the ladder's own `FROM` resolution would red-fail loudly (same symptom as #803), catching it on this PR rather than silently. `actionlint` clean.

Independent of #904 (that PR touches the job `if:`/`needs:` headers; this touches the build steps) — clean-merges in either order.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -496,11 +496,12 @@ jobs:
   # the generator's T1 substitutes for the build directive, both resolved
   # locally so `local up` never attempts a pull against a private GHCR ref.
   #
-  # All three image builds here use plain `docker build`, not buildx: two of
-  # the tags are consumed as Dockerfile FROM bases by a compose-triggered build
-  # later in the same job, and a docker-container buildx builder set as the
-  # default (as docker/setup-buildx-action does) breaks that FROM resolution.
-  # See the build steps' own comment for the full history (#697/#791).
+  # The three image builds here are GHA-layer-cached via a buildx builder that
+  # is created but NOT made the default (`setup-buildx-action` with
+  # `use: false`): two of the tags are consumed as Dockerfile FROM bases by a
+  # compose-triggered build later in the same job, which must stay on a plain
+  # `docker` builder to resolve them. See the build steps' own comment for the
+  # full history and why `use: false` is the fix (#697/#791/#803).
   e2e-ladder:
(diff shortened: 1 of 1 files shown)


alexrao2000:

Issue #630, opened 2026-07-17
Epic: replace the browser-held platform key with authenticated console sessions


TheConnMan:
## Problem

The console sends the shared platform administrator key from browser JavaScript. `apps/ui/src/api/config.ts` resolves it from `?api_key=`, else `VITE_API_KEY`, else the published dev default. In a sealed cluster the platform key is randomized, so the console either fails to authenticate or requires the administrator key in a URL. That exposes a credential with control over deployments, approvals, logs, traces, budgets and kill controls to browser history, request logs, referrers and plaintext NodePort traffic, and the only revocation is rotating the Secret and restarting the API, which breaks the worker and runner too.

Release blocker.

## Outcome

A single operator, single tenant, self hosted console uses browser safe authentication and authorization without exposing the platform key to browser code or URLs.

## Decision

[ADR-0083](docs/adr/0083-console-sessions-and-cli-minted-login-codes.md) (PR #1029): a server-managed **revocable session cookie**, established by exchanging a **single-use login code the CLI mints** under the platform key. Only hashes are stored; revocation is a column write; the cookie is `HttpOnly`/`Secure`/`SameSite=Strict`; and `require_api_key` accepts the platform key **or** a live session, in that order, staying the single dependency every router shares.

## Why this is an epic now

The first attempt, PR #645, was **5,759 lines across 43 files** and was closed unmerged: too drastic to be confident it worked, and hard to debug if it did not. Its design was sound and is carried forward in ADR-0083; what changes is the delivery.

The slices are ordered so that **slices 1-3 are additive and invisible** -- they can land and sit inert, with no existing authentication path changed -- and the single user-visible swap happens once, in slice 4, on top of a server side already proven. No slice is net-negative on its own.

| Slice | Issue | Change | Blast radius if wrong |
| --- | --- | --- | --- |
| 1 | #1044 | Session store, migration, the two exchange endpoints (wired to nothing) | None |
| 2 | #1045 | `require_api_key` also accepts a session; platform-key path first and byte-identical | One function, additive; machine callers never reach the new branch |
| 3 | #1046 | `curie console login` mints a code; the key is discovered, never printed | Additive verb |
| 4 | #1047 | Console uses the cookie **and** `?api_key=` is deleted, in the same PR | The only user-visible change, on a proven server side |
| 5 | #1048 | Prove no key reaches history, request logs, referrers, or built assets | None (tests and docs) |

[...]

PR #1029, opened 2026-07-28
ADR-0083: console sessions and CLI-minted login codes

alexrao2000:
Refs #630 (release blocker). **Decision only** — the implementation follows in reviewable slices.

## Why an ADR first
`apps/api/CLAUDE.md` says not to add a second auth scheme for a router "without raising it in an issue/PR first". This is that raising, and #630 is a change to how the platform is authenticated, so the reasoning belongs in the record before the diff.

## The decision
A server-managed, **revocable session cookie**, established by exchanging a **single-use login code the CLI mints** under the platform key. The browser never receives the platform key on any path.

- **Only hashes stored** (login code + session token), so a database read cannot replay a session.
- **Revocation is a column write** — a durable row a human can kill, not a signed token that stays valid until expiry.
- **`HttpOnly`, `Secure`, `SameSite=Strict`** — page script cannot exfiltrate the credential it authenticates with.
- **`require_api_key` accepts the platform key *or* a live session, in that order**, staying the single dependency every router shares — extending it rather than adding a second scheme, which is the boundary `apps/api/CLAUDE.md` draws.

That ordering is load-bearing beyond precedence: a machine caller returns *before* the session store is read, so a database outage can't take the platform-key path down with it.

## Provenance, and why not rebase #645
PR #645 built this on 07-17 and was **closed unmerged on 07-27 with no stated reason**, `CONFLICTING`, 432 commits behind main across the repository-wide rename, with its migration and ADR numbers both since taken.

I checked its design against this repo's constraints before adopting it — notably that it *explicitly respects* the "no second auth scheme" boundary by extending the shared dependency. It's sound, so the closure reads as **staleness, not rejection**. If anyone knows otherwise, that's the one thing that should change this ADR.

So the design is adopted and its four rejected alternatives carried forward, but the implementation will be written against current main rather than rebased — 5,759 lines across 43 files over a rename is more risk than value. Same "mined, not reused" call ADR-0060 Decision 6 made.

## Planned slices
1. API session store + Alembic revision + `require_api_key` extension
2. CLI `console login` verb (sources the key via existing `discover_api_key`, never prints it)
3. Console exchange + removing the `?api_key=` path
4. `cluster status` stops printing a secret-bearing URL
5. The leak proofs #630 requires: history, request logs, referrers, built assets

`curie dev docs-lint` clean; ADR index regenerated, not hand-edited.

### DIFF
--- a/docs/adr/0083-console-sessions-and-cli-minted-login-codes.md
+++ b/docs/adr/0083-console-sessions-and-cli-minted-login-codes.md
@@ -1,0 +1,117 @@
+# 83. Console sessions and CLI-minted login codes
+
+Date: 2026-07-28
+
+Status: Accepted
+
+Implements [#630](https://github.com/curie-eng/curie/issues/630).
+
+## Context
+
+The console sends the shared platform administrator key from browser JavaScript.
+`apps/ui/src/api/config.ts` resolves it from `?api_key=`, else `VITE_API_KEY`,
+else a published dev default. In a sealed cluster the platform key is randomized,
+so the console either fails to authenticate or requires the administrator key in
+a URL.
+
+That key authorizes deployments, approvals, logs, traces, budgets and the kill
+switch. Putting it in a URL puts it in browser history, request logs, referrers
+and any plaintext NodePort traffic, and the only way to revoke it is rotating the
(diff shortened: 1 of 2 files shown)


jw3329 (approved):
accepting

Issue #1062, opened 2026-07-29
Onboarding: a first-time user cannot get a bot running, and fails before writing any bot code


jw3329:
## The problem

Building a real bot on Curie end to end took roughly two days. The bot itself — persona, tools, evals — took about an hour and is 184 lines. Everything else was platform friction.

A user with no Kubernetes or AWS background could not have finished. They would stop at **step one**: `curie cluster up` from a fresh source checkout fails with `manifest unknown`, before they write a single line of their agent.

To actually finish required: Kubernetes NetworkPolicy semantics and kube-proxy DNAT ordering, Helm values, AWS EC2/IAM/OIDC/SSM/security groups, Docker multi-arch cross-compilation, the MCP streamable-http handshake, GitHub Actions with OIDC federation, and debugging silent failures across five systems that each reported success.

That is a platform engineer's skill set, not a bot author's.

## The measurable symptom

The resulting bundle repository:

```
the actual bot        184 lines   (SKILL.md, evals, .mcp.json, plugin.json)
platform plumbing     524 lines   (GitHub Actions, k8s Deployment/Service/NetworkPolicy)
```

**Three times more platform code than product code**, all of it written by the bot author, all of it work Curie should own. Every future bundle author will write their own slightly different, slightly wrong copy.

## What the flow should be

```bash
curie init acme-bot
# describe the bot in SKILL.md -- the only file you write

curie connect slack            # OAuth in the browser, pick a channel
curie deploy                   # provisions, wires, prints "live in #sre"
```

Roughly four steps. We took about forty.

## The irreducible user input

Only three things genuinely require the user; everything else is derivable:

1. **Intent** — what the bot should do. This is the product.
2. **Credentials to their own systems** — a Grafana token, Slack access. Curie cannot mint these.
3. **Where it runs** — their cloud account, or Curie-hosted.

Cluster provisioning, connector hosting, networking, secret plumbing, CI, and deploy are all Curie's.

## Sub-issues

Each is filed separately with reproduction detail:

- **Declarative connector hosting** — the single biggest win; eliminates 184 of the 524 plumbing lines
- **Agents have no lifecycle** — cost a full platform rebuild mid-project
- **Release artifacts are incoherent** — why `cluster up` fails out of the box
- **Failures are silent** — the largest single time sink
- **`cluster up` destroys `comms` configuration**
- **CLI discovers the API through the UI service**
- **Slack onboarding is manual and hardcodes Curie branding**
- **No conversable dev environment**
[...]

PR #1100, opened 2026-07-29
cluster up: preserve what cluster comms configured (#1067)

### LINKED ISSUE #1067: cluster up silently destroys what cluster comms configured
Refs #1062.

## Reproduce

```bash
curie cluster comms --slack --app-token xapp-... --bot-token xoxb-...
# -> "Slack connected", dispatcher running, "socket mode connected"

curie cluster up --set <anything>
# -> dispatcher Deployment gone, Slack tokens gone from release values
```

No warning, no mention in output. The bot stops receiving Slack messages and nothing says why.

## Cause

`cluster up` runs `helm upgrade --install` with the flags given. Values written by `comms` are not preserved unless re-passed, so a later `cluster up` reverts `dispatcher.deploy` to its default and drops `dispatcher.slack.*`.

## Impact

`cluster up` is the documented way to change platform settings — egress, model credential, image tags. Every one of those silently disconnects Slack if `comms` ran earlier.

The operational rule turns out to be "always run `comms` last". That is nowhere in the docs and is not discoverable before it bites. We hit it mid-build and spent time debugging a dispatcher that had been deleted by our own previous command.

## Suggested fix

Preserve `comms`-owned values across `cluster up` (the release is already resolved; `--reuse-values` semantics for that subtree), or refuse and tell the operator to re-run `comms`. Silently removing a working Slack connection is the worst of the three options.

jw3329:
Fixes #1067. Refs #1062.

## The problem

`comms` writes Slack tokens with `helm upgrade --reuse-values`. `up` does a **full** upgrade and drops anything it doesn't pass. So any `up` after a `comms` deleted the dispatcher and its tokens.

Nothing said so — no error, no warning, nothing in the diff mentioning Slack. The bot just stopped answering. Finding it meant noticing the dispatcher pod had vanished, then reading `helm get values` to see the tokens were gone.

The undocumented rule was *"always run comms last"* — an ordering constraint a tool should enforce, not expect people to remember.

## The fix

Follows the pattern already there for generated secrets (`resolve_generated_secrets`): read the release's recorded values, re-supply the keys `comms` owns, let an explicit `--set` win.

```
preserving 2 value(s) set by `cluster comms`; run comms again only to change them
```

**Only preserves, never invents.** Fresh install → nothing. Release that never had Slack → nothing. The empty strings `comms --disconnect` writes → nothing, so a disconnect stays disconnected.

## Verification

```
cargo test              → 532 + 53 + 5 + 10 + 17 + 13 + 14 + 4, all passing
cargo fmt --check       → clean
cargo clippy -D warnings → clean
```

Four tests: tokens preserved; explicit `--set` overrides; fresh/absent/disconnected all preserve nothing; and an end-to-end check that a preserved token reaches the **private values file** and never argv or the printed line — it's a live credential, so it gets the same handling as any generated secret.

**Not verified against a live cluster** — my AWS session expired mid-way. The logic is pure and covered, but the real-cluster round trip (comms → up → tokens survive) hasn't been run.

### DIFF
--- a/cli/src/ops.rs
+++ b/cli/src/ops.rs
@@ -754,6 +754,19 @@ pub fn model_egress_status_lines(
 /// their own paths (`cluster comms`, `CURIE_MODEL_CREDENTIALS`), not
 /// generated. `langfuse.encryptionKey` must be exactly 64 hex chars, so its 32
 /// bytes are load-bearing.
+/// Values owned by `cluster comms`, not by `cluster up`.
+///
+/// `comms` writes these with `helm upgrade --reuse-values`; `up` does a full
+/// upgrade and therefore drops anything it does not itself pass. That silently
+/// deleted the dispatcher and its Slack tokens whenever `up` ran after `comms`
+/// -- the bot simply stopped answering, with no error and nothing in the diff
+/// to suggest `up` had touched Slack at all (#1067).
+///
+/// Preserved the same way generated secrets are: read back from the release and
+/// re-supplied through the values file, never argv, so a bot token cannot leak
+/// into `ps` or shell history.
+const COMMS_MANAGED_KEYS: &[&str] = &["dispatcher.slack.appToken", "dispatcher.slack.botToken"];
+
 const REQUIRED_SECRETS: &[(&str, usize)] = &[
     ("postgres.auth.password", 24),
     ("valkey.password", 24),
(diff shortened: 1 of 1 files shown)


jw3329:
## Now verified against a live cluster

The PR originally said this was unit-tested only. It has now been run end to end on a real single-node k3s release with Slack connected.

**Old CLI — reproduces the bug:**

```
before:            SET/SET
installing release acme-bot: ok (41.4s)
after OLD:         GONE/GONE
  dispatcher:      DELETED
```

Both Slack tokens dropped and the dispatcher Deployment removed, from a `cluster up` that never mentioned Slack. No error, no warning.

**Fixed CLI — same command, same cluster, same starting state:**

```
after comms:       SET/SET
preserving 2 value(s) set by `cluster comms`; run comms again only to change them
after FIXED:       SET/SET
  dispatcher:      present
```

Then a full restore with credentials and egress: tokens `SET`, `fakeModel: False`, dispatcher rolled out, `socket mode connected`.

## A false negative worth recording

My first attempt at this appeared to show the bug did **not** reproduce — tokens survived the old CLI. That was wrong: the box runs a dev build, which needs `--chart`, and I had sent output to `/dev/null`. The command errored and changed nothing, which looked exactly like "preserved".

Worth noting because it is the same failure shape as the bug itself: a command that silently does nothing is indistinguishable from one that worked. The fix for the test was the same as the fix for the feature — stop hiding the output.

PR #1131, opened 2026-07-30
Restore first-run documentation to the current CLI contract

### LINKED ISSUE #1126: Restore first run documentation to the current CLI contract
Refs #1013, #1102, and ADR 0014.

The README readability rewrite now teaches several workflows that disagree with the shipped CLI.

1. The first cluster deploy omits --repo, then promises later Git pushes deploy automatically. The created agent has no repository binding and cannot be rebound.
2. Skill mode is described as an editable bind mount, but it runs an immutable snapshot and requires skill up --replace after edits.
3. The literal curie init tree omits the plugin manifest, gitignore, AGENTS.md, and the generated using Curie skill.
4. The tier table omits local and cluster eval although both commands are shipped and present in the command manifest.
5. docs/operations.md says cluster deploy cannot set repo_full_name, after #1102 added the creation time option.

Acceptance criteria:

1. Following the documented fresh cluster deploy creates an agent bound to the documented repository, and a signed dev push creates a version rather than returning ignored.
2. The skill edit loop names immutable snapshot behavior and the replace command.
3. Literal scaffold output matches a real curie init run or is explicitly marked abridged.
4. Tier command claims agree with cli/command-manifest.json.
5. The operations guide states that --repo applies only during agent creation.

TheConnMan:
The README first-run flow and the operations guide had drifted from the shipped CLI in five ways. Each is corrected against `cli/command-manifest.json` and the built binary.

1. **The first cluster deploy now binds the repo.** Step 3 passes `--repo <owner>/<name>`, and step 4 no longer promises a later "connect this bundle's repo" step, which has no code path: the binding is set only at agent creation and cannot be changed afterwards. Step 4 also no longer claims the step 3 binding survived the `cluster down --yes` that ends step 3, and it now names API reachability alongside the webhook and secret.
2. **Skill mode is described as the immutable snapshot it runs**, with `curie skill up --replace` named as the way an edit reaches a running runner. The prior "bind mounts its working directory" claim was wrong, and contradicted `QUICKSTART.md` and `cli/CLAUDE.md`.
3. **The literal `curie init` tree shows all seven scaffolded files**, verified against a real `curie init` run. The four that were missing are annotated in the bullet list below it.
4. **`eval` added to the `local` and `cluster` verb cells.** The sentence after the table no longer reads as an exhaustive verb list, since the table is a curated subset.
5. **The operations guide's `cluster deploy` flag table gains a `--repo` row** and states the flag applies only during agent creation. Its "needs no other flags by default" paragraph was telling readers the opposite of the git-flow section in the same file.

Verification: `scripts/check-docs.sh` passes; the scaffold tree was diffed against a real `curie init` run; `--repo`, `skill up --replace`, `local eval`, and `cluster eval` were each confirmed present in the built CLI; the 36 gitflow tests covering the bound-vs-unbound push paths pass.

Scope note: `curie local deploy` in step 2 deliberately keeps no `--repo`, since git-flow is demonstrated at the cluster tier only. The tier table was deliberately kept a curated subset rather than expanded to every verb in the manifest.

Closes #1126

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,8 +28,10 @@ the same plugin bundle climbs three tiers.
 An environment difference then shows up as a bug while progressing through these tiers, not a surprise
 your users hit - letting you iterate fast locally and ship with confidence.
 
-`local` and `cluster` run immutable versioned bundles; `skill` bind mounts its working directory instead,
-so you can edit and iterate fast while building the plugin bundle.
+All three tiers run an immutable bundle snapshot: `local` and `cluster` assign that snapshot a
+version, while `skill` identifies it by its content digest. What makes `skill` the fast loop is that
+it packs and boots that snapshot straight from your working directory in one command, with no
+platform in front.
 
 Curie provides an environment guarantee while climbing the three tiers. It is not a behavior
 guarantee: production traffic can still behave differently than your test cases, and no platform can
@@ -100,8 +102,9 @@ curie skill up
 curie skill message "hello, are you there?"
 ```
 
-A real reply streams back: no Slack, no platform yet. Edit `skills/my-agent/SKILL.md` and re-run to see your
-change answered. When done run the following command
+A real reply streams back: no Slack, no platform yet. `skill up` runs an immutable snapshot of the
+bundle, so after you edit `skills/my-agent/SKILL.md` the change only reaches a running runner once you
+restart it with `curie skill up --replace`. When done run the following command
 
 ```bash
 curie skill down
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -173,7 +195,10 @@ curie cluster down --yes
 
 **4. Ship it: your CI/CD**
 
-Connect this bundle's repo once (see [`docs/operations.md`](docs/operations.md#automatically-with-git-flow)), then:
+Git-flow needs the agent created with `--repo`, as step 3 showed. If you tore the cluster down at the
```
athusoo, line 205: In production we have to make sure that the database is not tied to the cluster - otherwise we will end up loosing logs etc. Ideally, along with the kubernetes cluster, we should also be pointed to a database and S3 bucket for persistent storage. How will that work?

athusoo (approved):

README.md
TheConnMan: Agreed. The quick start uses a disposable in cluster data stack, not the production topology. For production, set `postgres.deploy: false` and `minio.deploy: false`, point the chart at managed Postgres and an S3 compatible store, and provide credentials through existing Kubernetes Secrets. The chart redirects consumers, including bundle fetches, to those stores. I added this guidance and linked the complete configuration reference.

Issue #1076, opened 2026-07-29
allow_free_text=True is hardcoded: every approver is now forced through a modal with no opt-out, and the immediate action-id pair is dead code defended by a false comment (#1059)


TheConnMan:
#1059 (`37466904`) hardcodes `allow_free_text=True` on the approval `Confirm` intent, so every approver in every deployment is now routed through a two-step modal. There is no config, env var, or per-agent toggle. The PR is titled "Collect an optional note when an approval is approved or rejected"; the note is optional, the modal is not.

## Cause

`apps/worker/src/curie_worker/kernel.py:1090`:

```python
allow_free_text=True,
```

`kernel.py:1075` is the **only** production `ConfirmIntent(` constructor in the repo — `grep -rn "ConfirmIntent(" apps/ packages/ --include=*.py | grep -v tests` returns exactly it plus the model definition — and `kernel.py:1093` is the only `_sink.post` call. So `slack_sink.py:299` always passes `True` into `blocks.py:347-348`:

```python
approve_action = APPROVE_NOTE_ACTION_ID if allow_free_text else APPROVE_ACTION_ID
reject_action = REJECT_NOTE_ACTION_ID if allow_free_text else REJECT_ACTION_ID
```

A branch with exactly one reachable arm.

## Two consequences

**1. An unannounced product-wide UX change.** Every approver now takes an extra click and a modal on every approval, shipped inside a PR whose stated scope is collecting an optional note. The PR body anticipates the objection ("Always-on for approvals, which costs one extra click for an approver who does not want to leave a note") and names `allow_free_text` as "the single field to flip if it ever needs to be per-agent" — but nothing exposes that flip, so the decision is currently unrevertible without a code change.

**2. The immediate action-id pair is now dead for newly-posted cards, and the comment defending it describes a population that does not exist.** `apps/dispatcher/src/curie_dispatcher/handlers.py:279-282`:

```python
# The note-collecting variants (#1053): a click OPENS a dialog and resolves
# nothing; the submission below does the resolving. A card carries this pair
# only when the kernel emitted its Confirm intent with `allow_free_text`, so
# the immediate pair above stays the behavior for every other card.
```

[...]

Issue #1073, opened 2026-07-29
Resolving an unrouted approval wipes the card's summary in Slack: conversations.history does not return thread replies (#1059)


TheConnMan:
After #1059 (`37466904`), resolving an approval on an **unrouted** card (the zero-config default) replaces the card in Slack with a bare one-line verdict. The header, the tool-call summary, and the "Requested by" line are all destroyed.

## Cause

`apps/worker/src/curie_worker/kernel.py:1090` sets `allow_free_text=True` unconditionally, so every card renders the note buttons and resolves through `process_note_submission`. A `view_submission` payload carries no message, so that path re-reads the card:

`apps/dispatcher/src/curie_dispatcher/approval_actions.py:376-390`

```python
history = web_client.conversations_history(
    channel=channel, latest=card_ts, oldest=card_ts, inclusive=True, limit=1
)
```

`conversations.history` does not return **thread replies**. Slack's own reference says "To retrieve a message from a thread, check out `conversations.replies`", and the documented `latest`+`oldest`+`inclusive`+`limit=1` single-message recipe operates over the channel timeline.

The default card **is** a thread reply: with no route bound, `card_channel == qevent.reply_handle.channel` (`kernel.py:961`), so `in_requesting_channel` is true (`kernel.py:1060`) and the card posts with `thread_ts=thread` (`kernel.py:1097`). `handlers.py:95` sets `thread_ts = event.get("thread_ts") or event["ts"]`, so it is never `None` — for channel mentions and DMs alike.

So `message` is `{}`, and `_resolved_card_blocks({}, verdict)` (`approval_actions.py:160-177`) returns only the verdict context block, which is written over the live card at `:459-464`.

## Reproduction

Calling `process_note_submission` directly with a `conversations_history` that returns `{"messages": []}`:

```
text fallback: 'Approved by <@U_MANAGER>\nNote: looks fine'
blocks: [ {"type":"context","elements":[{"type":"mrkdwn","text":"Approved by <@U_MANAGER>\nNote: looks fine"}]} ]
```

With the card present, all four original blocks are retained and the verdict is appended. No guard in `_resolve_and_render` skips `chat_update` on an empty `message`. `chat_update(text=verdict, ...)` also overwrites the notification fallback, so the summary is lost from previews and screen-reader text. The same wipe exists on the 409 race path via `_refresh_settled_card` (`:556-561`).

## Failure scenario

[...]

Issue #419, opened 2026-07-15
Expiry-aware Slack approval card update


TheConnMan:
## Problem

When an approval expires (via the sweeper or the resolve-path expiry, #412), the Slack Block Kit approval card's Approve/Reject buttons remain live-looking. A click on an expired card gets a 410, but the card itself is never updated to reflect the expiry, so the approver sees no signal that the window closed.

## Fix direction

On expiry, update (or disable) the posted approval card to show it expired, mirroring how a resolved card is updated. This is worker/Slack-side work (the API owns the record; the worker owns the card surface).

## Notes
- Context: PR #417 (issue #412), ADR-0010. Deferred from #412 as it is out of the API-side sweeper's scope.

PR #1164, opened 2026-07-31
Settle the approval card when a resolution arrives through the API

### LINKED ISSUE #1084: Resolving an approval through the API leaves the Slack card live: stamping exists only on the dispatcher's click path (#1053 item 5)
## Why

Item 5 of #1053 asked for the approver channel's card to be stamped when an approval is resolved through the API rather than a click. PR #1059 deferred it, named the reason honestly (the dispatcher stamps from the click payload's original blocks, the worker would stamp from a remembered summary, and converging them is a design decision about a shared settled-card renderer), and offered to file the follow-up. This is that follow-up.

Today a resolution that arrives via `POST /approvals/{id}/resolve` or `curie <tier> approvals <AGENT> --resolve` leaves the card in the approver channel showing live Approve/Reject buttons. The record is settled, the buttons still invite clicks, and every later click earns a 409. The worker already knows where the card is: `approval_cards.ApprovalCardStore` holds the ref, and the expiry teardown (#419) already uses it to disable the card when an approval times out. The resolve path is the one terminal transition that does not.

This also blocks a decision. Discussion #1061 asks whether the demo and testing pain behind the self-approval proposal (#1054) is really a policy problem or a plumbing problem: `--resolve --as` exists but bypasses the dispatcher, so the card is never stamped and the flow looks broken even when it worked. Until API-resolve stamps the card, that question cannot be answered from evidence.

One adjacent gap belongs here because the fix exercises the same path: the claim that a note reaches the requester rests on `resumequeue.build_resume_turn` interpolating `Note: {approval.resolution_note}` into the resume turn (`apps/api/src/curie_api/resumequeue.py`), and no test in the tree asserts that text. The interpolation is production behavior carrying approver-written content into a session, and it is pinned nowhere.

## What

1. A shared settled-card rendering, so the dispatcher's click path and the worker's resolve-driven resume path produce the same visible outcome. #1073 already forces a rework of the dispatcher-side settled render (the history re-read wipes in-thread cards); design the two fixes together rather than shipping two more divergent renderers on this seam.
2. The worker's resume path pops the card ref from `ApprovalCardStore` and stamps the card the way the expiry teardown already does: buttons removed, decision, resolver, and any note visible.
3. A test that pins the resume turn text containing `Note: <text>` through `build_resume_turn`, so the note-reaches-the-requester claim stops resting on an unexercised line.

## Constraints

- The resume path lives in the sacred kernel's neighborhood; keep the worker change inside the existing card-store and sink surfaces and flag it for the adversarial review `apps/worker/CLAUDE.md` requires.
- Idempotence: a click-stamped card must not be re-stamped by the resume path, and a resume-stamped card must make a late click render the existing already-resolved refusal.
- Progressive enhancement per ADR-0020: a channel with no card ref (expired store, non-Slack surface) resolves normally with nothing to stamp.

## Done when

- Resolving via `POST /approvals/{id}/resolve` stamps the card: buttons gone, decision and resolver shown, note shown when present.
- Resolving via `curie local approvals <AGENT> --resolve <id> --as <u> --note "..."` produces the same stamped card, so the CLI flow and the click flow converge on one visible outcome.
- A test asserts the resume turn text contains `Note: <text>` for a noted resolution and omits the clause for a bare one.
- The in-thread (unrouted) card case is covered, not just the routed-channel case, given #1073.
- Hands-on pass at the local tier against a real workspace, transcript in the PR.

## Related

Deferred from #1053 (item 5) by #1059. Blocks the evidence question in discussion #1061 and therefore sequences before #1054 implementation. Interacts with #1073 (settled-card render on the dialog path) and #1074 (note content handling). Background: ADR-0020, #419 (expiry teardown, the pattern to follow).

yichenzhang-curietech:
Closes #1084. Third and last code PR of the Wave 2 settled-card sequence, after #1154 (#1076, the always-on decision) and #1157 (#1073, the thread read).

## What this fixes

A resolution that arrived through `POST /approvals/{id}/resolve` or `curie <tier> approvals --resolve` left the card in the approver channel showing **live Approve/Reject buttons**. The record was settled; the buttons still invited clicks; every later click earned a 409.

Stamping existed only on the dispatcher's click path, and that path never runs for an API resolve. The worker is the only component that still knows where the card is — the `ApprovalCardStore` ref it wrote at pause time, which the expiry teardown (#419) already uses for exactly this. The resolve path was the one terminal transition that popped that ref and threw it away, on the premise that *"the dispatcher already did from the click"* — true only when there **was** a click.

## One renderer, and why it lives where it does

`settled_approval_card` is in the **dispatcher** package, because that is the direction the dependency already runs: `curie_worker.blocks` imports the action ids from `curie_dispatcher.approval_actions` for the same anti-drift reason, and the reverse import would be a cycle. The worker's `resolved_approval_card` is a thin adapter over it.

"The same visible outcome" is **asserted, not claimed**. `test_the_rebuilt_settled_card_matches_the_edited_one` renders a live card, settles it both ways — edited from the live blocks as the click path does, rebuilt from the remembered summary as the resume path does — and compares block for block:

```python
assert rebuilt == edited, (
    "the rebuilt settled card must be identical to the edited one, or a CLI "
    "resolve and a click leave different cards."
)
```

That equality is the contract. Without it, convergence is a docstring.

## The verdict is read, not parsed

The resume turn does state the decision, the resolver and the note — but it states them in a sentence written for a language model:

```
[approval resolved] The request "..." was approved by U9. Note: approved for Q3. Continue the task accordingly: ...
```

Rebuilding a decision by regex over that would make the card's correctness depend on prose wording. So the approval id comes off the resume turn's **deterministic event id** (`approval-<id>-resolved`, which `resumequeue` documents as frozen because the worker's done-marker dedupes on it), and the facts come from the record.

## Supporting changes

| Change | Why |
|---|---|
| `ApprovalCardRef.requested_by` | The rebuild shows the same "Requested by" line the live card did. **Defaulted and read with `.get`** — these refs live in Valkey across a deploy, so a pre-#1084 entry must still load; it renders without the line rather than failing. |
| `ApprovalClient.get` behind a new `ApprovalReader` protocol | Separate from `ApprovalCreator` so a test fake for the pause path does not have to grow a read it never calls. In production both parameters get the one client. **Never raises**: its only caller is best-effort teardown on a turn that is already resuming. |
| `SlackSink.update_message(settled=...)` | A channel-neutral value, not fields smuggled into `OutboundMessage` — whose `status` already means a rendered status line, and overloading it would give one field two meanings depending on which sink method received it. `None` keeps #419's expiry behavior byte-identical. |

## Three refusals to stamp, each tested

Progressive enhancement per ADR-0020, and the same asymmetry #1073 established: a card that keeps its buttons is honest about the platform not having told Slack yet, and the next click gets the real answer from the API. A card stating a verdict the kernel had to guess is not.

- No reader configured → resumes normally, stamps nothing, **reply still delivered**.
- Record cannot be read → no stamp, ref still consumed (a ref left behind would be popped by an unrelated later approval on the same thread).
- Record not in a resolved state → no stamp.

## Idempotence

`pop` is GETDEL, so exactly one resume turn per thread gets the ref and a redelivery finds nothing. Within that single pass the stamp may land on a card the dispatcher already stamped from a click — that is a rewrite to an **equivalent** card, not a second verdict, and the shared renderer plus the equality test are what make "equivalent" true rather than hoped. A card stamped here and then clicked late gets the existing already-resolved refusal from the API, unchanged.

## The gap the issue named

> the claim that a note reaches the requester rests on `resumequeue.build_resume_turn` interpolating `Note: {...}`, and no test in the tree asserts that text.

Correct — the only other reference builds a turn to seed a dead-letter row and never reads it. `apps/api/tests/test_resume_turn_text.py` (5 tests) pins the noted clause **exactly** (`" Note: approved for Q3."`, spacing and period included), its total absence on a bare resolution (an empty `Note: .` would tell the model the approver left a blank reason rather than none), the rejection wording, the resolver as author, and that an expiry names no decider and carries the do-not-perform instruction.

## One existing test changed meaning, deliberately

`test_resolve_resume_leaves_the_card_to_the_dispatcher` asserted `h.sink.card_updates == []` as **correct behavior**. It is now `test_resolve_resume_stamps_the_card_from_the_record`, and the docstring states what changed and why. Flagging it as the one inverted assertion in this PR.

Also: `FakeSink.update_message` now **records** `settled` rather than accepting and dropping it. A fake that swallowed it would let the resolve path regress to rendering an expiry while every assertion in the suite stayed green — the exact shape of the hole Brian's audit found in #1059's stub.

## Sacred module

`kernel.py` is touched. Flagging it for the adversarial review `apps/worker/CLAUDE.md` requires. The change stays inside the existing card-store and sink surfaces, adds no kernel state, and the teardown remains best-effort and gated on the resume event id so an ordinary turn pays nothing. `_finalize_expired_card` is renamed `_finalize_settled_card` because it now handles both terminal transitions.

## Verification

Dev-stack Valkey, Postgres **and MinIO** up (the api suite needs the bundle store):

```
uv run ruff check .                             All checks passed
uv run mypy                                     no issues in 166 source files
uv run pytest apps/worker apps/dispatcher       662 passed, 11 skipped
uv run pytest apps/api/tests                    493 passed, 9 skipped, 1 failed (see below)
curie dev docs-lint                             OK
```

The one api failure, `test_control_integration.py::test_cost_composes_metrics_for_the_agent`, **fails identically on a clean tree** (verified by stashing this branch's changes) and is untouched by this PR — nothing here goes near cost or metrics.

## Still outstanding for Wave 2

The **live Slack pass** over the route-binding and note-dialog surfaces, which Brian asked for when #1073 and #1076 were taken. It is a human-in-the-loop run (a modal click cannot be synthesized over Socket Mode) and is scheduled separately. It should also confirm the `views.open` scope assumption in the app manifest, whose interactivity comment currently names only `block_actions` and not `view_submission`.

### DIFF
--- a/apps/api/tests/test_resume_turn_text.py
+++ b/apps/api/tests/test_resume_turn_text.py
@@ -1,0 +1,112 @@
+"""The platform-authored resume-turn text (#1084).
+
+``build_resume_turn`` is the one place an approver's note crosses from a durable
+record into a model's context. The claim that "the note reaches the requester"
+rests entirely on the interpolation in this function, and nothing in the tree
+asserted it: the only other reference builds a turn to seed a dead-letter row
+and never reads its text.
+
+That is worth its own file rather than a passing assertion elsewhere. The text
+is a contract with two consumers at once -- the model reading it as platform
+voice, and a skill matching on its ``[approval resolved]`` prefix -- so a
+wording change is a behavior change, and it should fail here rather than in a
+deployment.
+
+Pure unit tests over detached ORM instances: no session, no Valkey, nothing to
+skip on.
+"""
+
+from __future__ import annotations
+
+import uuid
+
+from curie_api.models import Approval, ApprovalStatus
+from curie_api.resumequeue import build_expiry_resume_turn, build_resume_turn
+
+
+def _resolved(
+    *,
+    status: str = ApprovalStatus.approved,
+    resolved_by: str | None = "U_MANAGER",
+    note: str | None = None,
+    summary: str = "Give ACME a 20% discount",
+) -> Approval:
+    """A detached, resolved ``Approval``: only the fields the builder reads."""
+
+    return Approval(
+        id=uuid.uuid4(),
+        agent_id=uuid.uuid4(),
+        conversation_id="C1-thread-0",
+        author="U_AE",
+        summary=summary,
+        reply_channel="C1",
+        reply_placeholder="p-1",
+        reply_endpoint=None,
+        dedupe_key="ev-1",
+        status=status,
+        resolved_by=resolved_by,
+        resolution_note=note,
+    )
+
+
+def test_a_noted_resolution_carries_the_note_into_the_resume_turn() -> None:
+    turn = build_resume_turn(_resolved(note="approved for Q3"))
+
+    # The exact clause, not just the substring: the space before and the period
+    # after are what keep it a sentence rather than text jammed against the
+    # attribution.
+    assert " Note: approved for Q3." in turn.text
+    # And the facts around it, so a rewrite cannot drop one while keeping the note.
+    assert turn.text.startswith("[approval resolved]")
+    assert '"Give ACME a 20% discount"' in turn.text
+    assert "was approved by U_MANAGER" in turn.text
+
+
+def test_a_bare_resolution_omits_the_note_clause_entirely() -> None:
+    """No note means no ``Note:`` at all, not an empty one.
+
+    An empty clause would read as "Note: ." to the model, which is a statement
+    that the approver left a blank reason rather than none.
+    """
+
+    turn = build_resume_turn(_resolved(note=None))
+
+    assert "Note:" not in turn.text
+    assert "was approved by U_MANAGER" in turn.text
+
+
+def test_a_rejection_says_rejected_and_still_carries_its_reason() -> None:
+    """The clause is decision-independent, and a rejection is where it matters
+    most: "rejected" without a reason is the half of the decision the requester
+    cannot act on."""
+
+    turn = build_resume_turn(
+        _resolved(status=ApprovalStatus.rejected, note="discount exceeds policy")
+    )
+
+    assert "was rejected by U_MANAGER" in turn.text
+    assert " Note: discount exceeds policy." in turn.text
+
+
+def test_the_resume_turn_is_authored_by_the_resolver() -> None:
+    """The author is the resolver, which is what lets the worker attribute a
+    settled card without parsing the prose (#1084)."""
+
+    turn = build_resume_turn(_resolved(note="fine"))
+    assert turn.author == "U_MANAGER"
+
+
+def test_an_expiry_turn_names_no_decider_and_no_note() -> None:
+    """The sibling path: nobody decided, so there is nobody to attribute and
+    nothing to quote. Asserted here so a shared edit to the two builders cannot
+    leak an approver clause into the expiry wording."""
+
+    turn = build_expiry_resume_turn(
+        _resolved(status=ApprovalStatus.expired, resolved_by=None, note=None)
+    )
+
+    assert turn.text.startswith("[approval expired]")
+    assert "Note:" not in turn.text
+    assert turn.author == "system"
+    # The instruction that makes an expiry safe: do not perform the gated action.
+    assert "Do not perform the gated action." in turn.text
--- a/apps/dispatcher/src/curie_dispatcher/approval_actions.py
+++ b/apps/dispatcher/src/curie_dispatcher/approval_actions.py
@@ -85,6 +85,13 @@ _VERDICT_LINE_MAX = 2900
 # verdict's context-block clamp above no longer covers it.
 _FALLBACK_TEXT_MAX = 39000
 
+# The live card's header, shared with the settled rebuild so the two cannot
+# drift: a rebuild under a different heading is a visibly different card for the
+# same decision. ``curie_worker.blocks.approval_card`` imports this rather than
+# repeating the literal.
+_APPROVAL_CARD_HEADER = "Approval required"
+APPROVAL_CARD_HEADER = _APPROVAL_CARD_HEADER
+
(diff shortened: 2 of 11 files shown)


yichenzhang-curietech:
CI note: the first run failed one job, `Python (ruff + mypy + pytest)`, on `apps/worker/tests/eval/test_recorder.py::test_records_per_case_results_and_reads_them_back` with `assert None == 1.0`. Re-running it.

I do not think it is this PR, and here is the evidence rather than the assertion:

- Nothing here touches the eval lane. `git diff --name-only` over this branch matches no path under `eval/`; the change is the approval card, the card store, the approvals client, the sink, and the kernel teardown.
- The test passes locally against the full compose stack (Langfuse + ClickHouse up), 6/6, run on this branch.

The failure shape also points at the test rather than the code. It polls for the async-ingested **traces** (40 attempts, 1s apart) and then reads back a **score** with no poll at all:

```python
for _ in range(40):
    traces = await _traces_for_version(client, version)
    if len(traces) >= 2:
        break
    await asyncio.sleep(1)
assert len(traces) == 2, ...
...
for trace in traces:
    assert await _score_for_trace(client, trace["id"]) == expected   # no wait
```

Langfuse ingests scores separately from traces, so a trace can be queryable before its score is. `assert None == 1.0` is exactly that gap: the trace assertions above it passed, and only the score read came back empty.

Worth saying plainly: this is the **only** failing CI run in the last 25, so "known flaky" is not something I can claim from history. If the re-run is green I would rather file the polling gap as its own issue than treat one green re-run as proof, since the race is visible in the test source regardless of how often it fires. Happy to do that, or to fix it inline here if reviewers would prefer it not linger.

yichenzhang-curietech:
Re-run is green: **26/26**.

As said above, I am not treating one green re-run as proof, so the polling gap is filed on its own as #1165 rather than left to be rediscovered. It is not fixed here: this PR does not touch the eval lane, and folding an unrelated test fix into it would make the diff harder to review for no benefit.

Issue #1183, opened 2026-07-31
curie local up --local-model hangs


athusoo:
### Before you file

- [x] I searched existing issues and did not find this one.
- [x] This is a bug, not a usage question.

### Version

0.5.0

### Which tier were you running?

local

### Platform

macOS arm64

### What happened

The command `curie local up --local-model` just hangs

### Steps to reproduce

Run 

```bash
curie local up --local-model
```

### Relevant output or logs

```shell

```

### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://github.com/curie-eng/curie/blob/main/CODE_OF_CONDUCT.md).

Issue #1165, opened 2026-07-31
Eval recorder test reads a Langfuse score without polling, while polling the trace it belongs to


yichenzhang-curietech:
`apps/worker/tests/eval/test_recorder.py::test_records_per_case_results_and_reads_them_back` polls Langfuse for the async-ingested **traces** and then reads back a **score** with no wait at all. Traces and scores ingest separately, so a trace can be queryable before its score is, and the test fails with `assert None == 1.0`.

## The gap, in the test's own source

```python
# Poll for the async-ingested traces (keyed by the unique version tag).
traces: list[dict[str, Any]] = []
for _ in range(40):
    traces = await _traces_for_version(client, version)
    if len(traces) >= 2:
        break
    await asyncio.sleep(1)
assert len(traces) == 2, f"eval traces did not materialize for {version}: {traces}"

...

# Each trace carries its eval_pass score matching pass/fail.
for trace in traces:
    expected = 1.0 if (trace.get("metadata") or {}).get("passed") else 0.0
    assert await _score_for_trace(client, trace["id"]) == expected   # <- no poll
```

The author clearly knew ingestion is asynchronous: the 40-attempt loop above exists for exactly that. The score read did not get the same treatment.

## Observed

CI run [30594997403](https://github.com/curie-eng/curie/actions/runs/30594997403), on the branch for #1084:

```
>       assert await _score_for_trace(client, trace["id"]) == expected
E       assert None == 1.0
apps/worker/tests/eval/test_recorder.py:326: AssertionError
```

The trace assertions immediately above it passed, so the traces had materialized and only the score had not. A re-run of the same commit was green (26/26).

## Frequency, stated honestly

This is the only failing CI run in the last 25, so it is rare rather than known-flaky. Filing it on the source rather than the frequency: the race is visible by reading the test, and a green re-run is not evidence that it is gone. A rare flake in a lane nobody owns is worse than a frequent one, because the next person to hit it will spend the debugging budget on their own diff first.

## Suggested fix

Poll the score the same way the trace is polled, with the same bound, and fail with the trace id named. Roughly:

```python
async def _score_for_trace_eventually(client, trace_id, *, attempts=40):
    for _ in range(attempts):
        score = await _score_for_trace(client, trace_id)
        if score is not None:
            return score
        await asyncio.sleep(1)
    return None
```

Worth checking the file's other reads for the same shape while there; this was the one that fired, not necessarily the only one.

## Done when

[...]

PR #1231, opened 2026-08-03
Surface the diagnosis, not the usage block, as a failed step reason

### LINKED ISSUE #1230: Operator verbs surface a CLI usage-boilerplate line as the failure reason instead of the real error
## What happens

When a command run through `run_step` exits nonzero, the operator sees a one-line reason drawn from the captured stderr. That reason is picked as **the last non-empty line**, which for any CLI that prints a usage block on a bad invocation is the trailing boilerplate rather than the error.

Real transcript, captured while investigating #1183:

```
✗ `docker` failed: Run 'docker --help' for more information
Error: `docker` exited nonzero
```

The captured stderr actually was:

```
unknown flag: --profile

Usage:  docker [OPTIONS] COMMAND [ARG...]

Run 'docker --help' for more information
```

The real cause — `unknown flag: --profile` — is the **first** line, four lines above the one we print. Everything except the last line is dropped to `ui.plumbing`, which is a no-op unless `--debug`, so on a default run the operator is told nothing actionable and has to re-run with `--debug` to learn what went wrong.

This is the same failure mode #1183 is about (an operator verb that does not say what is happening), one hop further down: here the information *was* captured, and then the summary threw the useful part away.

## Why

Two independent copies of the same heuristic:

- `failure_reason` (`cli/src/ops.rs`), used by the `cluster down` teardown result
- an inline copy inside `run_step` (`cli/src/ops.rs`), used by every operator verb that shells out to `helm` / `kubectl` / `docker compose`

Both do:

```rust
stderr.lines().rev().map(str::trim).find(|l| !l.is_empty()).unwrap_or("command failed")
```

`failure_reason`'s own doc comment already calls the `run_step` line its "pure standalone sibling", so the duplication is known and acknowledged — it just means a fix has to land in both places or they drift.

Note that the naive inverse ("take the first line instead") is **not** a correct fix: `helm` routinely prints warnings *before* its `Error:` line, so first-line would regress helm and kubectl, which last-line gets right today.

## Proposed fix

Keep last-non-empty as the base rule, but discard a **trailing usage/help block** before applying it:

- if the stderr contains a `Usage:` line, only consider the lines before the last such line
- otherwise, drop a trailing `Run '<...>' ...` / `See '<...>' ...` help pointer line
- then take the last non-empty trimmed line of what remains

Safety property to preserve: if that leaves nothing, fall back to today's result rather than to the empty string, so the change can only ever improve the surfaced reason, never blank it.

And collapse the duplicate: `run_step` calls `failure_reason` instead of re-implementing it.

Expected outcome on the transcript above:

```
✗ `docker` failed: unknown flag: --profile
```

## Verification

Unit tests over the pure `failure_reason`, each proving the *current* code picks the wrong line (red before the fix):

- docker's `unknown flag` + `Usage:` + `Run '...' for more information`
- docker's `is not a docker command` + bare `See 'docker --help'`
- kubectl's `error: unknown flag` + `See '... --help' for usage.`
- **regression guards** that must stay green: helm `WARNING:` line followed by `Error: ...` (last-line is correct there), a single-line stderr, and an empty stderr.

Plus a test that `run_step`'s reason and `failure_reason` agree on the same input, so the two cannot drift again.

## Non-goals

- Streaming the subprocess output live, or bounding it with a timeout. Those are the other two halves of what #1183 turned up and belong in their own changes; this issue is only about the summary line thrown away at the end.
- Falling back to stdout when stderr is empty. Some tools do report errors on stdout, but that is a separate widening and would need its own justification.

yichenzhang-curietech:
Closes #1230.

## The bug

`run_step` reports one line of a failed command's captured stderr, picked as the last non-empty line. For any CLI that renders help on a bad invocation, that is the trailing boilerplate rather than the diagnosis. Real transcript from the #1183 investigation:

```
✗ `docker` failed: Run 'docker --help' for more information
Error: `docker` exited nonzero
```

The captured stderr was:

```
unknown flag: --profile

Usage:  docker [OPTIONS] COMMAND [ARG...]

Run 'docker --help' for more information
```

`unknown flag: --profile` is the only actionable line and it is the **first** one. The rest went to `ui.plumbing`, a no-op without `--debug`, so a default run told the operator nothing and cost a second run to learn anything.

## The fix

Last-non-empty stays the **base** rule on purpose. The naive inverse (take the first line) would be a regression: `helm` and `kubectl` print warnings *before* their `Error:` line, so their diagnosis genuinely is the tail. What changes is that trailing help text is cut before the base rule applies:

- everything from the last `Usage:` header onward
- a trailing `Run '...' --help` / `See '...' --help` pointer, matched on shape rather than on a tool name so a new tool needs no new marker

**Safety property:** a stderr that is *only* help text has no diagnosis to recover, so the cut falls back to the base rule's answer instead of an empty string. The surfaced reason can improve, never blank.

The rule lived in two copies, and `failure_reason`'s own doc comment already called the inline `run_step` one its "pure standalone sibling". `run_step` now calls `failure_reason`; there is one implementation.

## Proof

Five unit tests over the pure `failure_reason`. Verified **red before the fix**:

```
---- failure_reason_skips_a_trailing_usage_block ----
  left: "Run 'docker --help' for more information"
 right: "unknown flag: --profile"

---- failure_reason_skips_a_bare_help_pointer ----
  left: "See 'docker --help'"
 right: "docker: 'compos' is not a docker command."

test result: FAILED. 3 passed; 2 failed
```

The 3 that passed pre-fix are the regression guards (helm's `WARNING:`-then-`Error:` ordering, single-line stderr, empty stderr, and the only-help-text fallback) — they were green before and stay green, which is what makes them guards rather than new behavior.

Real invocation, same binary path, before and after:

```
before (main):   ✗ `docker` failed: Run 'docker --help' for more information
after  (branch): ✗ `docker` failed: unknown flag: --profile
```

Produced by pointing `DOCKER_CONFIG` at a config directory with no `cli-plugins`, which makes the top-level `docker` binary parse `--profile` and emit exactly the stderr shape above.

Local gates: `cargo fmt --check`, `cargo clippy --all-targets -- -D warnings`, and the full `cargo test` (610 lib + every integration suite, 0 failures).

## Scope

No command surface changed, so no manifest regeneration.

Two things deliberately left out, both noted as non-goals on the issue:

- **Streaming the subprocess output live, and bounding it with a timeout.** The other two halves of what #1183 turned up; each needs its own change and probably an ADR.
- **Falling back to stdout when stderr is empty.** A separate widening that needs its own justification.

One adjacent gap found while doing this, **not** addressed here and worth its own issue if you agree: the reason only reaches `ui.failure` (a human stderr line). The returned error is the generic ``  `docker` exited nonzero ``, so under `--json` an agent consumer gets no reason at all. Folding the reason into the `bail!` would fix that, but it also changes the string `exit::classify` matches on, which is load-bearing for retry classification — too much to smuggle into this PR.

### DIFF
--- a/cli/src/ops.rs
+++ b/cli/src/ops.rs
@@ -1429,16 +1429,58 @@ fn is_connectivity_failure(stderr: &str) -> bool {
     MARKERS.iter().any(|marker| lower.contains(marker))
 }
 
+/// A usage-block header, e.g. docker's `Usage:  docker [OPTIONS] COMMAND ...`.
+/// Matched on the `usage:` prefix only: a diagnosis line that happens to talk
+/// about usage (`usage limit exceeded`) never starts with the colon form, while
+/// every CLI that renders help on a bad invocation does.
+fn is_usage_header(line: &str) -> bool {
+    line.len() >= 6 && line[..6].eq_ignore_ascii_case("usage:")
+}
+
+/// A trailing help pointer a CLI appends instead of, or after, a usage block --
+/// docker's `Run 'docker --help' for more information`, kubectl's
+/// `See 'kubectl get --help' for usage.`. Matched on SHAPE (a `Run '`/`See '`
+/// opener that mentions `--help`), not on a tool name, so this does not need a
+/// new marker per tool.
+fn is_help_pointer(line: &str) -> bool {
+    let lower = line.to_ascii_lowercase();
+    let opens = ["run '", "run \"", "see '", "see \""]
+        .iter()
+        .any(|opener| lower.starts_with(opener));
+    opens && lower.contains("--help")
+}
+
 /// One-line reason drawn from a captured stderr: the last non-empty trimmed
-/// line, or a short default when the stderr is empty. Pure standalone sibling of
-/// the `run_step` failure line, so a failing teardown's Display message names
-/// WHY it failed instead of dropping the stderr to `--debug` plumbing.
+/// line that is part of the DIAGNOSIS, or a short default when the stderr is
+/// empty. The single implementation behind both a failing teardown's Display
+/// message and the `run_step` failure line, so neither drops the stderr to
+/// `--debug` plumbing and the two cannot drift apart.
+///
+/// Last-non-empty is the base rule because `helm` and `kubectl` print warnings
+/// BEFORE their `Error:` line, so the tail is where their diagnosis lives. What
+/// #1230 fixed is that a CLI rejecting its own invocation inverts that: the
+/// diagnosis comes first and a usage block comes last, so the raw base rule
+/// surfaced `Run 'docker --help' for more information` and threw away
+/// `unknown flag: --profile`. Trailing help text is therefore cut before the
+/// base rule is applied.
+///
+/// Safety property: a stderr that is ONLY help text has no diagnosis to
+/// recover, so the cut falls back to the base rule's answer rather than to an
+/// empty string. This can improve the surfaced reason, never blank it.
 fn failure_reason(stderr: &str) -> &str {
-    stderr
-        .lines()
+    let lines: Vec<&str> = stderr.lines().map(str::trim).collect();
+    let base = lines.iter().rev().find(|l| !l.is_empty()).copied();
+    // Everything from the last usage header on is the tool's own help text.
+    let diagnosis_end = lines
+        .iter()
+        .rposition(|l| is_usage_header(l))
+        .unwrap_or(lines.len());
+    lines[..diagnosis_end]
+        .iter()
         .rev()
-        .map(str::trim)
-        .find(|l| !l.is_empty())
+        .find(|l| !l.is_empty() && !is_help_pointer(l))
+        .copied()
+        .or(base)
         .unwrap_or("command failed")
 }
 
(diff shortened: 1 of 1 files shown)


yichenzhang-curietech:
## CI status: the two red checks are both unrelated to this diff, which touches one Rust file

**`Python (ruff + mypy + pytest)` — resolved, was the known #1165 flake.**

It failed on `apps/worker/tests/eval/test_recorder.py::test_records_per_case_results_and_reads_them_back` with `assert None == 0.0`, the same test and the same shape (a Langfuse score read without polling the trace it belongs to) that #1165 already describes. This PR changes no Python. Re-run of the failed job only: **pass, 4m21s**.

**`pnpm audit (apps/ui)` — a new advisory published today, red on `main` too.**

`GHSA-fxqj-rqcc-2cmp` (postcss, moderate) was published **2026-08-03 17:11:39 UTC**. The last green `Dependency Audit` on `main` ran at **13:08 UTC**, before it existed. Nothing in this PR touches `apps/ui`; the next push to `main` will go red the same way, so this blocks every open PR, not just this one.

- `postcss@8.5.19`, transitive-only via `vite@8.1.5`
- vulnerable `<=8.5.22`, patched `>=8.5.23`, latest `8.5.25`
- `vite@8.1.5` already declares `postcss: ^8.5.17`, so 8.5.25 is inside the range it accepts — no `vite` bump needed

`pnpm update postcss -r`, with or without `--depth Infinity`, does **not** lift it: pnpm leaves an existing transitive resolution pinned. A `pnpm.overrides` entry does, and I verified the result locally:

```
postcss@8.5.25   →  pnpm audit: No known vulnerabilities found
apps/ui/package.json   |  5 +++++
apps/ui/pnpm-lock.yaml | 11 +++++++----
```

The lockfile hunk is only the postcss version; nothing else moves. `pnpm lint` clean, `pnpm test` 24 files / 143 tests passed, `pnpm build` succeeds.

Deliberately **not** included here — it is an unrelated dependency bump and does not belong riding along in a PR about a Rust error-summary line. Happy to send it as its own one-commit PR.

PR #1235, opened 2026-08-03
Refuse --local-model (when required ollama asset is missing) rather than downloading 11.4 GB implicitly (ADR 0093)

### LINKED ISSUE #1183: curie local up --local-model hangs

yichenzhang-curietech:
Closes #1183. Implements [ADR 0093](https://github.com/curie-eng/curie/blob/task/1183-local-model-preflight/docs/adr/0093-local-model-assets-are-pre-provisioned-never-implicitly-downloaded.md), added as `Draft` in the first commit.

> ⚠️ **Process note.** `docs/adr/AGENTS.md` step 4 is "begin implementation only after acceptance", and 0093 is a Draft. The two commits are split so the ADR can be reviewed, and flipped to `Accepted`, independently of the code. If you'd rather follow the letter of the procedure, the ADR commit can land alone first and I'll re-open the implementation against an Accepted 0093 — say the word.

## It was never a hang

`--local-model` fetches **~11.4 GB** before it can start anything, and announces none of it:

| asset | size | fetched by |
|---|---|---|
| `ollama/ollama:0.24.0` image | **8.93 GB** | the host docker daemon, as a `compose up` side effect |
| `qwen3:4b` weights (the default) | **2.5 GB** | the `ollama-pull` one-shot container |

Measured end to end on macOS arm64 over a ~30 MB/s link:

| state | wall clock | bytes curie printed |
|---|---|---|
| image cold, model cold | **232s** | **0** |
| image warm, model cold | 137s | 0 |
| both warm | **17.9s** | normal |

On a 50 Mbit/s link the cold case is roughly half an hour behind a spinner reading `starting dev stack`. There is also no upper bound: `--wait-timeout 15` against the cold image pull was measured still running **135s and exiting 0**, because that flag covers compose's readiness wait, not the pull before it.

The issue's "relevant output or logs" block is empty. That is the symptom, not an omission — there was nothing to paste. During the investigation a genuinely wedged `docker-credential-desktop` produced a byte-for-byte identical screen and never returned at all, which is the part an operator cannot be asked to distinguish.

## What changes

`--local-model` requires both assets to be present, and refuses before bringing anything up when either is not:

```console
$ curie local up --local-model
Error: local model assets are not on this machine, and curie does not download them implicitly:
  - docker image  ollama/ollama:0.24.0  (~8.9 GB)
  - model         qwen3:4b  (size depends on the model; the qwen3:4b default is ~2.5 GB)
fetch them now with:
  curie local up --local-model qwen3:4b --pull-model
a first fetch can take ~30 min on a 50 Mbit/s link; once both are cached a re-up is seconds
```

`--pull-model` accepts the download for that run. With both cached, nothing changes — same compose invocation, same 17.9s.

The preflight is strictly offline and escalates only as far as it must: `docker image inspect` (~32ms) settles the image, a missing volume settles the model without starting anything, and only a machine that has both pays for the one probe that reads the volume — a throwaway container off the Ollama image just shown to be present (~2s).

`skill up --local-model` answers identically per ADR 0041; only its volume (`<container>-ollama-data`) and fix hint differ. `cluster up --local-model` is untouched: its model is the in-chart inference Deployment, so there is nothing host-side to provision and no flag.

## Proof it actually refuses

AGENTS.md asks a guard to be proven by execution, not by assertion. All four states, real binary, real docker:

| case | state | result |
|---|---|---|
| **A** | image cached, `qwen3:0.6b` not | refuses in **0s**, lists **only** the model — not the 8.9 GB image the machine already has |
| **B** | both cached | **18.1s**, unchanged, stack up |
| **C** | image absent | refuses, lists **both** assets |
| **D** | `--pull-model`, model absent | no refusal; `curie-ollama-pull-1:running` — it reached the actual pull |

Case C removed `ollama/ollama:0.24.0` after re-tagging it aside, so the layers survived and the tag was restored afterwards — the missing-image path is exercised without an 8.93 GB re-download.

Case A vs D is the sharpest contrast: identical command, one errors at 0s, the other is still pulling 15s later.

## Tests

Seven new unit tests over the pure parts:

- `model_manifest_path_mirrors_ollama_layout` — the manifest path **is** the "already pulled?" test, so getting it wrong means a false refusal or the implicit download coming back. Verified against a real volume: `qwen3:4b` really does live at `models/manifests/registry.ollama.ai/library/qwen3/4b`. Covers bare names, no-tag (`latest`), namespaced, and self-hosted-registry refs.
- `model_manifest_path_does_not_mistake_a_host_port_for_a_tag` — splitting on the first colon instead of the last would look for a model named after the host, always miss, and refuse a machine that has the model.
- `missing_assets_message_names_only_what_is_missing_and_how_to_get_it` — including both mirror cases, because listing a cached asset would send the operator to re-fetch 8.9 GB they already have.
- `missing_assets_message_never_hands_out_a_raw_docker_command` — this message is the most tempting place in the codebase to break CLAUDE.md's single-surface rule.
- `compose_ollama_volume_and_image_match_the_compose_file` — **a drift guard, not a unit test.** The preflight names compose's volume and image from Rust, so they are two sibling definitions of one fact. A renamed volume or bumped tag without this constant following would probe an empty volume (refusing a machine that has the model) or inspect an image compose never runs (waving through one that does not). Same guard shape as the `endpoints_match_compose_file` directly above it.

Local gates: `cargo fmt --check`, `cargo clippy --all-targets -- -D warnings`, full `cargo test` (**37 suites, 622 lib tests, 0 failures**), `curie dev field-parity`, `curie dev emit-parity`, `bash scripts/check-docs.sh`.

## Docs

- **`docs/local-model.md`** — a new "First run: the assets are not downloaded for you" section with the sizes, the refusal, `--pull-model`, and the measured cold/warm numbers. Also **fixes an existing gap**: the page documented model sizes but never mentioned the 8.93 GB image at all, which is the larger half and the one whose download phase has no container to inspect. The volume name is corrected to `curie_ollama_data` (it said `ollama_data`, which is the compose declaration, not the created volume).
- **`cli/README.md`** — `--local-model` and `--pull-model` documented on both the `skill up` and `local up` rows. `--local-model` was previously undocumented in that table.
- Command manifests regenerated (`cli/command-manifest.json` and the committed `apps/ui/src/generated/commandManifest.ts`).

## Scope

`local rebuild --local-model` is deliberately not preflighted: it runs `--no-deps <service>`, so it never brings up `ollama`/`ollama-pull` and is not a download vector.

Still open, and explicitly out of scope in the ADR: a first `curie local up` with **no** `--local-model` still pulls ~9 other images silently and unboundedly, and any of them can wedge forever. Bounding operator verbs in wall clock and streaming their subprocess output are the other two halves #1183 turned up; they want their own issues and probably one shared ADR.


yichenzhang-curietech:
## CI: 24/26 pass. Both red checks are dependency audits that are **already red on `main`**

Everything this PR can affect is green — `Rust (fmt + clippy + test)`, `Python (ruff + mypy + pytest)`, `UI`, all three `E2E parity ladder` rungs, every image build, `Compose`, `Contracts`, `CodeQL`, `gitleaks`, `cargo audit`, and `Commit messages`.

The two failures are the `Dependency Audit` workflow, and it is failing repo-wide, not here:

```
failure  main                              2026-08-03T21:37:04Z
failure  main                              2026-08-03T21:08:16Z
success  main                              2026-08-03T13:08:10Z   <- last green
```

Two independent advisories both published today, after that last green run. This PR touches one Rust module, two docs and the generated manifests; it cannot cause either.

### `pip-audit` — `cryptography` 49.0.0, CVE-2026-69247

**Not reachable in this codebase.** The advisory is the `pkcs7_decrypt_der` / PEM / S/MIME Bleichenbacher oracle fixed in `cryptography` 50.0.0 (released 2026-07-31): distinguishable errors and timing when unwrapping a `RecipientInfo`'s `encryptedKey`, exploitable by a caller that **decrypts untrusted PKCS#7 messages**.

Curie decrypts no PKCS#7 and never imports the module:

- `cryptography` is not a direct dependency of any workspace package. Its only path in is `pyjwt[crypto]>=2.10` (`apps/api/pyproject.toml`).
- The one production use of that is `apps/api/src/curie_api/github_app.py` — `jwt.encode(..., algorithm="RS256")`, RSA **signing** for the GitHub App identity (ADR 0092). Signing does not touch the PKCS#7 decrypt path.
- `grep -rn 'pkcs7\|smime'` across `apps/`, `packages/` and `runner/` returns nothing. The only direct `cryptography` imports anywhere are `serialization` + `rsa` in `apps/api/tests/test_github_app.py`, generating a test key.

So the exposure is nil, but the gate is deliberately fail-closed on *any* advisory rather than doing reachability analysis, and the bump is the right move regardless. **Dependabot #1232 already does exactly it** (49.0.0 → 50.0.0); #1234 also touches the Python manifests. Merging either clears this job. Our RS256 usage is unaffected by 50.0.0's FFDH deprecation.

### `pnpm audit` — `postcss` 8.5.19, GHSA-fxqj-rqcc-2cmp

Published 2026-08-03 17:11:39 UTC, also after main's last green run. Diagnosed in detail on #1231: transitive-only via `vite@8.1.5`, vulnerable `<=8.5.22`, patched `>=8.5.23`.

Worth flagging that **no Dependabot PR exists for this one and none will**: the resolution is pinned transitively, and `pnpm update postcss -r` does not lift it even with `--depth Infinity`. It needs a `pnpm.overrides` entry, verified locally as a 5-line `package.json` addition plus a one-version lockfile hunk, after which `pnpm audit` is clean and `pnpm lint` / `pnpm test` (24 files, 143 tests) / `pnpm build` all pass.

Both are unrelated dependency work and are deliberately not riding along in this PR.

Issue #27, opened 2026-07-07
Epic: channel pluggability beyond Slack


TheConnMan:
The communication channel is a swappable job (`docs/architecture-vision.md`, job 6), and the CLI already proves the service-level swap by minting the same wire payload with zero Slack contact. But the protocol itself is still Slack-shaped, so a second channel cannot exist yet.

High-level scope, to be broken into issues when picked up:

- [ ] Channel-neutral queue contract (#7) is the prerequisite and is filed as a ready issue
- [ ] Per-turn reply routing (#19) so multiple ingress paths coexist
- [ ] A second real channel adapter (ingress + ReplySink pair), chosen by actual demand: Teams, email, or a plain webhook channel
- [ ] Only then, if a third channel appears, consider a channel adapter registry

Deliberate restraint: do not build a multi-channel adapter framework before a second real channel exists; the second implementation teaches the interface.

---

## Breakdown into issues (added 2026-07-10)

Note: the two prerequisite bullets are already filed as their own issues — channel-neutral queue contract #7 and per-turn reply routing #19 — and are **not** re-created here. The child below is the second real adapter that teaches the interface. A channel adapter registry remains deliberately deferred until a third channel exists.

- [ ] Channel pluggability: second real channel adapter (ingress + ReplySink) #263

PR #1276, opened 2026-08-04
Draft ADR-0095: channel-scoped memory lifecycle

TheConnMan:
Adds ADR-0095 as a Draft per ADR-0085: merging publishes it for discussion and does not authorize implementation.

The decision: channel memory is a third scope over the durable state store, beside per-agent memory (ADR-0025) and per-thread transcripts (ADR-0029), defined as one surface-agnostic lifecycle with Slack as the first adapter.

1. Surface interface: an opaque scope key on every inbound turn, plus an optional bootstrap trigger carrying an adapter-fetched backlog. Everything else is platform machinery, identical across surfaces.
2. Storage and injection: a `channel-memory` reserved namespace holding a curated markdown document, an append-only notes log, operator-only custom instructions (platform-key writes only, outrank memory), and a watermark. Injected as a boot preamble in the existing composition seam, under a hard budget enforced at the API state router.
3. Update: hybrid. Explicit saves land immediately as appended notes; a nightly scheduler enqueues per-scope compaction turns (silent, transcript-less turn kinds through the kernel) that fold notes in, correct stale facts, and advance the watermark under CAS.

Sequencing: ADR-0079 is Accepted but unimplemented; the new turn kinds and silent output path are sacred-kernel work behind it. Known limits (rebinding orphans memory, transcript-to-scope attribution as a #27 prerequisite) are stated in Consequences, and a validation gate lists the evidence required for acceptance.

ADR index regenerated via `scripts/check-docs.sh`, not hand-edited.

### DIFF
--- a/docs/adr/0095-channel-scoped-memory-lifecycle.md
+++ b/docs/adr/0095-channel-scoped-memory-lifecycle.md
@@ -1,0 +1,377 @@
+# 95. Channel-scoped memory: one surface-agnostic lifecycle for bootstrap, injection, and update
+
+Date: 2026-08-04
+
+Status: Draft
+
+Extends [ADR-0025](0025-memory-port-and-first-loader.md) (per-agent memory over
+the durable state store) and [ADR-0029](0029-conversation-history-port-and-first-loader.md)
+(per-thread transcripts over the same store) with a third scope: the place a
+conversation happens in. Designed against
+[ADR-0020](0020-message-port-rendering-free-channel-interface.md) (the
+rendering-free channel port) and [ADR-0012](0012-substrate-and-channel-agnostic-core.md)
+(the runner never learns its channel), and sequenced behind
+[ADR-0079](0079-inbound-triggers-as-a-new-event-kind.md) (inbound triggers as a
+new event kind), whose `source` field and placeholder-less output path this
+lifecycle builds on. Distinct from [ADR-0030](0030-proactive-within-episode-memory-agent.md)
+(within-episode memory), which addresses behavioral decay inside one run; this
+ADR addresses context continuity across runs in one place. Supersedes none.
+
+## Context
+
+Threaded work is the product's unit of execution: every mention boots a fresh
+session (ADR-0003, stateless-first) whose context today is the agent's
+cross-session lessons (ADR-0025), the thread's own transcript (ADR-0029), and
+the bundle's `systemPrompt`. Nothing carries what the *channel* knows: the
+running projects, conventions, decisions, and people that every thread in that
+channel implicitly assumes. A new thread either lacks that context or would
+have to re-read the channel backlog per turn, which does not scale in tokens
+or latency. The fix used by every comparable product is a per-place memory
+document: consume the channel once, keep a condensed durable note, inject it
+into every thread.
+
+Three facts about the current system shape the design:
+
+1. **The seams already exist.** Runner boot composes preambles in one place
+   (`runner/src/curie_runner/__main__.py`, `_compose_system_prompt`: agent
+   memory, then conversation, then the bundle prompt). The worker mints
+   memory and history refs in one place (`binding.py`, `boot_env`). Storage is
+   the durable state store (`WorkflowStateEntry`: agent-scoped, namespaced,
+   size-capped, CAS-versioned) with reserved namespaces `memory` and
+   `transcript`. A consolidation pipeline (dedup, provenance union, replace)
+   already exists in `runner/src/curie_runner/memory.py` but has no trigger;
+   this lifecycle gives its shape a caller.
+
+2. **"Per channel" and "per agent" are the same granularity today.** An agent
+   binds exactly one Slack channel (`agents.slack_channel`, 1:1; the
+   channel-ingress interface doc flags this as the largest remaining
+   Slack-shaped surface, and multi-channel is deliberately deferred to epic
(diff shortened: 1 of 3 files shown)


docs/adr/0095-channel-scoped-memory-lifecycle.md
```diff
@@ -0,0 +1,377 @@
...
+[ADR-0079](0079-inbound-triggers-as-a-new-event-kind.md) (inbound triggers as a
+new event kind), whose `source` field and placeholder-less output path this
+lifecycle builds on. Distinct from [ADR-0030](0030-proactive-within-episode-memory-agent.md)
+(within-episode memory), which addresses behavioral decay inside one run; this
+ADR addresses context continuity across runs in one place. Supersedes none.
+
+## Context
+
+Threaded work is the product's unit of execution: every mention boots a fresh
+session (ADR-0003, stateless-first) whose context today is the agent's
```
TheConnMan: Is this implemented? I don't recall us having this. This would need to follow the same life cycle of memory management and whatnot. This should just be a type of memory in addition to what we're talking about here. I think we need this ADR 95 to fully supersede ADR 25, the cross-session lesson.

docs/adr/0095-tiered-memory-lifecycle.md
```diff
@@ -0,0 +1,377 @@
...
+   (`runner/src/curie_runner/__main__.py`, `_compose_system_prompt`: agent
+   memory, then conversation, then the bundle prompt). The worker mints
+   memory and history refs in one place (`binding.py`, `boot_env`). Storage is
+   the durable state store (`WorkflowStateEntry`: agent-scoped, namespaced,
+   size-capped, CAS-versioned) with reserved namespaces `memory` and
+   `transcript`. A consolidation pipeline (dedup, provenance union, replace)
+   already exists in `runner/src/curie_runner/memory.py` but has no trigger;
+   this lifecycle gives its shape a caller.
+
+2. **"Per channel" and "per agent" are the same granularity today.** An agent
```
TheConnMan, line 64: Yeah down the line we should at least plan for this. In this ADR I didn't plan for the per agent but those are kind of global learnings across the channel. We have:

* thread context
* channel context
* agent context

That's the hierarchy and then we have system prompt.

docs/adr/0095-channel-scoped-memory-lifecycle.md
```diff
@@ -0,0 +1,377 @@
...
+   optimization, not a requirement of the interface.
+
+Injection, budget enforcement, and compaction are platform machinery and are
+identical across surfaces.
+
+### Storage
+
+- A new reserved state-store namespace, `channel-memory`, joins `memory` and
+  `transcript` in `RESERVED_NAMESPACES` (both the API and runner mirrors, per
+  the warning in `routers/state.py`). Per scope it holds four keys:
```
TheConnMan: This is all memory. I think it should still be under memory. We just have the different tiers: we have agent memory below that, channel memory, and I guess above all of it is system prompt but that's kind of in there. I feel like it should all be under memory. It's all memory.

athusoo (changes requested):
I have been doing some synthesis of latest literature on memory. I think just summarization is not going to work. We have to store the entire history and we have to be able to retrieve from that history. Otherwise accuracy will suffer. 

These are the two most relevant research from reputable sources -

https://arxiv.org/abs/2602.17913
https://arxiv.org/abs/2606.29778

athusoo:
Broadly, this is what the papers (TierMem) points to -


Accuracy (LoCoMo) | Tokens/query | Latency | Failure mode
-- | -- | -- | --
A — compact & delete | 0.755 | 260 | 1.24s | Irrecoverable. UOR 14.7–23.3% — once compacted, a dropped fact is gone forever, no matter how good future retrieval gets
C — compact, keep raw, router escalates | 0.851 (97.6% of ceiling) | ~3,980 (54% less than raw) | 6.76s (61% less than raw) | Router has false confidence 34.3% of the time; even correct escalation still fails 55.7% of the time
B — raw log + search tool, no compaction | 0.873 (TierMem ceiling) — and grep/SUMER studies show this can go higher (SUMER: +43% SOTA; PwC grep: up to 93.1%) | 7,398 (full cost) | 17.18s | Weak at reconciling updated facts (no structural "this supersedes that"); accuracy is harness/delivery-dependent, not free

Given this I think we should seriously look at the following - 

- keep all the raw logs of the conversations with time stamps etc and named in a way that the LLM can figure out which file to look at by just doing directory search.
- put a grep or key word search tool on these files
- put a tool to read file contents

And let this be the memory that is provided to the LLMs. This is the best for accuracy though it is not too great for latency - and the later we can handle by giving feedback to the user in case it is a chat type of a channel (for email we don't have to do any of that).

The token costs will only go down over time and we may not have to use very powerful models.

Also this strategy has worked quite well for us in the agents that we have designed.

TheConnMan:
@athusoo thanks for the note. My read from these (summarized) is that compaction alone is not good enough, we must have access to raw content. In the Slack case that means read access to the channel to search when necessary. Ideally, compacted content (perhaps with links to raw content) is sufficient. When it's not, though, we must provide access to raw data. That can get expensive, so it should only be done occasionally, but is a release valve for those use cases.

I updated the ADR. Does that match your expectation?

athusoo (approved):
I am good with this.

Issue #1239, opened 2026-08-04
Poll for new commits, so an unreachable cluster still deploys


jw3329:
## The problem

Curie is self-hosted: adopters run it in their own cluster. Many of those clusters are not reachable from the internet.

Today a deploy is triggered only by a GitHub webhook — an inbound request. A cluster behind a firewall or NAT cannot receive one, so git-flow simply never fires and the adopter has no push-to-deploy at all.

|                        | cluster behind a firewall |
|------------------------|---------------------------|
| GitHub webhook → Curie | cannot reach it           |
| Actions POST → Curie   | cannot reach it           |
| **Curie → GitHub**     | **works — plain egress**  |

Outbound always works; inbound often does not.

## Prior art

Argo CD — the closest analogue, a controller in your cluster that reads repos — **polls by default** and treats webhooks as an optimization for installs that happen to be reachable. Flux does the same. This is the settled answer for the shape Curie is.

## Shape

- A loop that, per agent's bound repository, asks GitHub for the current sha of the branches `deploy.yaml` names, and runs the existing deploy path when it moves
- Reuse the credential resolver from ADR-0092 (`GitHubCredentials.token_for`) — no new auth
- The webhook stays as the fast path; polling is the floor, not a replacement
- Interval in values, off or generous by default so it costs nothing on a reachable install

## Worth deciding in the issue

Whether polling is per-agent or per-repository. Several agents share a repository (ADR-0091), so per-repository is one API call instead of N.

## Done when

An adopter whose cluster accepts no inbound traffic can push to `dev` and see the dev agent update, with no webhook configured.

jw3329:
Done in #1246.

`api.commitPollIntervalSeconds`, off by default. The API asks GitHub whether the branches each `deploy.yaml` names have moved, and runs the ordinary deploy when they have.

It reuses `gitflow.process_push` rather than reimplementing deploy, and supplies the config-derived `clone_url` — the same value the origin check compares against — so a polled deploy passes that check by construction, not coincidence.

Per repository rather than per agent, since several agents share one repository (ADR-0091) and per-agent polling would race N deploys of the same commit. A failing repository doesn't stop the others. A restart doesn't redeploy current HEAD.

Not enabled on the acme-bot cluster: its webhook reaches it, so polling would only add API calls.

PR #1279, opened 2026-08-04
Onboard a repository in one command

jw3329:
Standing up an agent repository meant one `cluster deploy` per target, plus a trip to GitHub's webhook settings. Forgetting a target leaves an agent that exists and never updates — a failure with no error attached to it.

```bash
curie cluster deploy --all-targets
```

Deploys every target `deploy.yaml` declares. **Dev before prod**, so a run that fails part-way leaves prod on its previous version rather than ahead of a dev that never landed — recoverable in one direction only.

## The webhook step is gone, not automated

I started building a `curie cluster onboard` that would create the GitHub webhook for you. Then I realised commit polling (#1239, merged) makes the webhook **optional entirely**:

```yaml
api:
  commitPollIntervalSeconds: 60
```

Removing a setup step beats automating one. And automating it would have meant either widening the App beyond `Contents: Read` — giving Curie standing permission to rewrite webhooks on every repo it can see — or handling the operator's own GitHub token. Neither is worth it for something polling makes unnecessary.

So onboarding is now: **write `deploy.yaml`, run one command.**

## No second YAML parser

The target list comes from a new `POST /deploy-targets/list`, not from parsing `deploy.yaml` in Rust. ADR-0089 put the parser in the API precisely so the CLI and the validator cannot disagree about where a deploy lands, and the schema notes the Rust YAML ecosystem has no maintained `serde_yaml` successor to depend on.

I nearly added one anyway — caught it before writing the dependency.

Both endpoints now share one `_parse`, with a test asserting they reject an invalid file *identically*, so a second parser cannot creep back in through the new endpoint:

```
test_list_and_resolve_agree_on_an_invalid_file
```

## Verification

```
10 passed   deploy-target endpoints
37 test binaries pass, 0 failures   (Rust)
ruff + mypy clean
```

Behaviour checked by running it:

```
--all-targets with --target  -> error: cannot be used with '--target <TARGET>'
--all-targets, no deploy.yaml -> Error: --all-targets needs a deploy.yaml in the
                                 bundle, but ./deploy.yaml could not be read
```

Both committed manifests regenerated (`cli/command-manifest.json`, `apps/ui/src/generated/commandManifest.ts`), and the new mirror structs declared in `cli/api-mirrors.json` for the field-parity gate.

Refs #1239

### DIFF
--- a/apps/api/openapi.json
+++ b/apps/api/openapi.json
@@ -1669,6 +1669,21 @@
         "title": "KillState",
         "type": "object"
       },
+      "ListedTargets": {
+        "description": "Every target a ``deploy.yaml`` declares, dev before prod.\n\nOrdered so a caller onboarding a repository deploys dev first. A run that\nfails part-way then leaves prod BEHIND rather than ahead of a dev that\nnever landed -- recoverable in one direction, not the other.",
+        "properties": {
+          "targets": {
+            "default": [],
+            "items": {
+              "$ref": "#/components/schemas/NamedTarget"
+            },
+            "title": "Targets",
+            "type": "array"
+          }
+        },
+        "title": "ListedTargets",
+        "type": "object"
+      },
       "LoadPackConfig": {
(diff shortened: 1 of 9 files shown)


TheConnMan:
@jw3329 please fix the gitleaks issue. Due to a bug in the scanner, your failure here is blocking all other PRs.

PR #1281, opened 2026-08-04
Declare the arm64 Linux binary, so v0.6.0 can publish

jw3329:
**The v0.6.0 release build failed** at its final step, so the tag exists but nothing was published:

```
ERROR: file(s) staged for release but not declared in integrity.py:
curie-aarch64-unknown-linux-gnu
```

The gate was right to fire. The `cli-binaries` matrix builds **three** targets, and the workflow already generates an SBOM per target — but `BINARY_ASSETS` listed two.

## Why this one matters

The missing binary is `aarch64-unknown-linux-gnu` — the CLI for **arm64 Linux**, which is what a k3s-on-Graviton install needs. A release that shipped without it would be unusable on exactly the cluster shape this platform is most often self-hosted on.

## A comment is not a mechanism

`BINARY_ASSETS` carried this:

```python
# The CLI targets built by the `cli-binaries` job. Keep in lockstep with its matrix.
```

The matrix gained a target; the tuple didn't. The mismatch surfaced only on a tag, at the very end, after every image had already been built and pushed.

So the fix is two parts:

1. **Add the target.**
2. **Assert the two agree**, reading the matrix out of `release.yaml` rather than restating it. Test fixtures now mirror `BINARY_ASSETS` for the same reason, instead of listing binaries a third time.

Falsified — deleting the target reproduces the release failure in 20ms instead of at the end of a tagged build:

```
E  built but not declared: ['aarch64-unknown-linux-gnu']
```

```
33 passed
```

## To actually ship v0.6.0

Merge this, then re-run the release workflow on the tag (or re-tag). Nothing else in the run failed — images and manifests all built.

### DIFF
--- a/release/integrity.py
+++ b/release/integrity.py
@@ -41,6 +41,7 @@ SBOM_SUFFIX = ".spdx.json"
 # The CLI targets built by the `cli-binaries` job. Keep in lockstep with its matrix.
 BINARY_ASSETS = (
     "curie-x86_64-unknown-linux-gnu",
+    "curie-aarch64-unknown-linux-gnu",
     "curie-aarch64-apple-darwin",
 )
 COMPOSE_ASSET = "compose.release.yaml"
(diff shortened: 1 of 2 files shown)


TheConnMan:
@jw3329 please don't merge this, I'm actively fixing this.

TheConnMan:
Duplicates #1280.

PR #1285, opened 2026-08-04
Draft ADR-0096: port adapters are deployed services

TheConnMan:
Decides how a third party plugs its own implementation into any of the platform's swappable seams, motivated by the Slack-alike and specialized-email surface asks.

The five decisions:

1. Agent bundles and platform adapters are distinct plugin kinds that never merge; connectors are the decided third position (bundle-declared, platform-hosted, per ADR-0086/0094).
2. The default adapter shape is a deployed service speaking the port's versioned wire contract, selected by composition. What a third party ships is a declarative adapter manifest the platform composes from (a `curie adapter` verb family as the operator surface), so installing an adapter is a drop-in experience while the mechanism stays a service behind a wire.
3. A port is pluggable when its contract is published, versioned, drift-gated, and carries a conformance kit plus a manifest-schema entry. Trust is bounded in both directions: a reply endpoint never receives any credential other than its own, third-party ingress goes through the authenticated hook surface only, and a scoped credential for the interactivity return path precedes any second channel adapter.
4. In-process entry-point contributions (generalizing the `curie.harness` registry to `curie.<port>` groups) are the exception for ports that cannot cross a process boundary.
5. No generic framework now; ports are promoted one at a time on demand, channel first, with the CLI stub promoted to reference adapter.

Alternatives rejected: extending the bundle format with platform extensions, entry points everywhere, a universal registry across all eighteen seams now, and a go-plugin-style RPC protocol.

### DIFF
--- a/docs/adr/0096-port-adapters-are-deployed-services.md
+++ b/docs/adr/0096-port-adapters-are-deployed-services.md
@@ -1,0 +1,354 @@
+# 96. A third-party port adapter is a deployed service, not a loaded plugin
+
+Date: 2026-08-04
+
+Status: Draft
+
+Generalizes the harness trio to every port.
+[ADR-0060](0060-the-harness-is-a-declared-package.md) decides what a harness is
+(a declared package), [ADR-0061](0061-out-of-process-harness-boundary.md)
+decides where its boundary sits (a process, not a Protocol), and
+[ADR-0062](0062-harness-conformance-has-teeth.md) decides how we find out
+whether one works. Those three answer the question for one seam. ADR-0061 and
+ADR-0062 are both still Drafts, and 0061 is gated on a spike by its own decision
+5, so nothing here waits on either being accepted: decisions 2 and 3 below stand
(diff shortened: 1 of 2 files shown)


athusoo:
@TheConnMan - so in a nutshell we will have different microservices for slack, email, SMS etc. and anything pluggable that the user wants to stand up? Those will encompass how the adapter authenticates to the channel, how it gets messages from that channel, how it figures out which messages belong to the same thread etc. The actual dispatching via ValKey is done by the dispatcher. Is my understanding correct?

The adaptor only needs to be able to talk to the channel platform?

TheConnMan:
Almost. They're not microservices because they're not deployed separately. It's more like a library with implementations. We ship a variety of implementations, but you can write your own and plug it into the deployment. On boot, the system can use those custom implementations as long as they implement the interface. For a channel adapter, it comes in through the API's authenticated hook surface, never Valkey directly.

The channel is the most natural fit for this. There's a few other interfaces that can work like this as well (e.g. an S3 port instead of RustFS).

jw3329 (approved):

PR #1295, opened 2026-08-05
Shimmer by default, with a caption written for the surface it lands on

### LINKED ISSUE #1182: Openrouter calls take a long time vs anthropic calls
### Before you file

- [x] I searched existing issues and did not find this one.
- [x] This is a bug, not a usage question.

### Version

0.5.0

### Which tier were you running?

local

### Platform

macOS arm64

### What happened

curie local message calls that are going to Anthropic models return much faster than the ones that go to OpenRouter

### Steps to reproduce

1. First try Anthropic
```bash
export CURIE_CREDENTIALS=sk-ant-....
curie local up
curie local deploy --plugin-dir . --slack-channel C0123ABCD --api-url http://localhost:28000
curie local message "what is the weather like in new york?
```
on my computer this returned in 6.1s

```bash
curie local down
```

2. Then try GLM via OpenRouter
```bash
export CURIE_CREDENTIALS=sk-or-...
export CURIE_MODEL=glm-5.2

curie local up
curie local deploy --plugin-dir . --slack-channel C0123ABCD --api-url http://localhost:28000
curie local message "what is the weather like in new york?
```

this took 45s

```bash
curie local down
```

### Relevant output or logs

```shell

```

### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://github.com/curie-eng/curie/blob/main/CODE_OF_CONDUCT.md).

yichenzhang-curietech:
Refs #1182. Does **not** close it — see *Scope* at the bottom.

## The gap

During a turn, the only thing that moves in a Slack thread is the placeholder message, and it only moves when the model emits text. A reasoning model reasons *before* it answers, so the whole pre-token stretch is visibly frozen — and an operator cannot tell that from a wedge.

Measured on the path curie actually dials (OpenRouter's Anthropic-compatible endpoint), same prompt:

| model | wall clock | thinking tokens |
|---|---|---|
| `anthropic/claude-sonnet-4.5` | 2.4s | **0** |
| `z-ai/glm-5.2` | 8.1s | **268** (70% of output) |

That is not GLM being unusual — `deepseek/deepseek-r1` defaults to 259 and `qwen/qwen3-235b-a22b-thinking-2507` to 613. curie sets no thinking configuration at all; these are the models' own defaults. The reporter compared a model that reasons against one that doesn't and saw 6s vs 45s.

## What this does

Slack already has the affordance for exactly this: the **assistant-thread status**, the animated caption beside the app name ([`assistant.threads.setStatus`](https://docs.slack.dev/reference/methods/assistant.threads.setStatus)). Its own docs describe it as *"setting expectations for potentially slow responses."*

The plumbing has been in the tree since the shimmer landed, only switched off. This turns it on by default, and fixes the caption it carries.

**1. `CURIE_SHIMMER` defaults to on, on both services.** The dispatcher sets the caption; the worker is the only side holding the reply handle when the turn ends, so if it doesn't clear, nothing does. Slack's own auto-clear fires when the app **posts** a message, and this pipeline **edits** the placeholder instead, so that path never triggers. One env name drives both, and the two defaults now agree.

**2. The caption gets its own text.** It was being fed `placeholder_text`, but Slack renders a status as `<App Name> <status>` and inserts the app name itself, so the shimmer read:

```
Curie On it. Working on your request.
```

`CURIE_STATUS_TEXT` (default `is working on your request...`) splits the two surfaces. They have genuinely different grammar — the placeholder is a message body and stands alone, the caption is a fragment glued to the app name — so one string could never be right for both.

## Why defaulting it on is safe

The caption is **strictly additive**: it is not the message, it leaves no `(edited)` marker, and it notifies nobody. Live-edit streaming is untouched — the placeholder still streams as before, this just adds a second signal for the stretch where there is nothing to stream.

Where the feature isn't enabled, `setStatus` is rejected, the call is already best-effort (`except` → debug log), and the turn is unaffected. The operator sees no caption and nothing else changes.

The one thing that can't be automated is Slack's app-level **Agents & AI Apps** toggle — it has no manifest key. The `assistant:write` scope was already in the manifest; only the toggle is manual. The runbook now names it as a numbered setup step instead of an aside addressed to people who had opted into shimmer, because that framing went stale the moment the default flipped.

## Tests

Every production change is covered by a test that **fails without it**. Verified by reverting each one in turn:

| reverted | fails |
|---|---|
| `handlers.py` back to `placeholder_text` | `test_shimmer_sets_assistant_status_after_placeholder` |
| dispatcher default back to `False` | `test_defaults_parity_with_from_env`, `test_shimmer_is_on_by_default_with_no_explicit_config` |
| worker default back to `False` | `test_defaults_parity_with_from_env` (worker) |

New: `test_shimmer_is_on_by_default_with_no_explicit_config` drives the real dispatch path with an unmodified config — the existing test `model_copy`s the flag on, which no longer proves anything — and `test_status_is_cleared_by_default` is its worker-side mirror.

**Two test-quality fixes ride along, both consequences of the flip:**

- The `CURIE_SHIMMER` env-read cases (dispatcher and worker) asserted the value that is now the default, so they would have passed against an alias that was never read at all. Both now probe the non-default value.
- The two shimmer-*off* kernel tests leaned on the default instead of pinning the flag they exist to exercise. Both now pass `shimmer=False` explicitly. Their assertions are unchanged — only the fixture is, which is the point.

Gates: `ruff check`, `mypy` (176 files), `pytest apps/dispatcher/tests apps/worker/tests` (**743 passed, 17 skipped**), `bash scripts/check-docs.sh`. No kernel logic changed — the kernel edits are test fixtures and comments only.

## Docs

- **`docs/behavior-packs.md`** — a new section on where a load line actually appears, since load lines feed this same caption. Includes the `<App Name>` grammar rule as a ✅/❌ table, because a pack author writing `Crunching the numbers...` gets `Curie Crunching the numbers...`.
- **`docs/slack-local-runbook.md`** — the Agents & AI Apps toggle promoted to a setup step. The old paragraph ("skip both if you are not using shimmer") was stale as of this change.
- **`apps/dispatcher/README.md`** — `CURIE_STATUS_TEXT` and `CURIE_SHIMMER` added to the env table. `CURIE_SHIMMER` was missing from it entirely.
- One **pre-existing** stale line fixed: behavior-packs said the shimmer surface could draw from load lines *"once wired"*. It has been wired — `Kernel._set_shimmer` — while the other half it describes (seeding the placeholder **message**) genuinely is still deferred. Corrected to say which half is which.

## Scope

Deliberately **not** here:

- **Refreshing the caption on a timer.** Slack expires a status after ~2 minutes if nothing refreshes it, so a turn longer than that loses its caption partway through. The fix has to run inside the kernel while the turn awaits the runner, which makes it a sacred-module change with the escalated review — its own PR, not a passenger on a config-and-docs change.
- **The thinking/effort knob.** `ClaudeAgentOptions` exposes `thinking`, `effort` and `max_thinking_tokens`; curie sets none of them. Letting an operator turn the reasoning *down*, rather than only watch it happen, is what actually answers #1182's title, and it needs its own ADR (operator-owned, no bundle override) first. **That change closes #1182; this one only makes the wait legible.**

### DIFF
--- a/apps/dispatcher/README.md
+++ b/apps/dispatcher/README.md
@@ -73,6 +73,8 @@ Read from the environment by `DispatcherConfig()` (a `pydantic_settings.BaseSett
 | `CURIE_DEDUPE_PREFIX` | `curie:dedupe:` | dedupe key prefix |
 | `CURIE_DEDUPE_TTL_SECONDS` | `3600` | dedupe guard TTL |
 | `CURIE_PLACEHOLDER_TEXT` | `On it. Working on your request.` | placeholder reply text |
+| `CURIE_STATUS_TEXT` | `is working on your request...` | the shimmer caption, NOT the message. Slack renders it as `<App Name> <status>` and inserts the app name, so write it as a continuation, not a sentence |
+| `CURIE_SHIMMER` | `true` | set the Slack assistant-thread status (the animated caption by the app name) while a turn runs. Also read by the worker, which clears it when the turn ends. Needs the app's **Agents & AI Apps** feature enabled by hand; without it the call is rejected and swallowed best-effort |
 | `CURIE_BACKOFF_INITIAL_SECONDS` | `1.0` | first reconnect backoff |
 | `CURIE_BACKOFF_MAX_SECONDS` | `30.0` | backoff cap |
 | `CURIE_BACKOFF_MULTIPLIER` | `2.0` | backoff growth factor |
--- a/apps/dispatcher/src/curie_dispatcher/config.py
+++ b/apps/dispatcher/src/curie_dispatcher/config.py
@@ -19,6 +19,7 @@ Env mapping:
     CURIE_DEDUPE_PREFIX      -> dedupe_prefix
     CURIE_DEDUPE_TTL_SECONDS -> dedupe_ttl_seconds
     CURIE_PLACEHOLDER_TEXT   -> placeholder_text
+    CURIE_STATUS_TEXT        -> status_text (the shimmer caption, NOT the message)
     CURIE_SHIMMER            -> shimmer (assistant-thread status while working)
     CURIE_BACKOFF_INITIAL_SECONDS -> backoff_initial_seconds
     CURIE_BACKOFF_MAX_SECONDS     -> backoff_max_seconds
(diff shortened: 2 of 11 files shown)


TheConnMan:
Neat!

PR #1321, opened 2026-08-05
`curie apply` refuses to delete a stateful component

jw3329:
`curie diff` learned to warn when the deployed chart differs from the one the CLI would apply (#1306). **`apply` itself had no guard and would have gone ahead.**

## The case is live, not hypothetical

sre-bot runs chart **0.5.1**, whose object store is `minio`. Chart **0.6.0** renamed it to `rustfs` and renders no minio at all. `up` does a *full* upgrade, so applying would **delete the running StatefulSet**.

And that store is not incidental: every sandbox pod has a **`bundle-fetch` init container**, and a sandbox is created **per Slack thread**. Every `bundle_ref` in Postgres points into that store. So the failure isn't a broken rollback — **the next Slack message fails.**

## The guard

`apply` now reads the release's StatefulSets, renders the target chart with the *same effective values the upgrade would send*, and refuses when a live one is absent from the render:

```
refusing to apply: this would DELETE 1 stateful component(s) the release is
running, and the persistent data with them:
  sre-bot-minio
...
Re-run with --allow-stateful-removal once the data is migrated or you accept losing it.
```

Three deliberate choices:

- **StatefulSets specifically**, not every workload. They carry the PVCs, so they're the ones whose removal loses data rather than restarting a process. Guarding everything would make the flag reflexive and protect nothing.
- **Runs under `--dry-run` too.** The plan a dry run prints is exactly the plan that would destroy the store, so the reader deserves the same refusal.
- **Parses the render; does not grep it.** `kind: StatefulSet` appears inside ConfigMap payloads, and matching that would invent a component the chart never creates. There's a test for exactly that.

## Verification

Five unit tests on the pure decision — the rename case, an unchanged set, an added component, multi-doc parsing, and the ConfigMap false positive.

**Falsified:** stubbing the removal comparison to return nothing fails `a_renamed_store_is_reported_as_a_removal`.

`fmt`, `clippy --all-targets -D warnings`, 38 suites green, command manifest regenerated.

## What is *not* verified

**This has not run against the live cluster.** I built the linux binary and was staging it when the AWS session expired. The pure decision is covered; the `kubectl get statefulset` and `helm template` reads around it are not.

Given that every real bug this session came from running against the actual cluster rather than from tests, I'd treat that as a genuine gap — the guard is worth landing regardless, since today `apply` has no protection at all, but it deserves a live run before anyone trusts it.

### DIFF
--- a/apps/ui/src/generated/commandManifest.ts
+++ b/apps/ui/src/generated/commandManifest.ts
@@ -3788,6 +3788,18 @@ export const commandManifest = {
           "long": "chart",
           "positional": false,
           "required": false
+        },
+        {
+          "global": false,
+          "help": "Proceed even when the upgrade would DELETE a stateful component the release is running, losing its data. Refused by default: a chart that renames a store (minio -> rustfs in 0.6.0) drops the old one, and every sandbox reads the bundle store at start, so the next turn fails rather than merely a rollback",
+          "id": "allow_stateful_removal",
+          "long": "allow-stateful-removal",
+          "positional": false,
+          "possible_values": [
+            "true",
+            "false"
+          ],
+          "required": false
         }
       ],
       "hidden": false,
--- a/cli/command-manifest.json
+++ b/cli/command-manifest.json
@@ -3785,6 +3785,18 @@
           "long": "chart",
           "positional": false,
           "required": false
+        },
+        {
+          "global": false,
+          "help": "Proceed even when the upgrade would DELETE a stateful component the release is running, losing its data. Refused by default: a chart that renames a store (minio -> rustfs in 0.6.0) drops the old one, and every sandbox reads the bundle store at start, so the next turn fails rather than merely a rollback",
+          "id": "allow_stateful_removal",
+          "long": "allow-stateful-removal",
+          "positional": false,
+          "possible_values": [
+            "true",
+            "false"
+          ],
+          "required": false
         }
       ],
       "hidden": false,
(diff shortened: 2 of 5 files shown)


jw3329:
## Validated against the live cluster — and it was wrong

I said in the description that this hadn't run against a real cluster. It has now, and the first run found a **false positive** in the guard.

Against the live sre-bot release it reported **four** components as about to be deleted:

```
refusing to apply: this would DELETE 4 stateful component(s) ...
  sre-bot-clickhouse
  sre-bot-minio
  sre-bot-postgres
  sre-bot-valkey
```

Only `minio` is really lost. The release was installed with `nameOverride=sre-bot`, so its resources are `sre-bot-<component>` while the chart renders `sre-bot-curie-<component>`. The guard compared `metadata.name`, so every component looked absent from the render — and *any* value affecting resource naming would have done the same.

That is exactly the failure this PR's own text warned about: show an operator four false removals and they learn to pass `--allow-stateful-removal` by reflex, then lose minio for real.

## Fixed

Identity is now `app.kubernetes.io/component` — carried on both sides, perturbed by no naming value, read from `spec.selector.matchLabels` on the live objects and from the render. The error still reports the resource **name**, which is what an operator recognises in their cluster.

Re-run against the same release:

```
refusing to apply: this would DELETE 1 stateful component(s) ...
  sre-bot-minio
```

**4 → 1.** Falsified by switching back to name comparison, which fails four of six tests including a new one pinning the `nameOverride` case.

## Worth stating plainly

My unit tests passed the entire time. They compared hand-written name lists I invented for both sides, so they agreed with the bug. That's the fourth time this session a green test suite hid a defect that one real run exposed — same as `governs_key`, the plaintext credential, and the chart blindness.

PR #1337, opened 2026-08-05
Stop ClickHouse from starving the node with its own logs

### LINKED ISSUE #1335: ClickHouse's own system logging can consume the whole node and surface as an opaque runner-error
## What happened

On a single-node install (t4g.large, 2 vCPU), agent turns began failing in Slack
with:

```
The run failed (runner-error) after 3 attempt(s). Flagging for a human.
```

The worker's own logs give the proximate cause:

```
ClaimTimeoutError('claim curie-thread-2190be4f36-44dd25 not bound within 90.0s')
```

The node was CPU-starved: load average 6.57 on 2 cores, `/proc/pressure/cpu`
`some avg10=92.83`, while IO pressure was `0.02` and memory pressure `0.36`.
Every pod read `Running`, the node read `Ready`, and no alert fired — the shape
`sandbox/substrate.py` already documents in its ClaimTimeoutError comment.

## Root cause: a merge that can never succeed, retried forever

The CPU was entirely consumed by ClickHouse merging **its own system log
tables**. Storage breakdown at the time:

| database | size |
|---|---|
| `system` | **2.72 GiB** |
| `default` (the actual Langfuse data) | **31.56 KiB** |

A ~90,000:1 ratio of self-telemetry to payload, on an install with 12 traces and
127 observations.

| table | size | rows |
|---|---|---|
| `system.text_log` | 2.30 GiB | 63,295,036 |
| `system.trace_log` | 215 MiB | 14,212,454 |
| `system.metric_log` | 140 MiB | 676,824 |
| `system.asynchronous_metric_log` | 73 MiB | 120,786,562 |

The engine driving it, from `system.text_log` (17,465 Error rows/hour, all
identical):

```
Exception while executing background task {61ecb0a0-…::202607_30066_33598_279}:
Code: 241. DB::Exception: Memory limit (total) exceeded: would use 1.36 GiB
(attempt to allocate chunk of 4361168 bytes), maximum: 1.35 GiB
```

`61ecb0a0-…` is `system.metric_log`. The loop:

1. The chart caps ClickHouse at 1536Mi; its internal ceiling lands at 1.35 GiB.
2. `system.metric_log` accumulates enough parts that merging them needs more
   than that.
3. The merge fails. ClickHouse retries failed merges indefinitely.
4. Each attempt burns CPU and writes a **~4 KB stack trace** into
   `system.text_log`, at Debug level.
5. `text_log` grows, needs its own merges, which log too.
6. Because merges keep failing, parts **accumulate**, so the next merge is
   **larger**, so it fails harder. The merge range above spans ~3,500 parts.

It is a death spiral, not linear growth, and it is entirely self-sustaining —
independent of agent traffic. Daily average CPU on this box went 17% → 27% →
30% → 38% → 65% → 81% → 98% over eight days with no change in Slack volume.

Log level is a contributing factor. One hour of `system.text_log`:

| level | rows |
|---|---|
| Debug | 253,727 |
| Trace | 58,186 |
| Error | 17,465 |
| Information | **15** |
| Warning | **5** |

Top loggers are all merge machinery: `MergeTreeSequentialSource` (174,775),
`MergeTask::PrepareStage` (37,208), `MergeTreeBackgroundExecutor` (17,445).

## Why this is a chart defect, not an operator error

Four defaults combine into a trap, and no single one looks wrong alone:

- system log tables enabled (`text_log` is **off** by default in vanilla
  ClickHouse)
- logger level at Debug rather than `information`/`warning`
- no TTL on any system log table
- a 3 GiB PVC and a 1.35 GiB memory ceiling **too small to merge what that
  logging produces**

Any self-hosted install will reach this given enough uptime. It took eight days
here. The PVC was at 91% when found.

What makes it expensive to debug is the distance between cause and symptom: the
operator sees `runner-error` in Slack. Nothing connects that to ClickHouse's
logging config, and every dashboard reads healthy because CPU starvation sets no
node condition.

## Remediation applied (runtime only, no chart change)

```sql
TRUNCATE TABLE system.metric_log;              -- removes the impossible merge
TRUNCATE TABLE system.text_log;
TRUNCATE TABLE system.trace_log;
TRUNCATE TABLE system.asynchronous_metric_log;
TRUNCATE TABLE system.error_log;
ALTER TABLE system.<each> MODIFY TTL event_date + INTERVAL 1 DAY;
```

Result, within three minutes:

| | before | after |
|---|---|---|
| load average | 6.57 | 0.67 |
| CPU pressure `some avg10` | 92.89 | 5.94 |
| ClickHouse CPU | 999m (pinned at its 1000m limit) | 45m |
| running merges | 3–5 | 0 |
| `system` database | 2.72 GiB | 350 KiB |
| `text_log` rows | 63,295,036 | 1,928 |

Note the truncate of `text_log` alone would **not** have fixed it — that table
is the exhaust. The failing merge is on `metric_log`, and without truncating it
the loop refills `text_log` within days.

## Suggested fix in the chart

1. Set the ClickHouse logger level to `warning` (or `information`).
2. Disable `text_log`, `trace_log`, `metric_log` and `asynchronous_metric_log`,
   or ship them with a short TTL. On a trace store holding kilobytes, none of
   them earn their cost.
3. If they stay enabled, size the memory limit against what merging them
   requires, or lower ClickHouse's merge memory settings so a merge cannot
   exceed the ceiling and wedge.
4. Consider whether `runner-error` can carry a hint when the node is under
   sustained CPU pressure — `substrate.py` already knows this is the likeliest
   cause and says so in a comment no operator will read.

Environment: curie chart with `langfuse.deploy=true`, `clickhouse.deploy=true`,
ClickHouse 24.8.14.39, k3s v1.36.2, single node t4g.large (arm64), 3 GiB
local-path PVC.

jw3329:
Fixes #1335.

ClickHouse ships configured for a dedicated analytics box. This chart runs it as
an embedded store for kilobytes of Langfuse traces, with **no config overlay at
all** — so it keeps the image defaults: logger at `trace`, and 30 days of
`text_log` / `trace_log` / `metric_log` / `asynchronous_metric_log`.

Measured on a live single-node install after eight days:

| database | size |
|---|---|
| `system` | **2.72 GiB** — ClickHouse logging about itself |
| `default` | **31.56 KiB** — the actual Langfuse data |

A 90,000:1 ratio, in a 3Gi volume that was 91% full.

## It ends in a merge that cannot succeed, retried forever

Once `system.metric_log` needs more than the 1536Mi limit to merge, the merge
throws `MEMORY_LIMIT_EXCEEDED`. ClickHouse retries failed merges indefinitely.
Each attempt writes a **~4 KiB stack trace into `text_log`**, which then needs
merging itself — and because failures make parts accumulate, the next merge is
*larger* and fails *harder*.

One hour of `text_log`: **253,727 Debug and 58,186 Trace rows against 15
Information.** Every top logger was merge machinery (`MergeTreeSequentialSource`
174,775, `MergeTask::PrepareStage` 37,208).

A positive feedback loop with no relationship to traffic:

| Day | 1 | 2 | 3 | 4 | 5 | 6 | 7 | 8 |
|---|---|---|---|---|---|---|---|---|
| Avg CPU | 17% | 27% | 30% | 38% | 65% | 81% | 98% | 92% |

Slack volume never changed. Eventually sandbox pods could not win enough CPU to
bind their claim inside the 90s deadline, and every turn escalated as an opaque
`runner-error`.

**Nothing catches it.** CPU starvation sets no node condition — the node reads
`Ready`, pods read `Running`, probes stay green, disk looks fine. `substrate.py`
already predicts this exact shape in a comment, which is the only place it's
written down.

## The fix

A `config.d` overlay — the same mechanism the official image uses for its own
`docker_related_config.xml`, so no image change and no entrypoint override:

- logger → `warning`
- the four firehose tables **removed**
- `query_log` / `part_log` / `error_log` kept (low volume, genuinely useful) with
  a **3-day TTL**, so none can become the next `text_log`
- `clickhouse.systemLogs.enabled=true` brings the four back for anyone debugging
  ClickHouse itself — still TTL-bounded

## This is the second fix for this failure

The first capped ClickHouse's CPU limit from `2` to `1` — see the existing
`resources` comment in values.yaml, which describes this same incident. That
halved the blast radius and left the cause untouched: ClickHouse still wanted a
whole core, permanently, and on a 2-vCPU node half the machine is still fatal to
a 90s deadline. **The same escalation came back.** That comment now records why
the ceiling alone wasn't enough — the core was going to merging its own logs,
not to serving Langfuse.

## Gate

`charts/curie/ci/clickhouse-logging-assertions.sh`, wired into `helm-ci`. It
asserts the log level, the removals, a TTL on every surviving table, plus two
things that are silently fatal if dropped:

- **the `subPath` mount** — mounting the directory would hide the image's own
  `docker_related_config.xml`
- **the checksum annotation** — without it a values edit renders a new ConfigMap
  that the running server never reads, so the fix looks applied while the node
  keeps starving

It also parses the rendered XML in both modes, since a malformed `config.d` file
stops the server from starting and the render is the cheap place to catch it.

**Mutation-tested** — reverting the log level, dropping the `subPath`, removing
the checksum annotation, and un-TTLing `query_log` are each caught. The first
run of the checksum mutation surfaced a bug in the assertion itself (an emptied
`annotations:` parses as `None`, raising `TypeError` instead of failing
cleanly); fixed and re-verified.

## Verification

`helm lint` clean; full chart plus `values-dev`, `values-e2e-harness` and
`values-e2e-nogvisor` all render; `deploy=false` renders nothing;
`persistence.enabled=false` still gets its `data` emptyDir alongside the new
`config` volume.

Applied by hand to the affected install first (truncate + runtime TTLs), which
is what this change makes permanent:

| | before | after |
|---|---|---|
| load average | 6.57 | 0.29 |
| CPU pressure `some avg10` | 92.89 | 5.57 |
| ClickHouse CPU | 999m (pinned at its limit) | 45m |
| running merges | 3–5 | 0 |
| `system` database | 2.72 GiB | 350 KiB |
| `text_log` rows | 63,295,036 | ~2,000 |

Note the runtime version does **not** survive a restart: ClickHouse rebuilds
system log tables from config at startup and renames any whose structure
doesn't match, so a hand-applied TTL is reset to the config's 30 days. That's
precisely why this belongs in the chart.

**Unrelated:** `ci/render-assertions.sh` fails identically before and after this
change on macOS bash (`postgresPassword: unbound variable`), so it's untouched.

### DIFF
--- a/.github/workflows/helm-ci.yaml
+++ b/.github/workflows/helm-ci.yaml
@@ -98,6 +98,18 @@ jobs:
           set -euo pipefail
           bash charts/curie/ci/litellm-sidecar-assertions.sh
 
+      # Prove ClickHouse's own telemetry stays bounded: logger at `warning`, the
+      # four firehose system log tables removed, and whatever remains carrying a
+      # TTL. At image defaults ClickHouse logs at `trace` and keeps 30 days of
+      # self-telemetry -- measured at 2.72 GiB against 31 KiB of real Langfuse
+      # data, filling a 3Gi volume in eight days and ending in a merge that
+      # exceeded the memory limit and retried forever, starving the node until
+      # every agent turn escalated as an opaque `runner-error`.
+      - name: helm render assertions (clickhouse self-telemetry)
+        run: |
+          set -euo pipefail
+          bash charts/curie/ci/clickhouse-logging-assertions.sh
+
       # Prove the Langfuse web+worker containers carry a numeric runAsUser
       # alongside runAsNonRoot (issue #351) so the kubelet can verify non-root
       # and the pods actually start on an enforcing cluster.
--- a/charts/curie/ci/clickhouse-logging-assertions.sh
+++ b/charts/curie/ci/clickhouse-logging-assertions.sh
@@ -1,0 +1,160 @@
+#!/usr/bin/env bash
+#
+# Render-assertion test for the ClickHouse self-telemetry defaults. Proves:
+#
+#   1. DEFAULT render: a config.d overlay ConfigMap exists, sets the logger to
+#      `warning`, and REMOVES text_log/trace_log/metric_log/
+#      asynchronous_metric_log outright.
+#   2. DEFAULT render: the diagnostic tables that remain (query_log, part_log,
+#      error_log) each carry a TTL, so none of them can become the next
+#      text_log.
+#   3. The StatefulSet actually consumes it: a `config` volume, a subPath mount
+#      into config.d, and a checksum annotation so an edit rolls the pod.
+#   4. subPath specifically -- mounting the directory would hide the image's
+#      own docker_related_config.xml.
+#   5. OVERRIDE: `systemLogs.enabled=true` keeps the four tables but TTL-bounds
+#      them rather than leaving them unbounded, and `retentionDays` is honoured.
+#   6. The overlay is well-formed XML in both modes. A malformed config.d file
+#      makes ClickHouse refuse to start, and the render is the only place that
+#      is cheap to catch.
+#   7. `clickhouse.deploy=false` renders nothing, and `persistence.enabled=false`
+#      still gets its `data` emptyDir alongside the new `config` volume.
+#
+# Why this is worth a gate: stock ClickHouse config is sized for a dedicated
+# analytics box, and the chart runs it as an embedded store for kilobytes of
+# Langfuse traces. Left at image defaults it logs at `trace` and retains 30 days
+# of its own telemetry. Measured on a real install: 2.72 GiB of self-telemetry
+# against 31 KiB of Langfuse data, filling a 3Gi volume in eight days, ending in
+# a system.metric_log merge that exceeded the memory limit and was retried
+# forever -- each failure writing a ~4 KiB stack trace into text_log, which then
+# needed merging itself. Daily CPU went 17% -> 98% over eight days with no
+# change in traffic, and every agent turn escalated as an opaque `runner-error`
+# because the sandbox could not win enough CPU to bind its claim inside 90s.
+#
+# A regression here is silent for about a week and then takes the node out, so
+# it is asserted rather than trusted.
+#
+# Runnable locally (from anywhere) and from CI. Fails loudly.
+set -euo pipefail
+
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+CHART="$(cd "$SCRIPT_DIR/.." && pwd)"
+
+TMP="$(mktemp -d)"
+trap 'rm -rf "$TMP"' EXIT
+
+TPL=templates/clickhouse.yaml
+
+fail() { echo "FAIL: $*" >&2; exit 1; }
+
+DEFAULT="$TMP/default.yaml"
+VERBOSE="$TMP/verbose.yaml"
+NOPERSIST="$TMP/nopersist.yaml"
+OFF="$TMP/off.yaml"
+
+echo "=== Rendering ClickHouse (defaults) ==="
+helm template rel "$CHART" --show-only "$TPL" > "$DEFAULT"
+
+echo "=== Rendering ClickHouse (systemLogs.enabled=true, retentionDays=7) ==="
+helm template rel "$CHART" --show-only "$TPL" \
+  --set clickhouse.systemLogs.enabled=true \
+  --set clickhouse.systemLogs.retentionDays=7 > "$VERBOSE"
+
+echo "=== Rendering ClickHouse (persistence disabled) ==="
+helm template rel "$CHART" --show-only "$TPL" \
+  --set clickhouse.persistence.enabled=false > "$NOPERSIST"
+
+echo "=== Rendering ClickHouse (deploy=false) ==="
+helm template rel "$CHART" --show-only "$TPL" \
+  --set clickhouse.deploy=false > "$OFF" 2>/dev/null || true
+
+# ---------------------------------------------------------------- 1, 2, 4, 6
+python3 - "$DEFAULT" "$VERBOSE" <<'PY'
+import sys, yaml, xml.etree.ElementTree as ET
+
+FIREHOSE = ["text_log", "trace_log", "metric_log", "asynchronous_metric_log"]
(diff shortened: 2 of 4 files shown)


jw3329:
## Revised after booting it — the design changed twice

Pushed 96bf8b2. Rendering was not enough; running `clickhouse-server:24.8` against the rendered overlay found two things the render could not.

### 1. Removing `text_log` was wrong

The image ships `<text_log><level>trace</level></text_log>` — and that **level** is the firehose, not the table. `text_log` supports a level filter, so it can be quiet *and* present.

| | rows |
|---|---|
| unfiltered (current default) | ~312,000 / hour |
| filtered to `warning` (measured live) | **10 total**, all `Warning`, after 15 queries + a forced error |

Keeping it matters: a queryable `text_log` is exactly what diagnosed this incident, and unlike the log file it survives the pod. The three timer-driven samplers are still removed — they have no level filter, so on/off is the only lever.

### 2. Errors were reaching nobody

The image logs **only to files** under `/var/log/clickhouse-server/`. On Kubernetes that means `kubectl logs` is empty and the diagnostics die with the container. Verified: with the first version, a forced error appeared on neither stdout nor in any table. Added `<console>1</console>` — now 10 Error/Warning lines land on stdout where an operator will find them. Cheap at `warning`.

### 3. The boot caught a defect in my own change

`--` inside an XML comment is illegal, and this repo's prose dash style put **four** of them in the overlay. ElementTree parses it happily; the Poco SAX parser ClickHouse uses does not:

```
Failed to merge config with '...curie-logging.xml': SAXParseException: Invalid token
exit 232
```

A comment would have taken the store down on every install, and it passed an ElementTree-only check. The assertion now tests for it explicitly — the point of a render gate is to catch what the *server* rejects, not what Python tolerates.

Also added a TTL to `processors_profile_log`, which the live boot showed the image creates too. Missed because it is small today, which `text_log` also once was.

### Verification

Booted on `clickhouse-server:24.8` (arm64): boots clean, samplers absent, `text_log` present and filtered, TTLs applied, errors on stdout. Mutation-tested at seven mutations now — the comment dash, dropping `console`, and unfiltering `text_log` each caught, plus the original four. `helm lint` clean; dev and both e2e overlays render.

PR #1370, opened 2026-08-06
Draft ADR-0099: hooks are bundle-declared turns the system starts

TheConnMan:
Draft ADR defining agent hooks: bundle-declared background turns with a standing prompt, run through the kernel with no user present and no channel reply.

- Trigger vocabulary v1: cron (API sweeper-loop scheduler), bind (control-plane emitted), and the webhook ingress ADR-0079 already decided
- Standing-prompt authorization; approval gates still apply via the existing async card/resume machinery
- Silent turns with a durable run record (ran/skipped/failed); per-(agent, hook) CAS skip serialization; no backfill, no retries
- Per-hook systemPrompt, model, and sealed-env overrides; operator-authority keys stay fenced
- Tier-parity CLI test-fire verb
- No platform frequency caps while the bundle author's keys pay; named as the first requirement of any multi-tenant deployment

Sequences behind ADR-0079's unimplemented contract work. The memory ADR in #1276 names its bootstrap and compaction as the first two consumers of this machinery.

### DIFF
--- a/docs/adr/0099-hooks-are-bundle-declared-turns-the-system-starts.md
+++ b/docs/adr/0099-hooks-are-bundle-declared-turns-the-system-starts.md
@@ -1,0 +1,220 @@
+# 99. Hooks are bundle-declared turns the system starts
+
+Date: 2026-08-06
+
+Status: Draft
+
+## Context
+
+Every turn Curie runs today begins with a user message. `QueuedTurn`
+(`packages/aci-protocol/src/aci_protocol/turn.py`) carries an `author`, a
+`text`, and a `ReplyHandle` whose `placeholder` is required; the kernel
+(`apps/worker/src/curie_worker/kernel.py`) unconditionally edits that
+placeholder as it streams. There is no way for the system itself to start a
+turn: no schedule, no reaction to a platform event, no background work of any
+kind.
+
+Three prior decisions have been converging on this gap without closing it:
+
+- [ADR-0079](0079-inbound-triggers-as-a-new-event-kind.md) (Accepted,
+  unimplemented) decided the wire shape: a `source` field on the queued event
+  (`slack | webhook | cron`), an optional placeholder, and an API ingress at
+  `POST /hooks/{agent}/{hook}`. None of it is built; `QueuedTurn` still has no
+  `source` and `placeholder` is still required.
+- The bundle format already carries a `triggers` authoring extension
+  (`TriggerDeclaration` in `packages/plugin-format/src/plugin_format/models.py`,
+  types `cron` and `webhook`). It is validated at deploy time and consumed by
+  nothing; Epic #29 tracks the runtime that was never started.
+- ADR-0095 (Draft, in review) needs exactly this machinery twice: a memory
+  bootstrap turn when an agent is bound to a channel, and a nightly compaction
+  turn. It deliberately deferred the machinery to this ADR's question: what
+  runs a turn no user started?
+
+So the declaration exists, the wire contract exists on paper, and the first
+two consumers are already specified. What no ADR has decided is the semantics:
+what authorizes such a turn, what prompt, model, and credentials it runs with,
+what happens when a fire overlaps a still-running fire, where the output goes
+when nobody is watching, and how an author tests one before trusting it to a
+schedule.
+
+External systems that solved this converge on a small vocabulary (researched
+2026-08). Durable-execution frameworks (Temporal, Inngest, Restate, Hatchet)
+all offer per-entity serialization keyed by the caller with a named overlap
+policy, and their default for periodic maintenance is "skip if still running."
+Letta's sleep-time agents designate the background agent as the sole heavy
+writer to shared memory rather than locking, and recommend it run the strong
+model, not a cheap one, because consolidation quality compounds. Claude Code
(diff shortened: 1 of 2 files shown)


jw3329 (approved):

yichenzhang-curietech (commented):
Reviewed alongside #1371, since they land together and one of my findings only shows up when you read both.

The research behind this is unusually thorough and the Alternatives section is doing real work, especially the reasoning for skip-over-buffer and for no catch-up. Five things I would want settled before this is Accepted.

**1. The skill tier does not have an API, so it cannot have the scheduler or the run record.**

Consequences says "The scheduler is an API-side loop, so hooks fire wherever the API runs: skill, local, and cluster tiers all get them without substrate-specific plumbing." The skill tier has no API and no Postgres: `curie skill message` talks straight to a local runner container's ACI surface, bypassing the dispatcher and worker by design (cli/CLAUDE.md), and `curie skill` exposes only up/down/status/message/eval/approvals. So there is no loop to host the scheduler and no table to hold a run record.

ADR-0077 already set the precedent for exactly this shape by deciding there are no skill-tier durable approvals and reporting the capability unavailable by design. Suggest saying the same here explicitly: the skill tier gets manual test-fire only, with no schedule and no durable run record. That keeps the tier-parity rule honest rather than quietly unmet.

**2. Skip-on-overlap with no lease can starve a hook permanently.**

"A fire that finds the previous run still in flight records `skipped` and does not queue." If a run never settles because a worker died or a sandbox wedged, the CAS claim stays held and every subsequent fire skips forever. The ADR is right that a skip is a recorded outcome rather than a silent no-op, so it is visible in a query, but visible-and-never-firing-again is still a stall that no component will resolve on its own.

The worker already solved this class: a claim carries a timeout and unfinished work is reclaimed. Suggest the run claim carry a lease, that a claim past its lease be reclaimable, and that the reclaim itself be a recorded outcome so the stall is diagnosable rather than inferred from a gap.

**3. "The author pays" holds for tokens but not for the rest of the cost.**

The v1 argument for no frequency floor is that "the bundle author's schedule burns the bundle author's keys, so the author owns the trade." That is true of model spend and not of anything else. A one-minute cron spawns an operator-funded sandbox every minute, plus queue traffic and scheduler work on the operator's cluster. The asymmetry does not wait for multi-tenancy; it exists as soon as the bundle author and the cluster operator are different people, which is the ordinary case for a deployed agent.

I am not arguing for a platform-wide floor, which the ADR rejects for good reasons. Suggest either an operator-side floor (a values knob, off by default) or, at minimum, naming the operator-borne cost in Consequences so it is a known trade rather than an unnoticed one.

**4. Interaction with the kill switch and with budgets is unstated.**

Both exist today. If an agent is killed or over budget, does a scheduled fire still run? Fail-closed looks obviously correct, but an implementer should not have to derive it, and a hook that keeps firing past a kill switch would be a surprising way to discover the answer.

**5. The frozen-contract change should name its version consequence here.**

`ReplyHandle.placeholder` is `placeholder: str` with no default in `packages/aci-protocol`. Making it optional is a change to a closed schema in the frozen contract. Under the compatibility rule in ADR-0101 and ADR-0103 the version implication is decided by the shape of the change, and it is cheaper to state it in the ADR that causes it than to have the lane discover it at the gate. One sentence in Consequences would do it.

Nothing above changes the shape of the decision, which I think is right. 1 and 2 are the two I would not merge without.

TheConnMan:
Rebased onto main and pushed; the conflict was only the generated ADR index. All five land, taken as follows.

**1.** Correct, and the Consequences claim was wrong as written. The skill tier now gets manual test-fire only: no loop to host the scheduler, no table for a run record, and the schedule and record verbs report the capability unavailable by design, the ADR-0077 shape you point at.

**2.** Taken. The claim carries a lease, a claim held past it is reclaimable by the next fire, and `reclaimed` is its own recorded outcome so the stall is diagnosable rather than inferred from a gap. Same shape as `reclaim_min_idle_ms` and `XAUTOCLAIM` in `stream_consumer.py`.

**3.** Taken as the Consequences sentence rather than a floor. The operator carries the scheduler loop, the queue traffic, and a fresh sandbox per fire, and that asymmetry exists as soon as the author and the operator are different people. The paragraph now also says what is rejected is a platform wide floor, not an operator setting one for their own cluster.

**4.** Fail closed, now stated: a fire for a killed or over budget agent records `blocked` and never starts a turn. That drove the outcome vocabulary too. `skipped`, `blocked`, `reclaimed`, and `failed` stay distinct members, because the question a run record has to answer is why a hook has not run since Tuesday, and that is unreadable if every non-run collapses into one value.

**5.** Substance taken, citation not. ADR-0101 amends ADR-0074, which governs the agent-facing CLI result schemas under `schemas.curietech.ai/cli/`, and ADR-0103's gate is `cli/tests/schema_inventory.rs`. `packages/aci-protocol` is a separate regime with its own frozen-interface rule: bump `PROTOCOL_VERSION` (0.2.0 today), regenerate the three language artifacts, gated by `test_schema_compat.py` and `test_wire_lock.py`. Consequences now names that. Worth noting the optional placeholder is ADR-0079's decision and already Accepted; this ADR only sequences behind it.

PR #1371, opened 2026-08-06
ADR 0100: agents search their own surface

TheConnMan:
Draft ADR defining a narrow experiment: an agent may search only its installed surface and current channel.\n\nThe platform must enforce that boundary. The MCP spike must prove useful retrieval, deny reads outside the bound surface, preserve provenance, retain no copied corpus, and enforce bounds below the model.\n\nMemory and lifecycle compaction remain separate decisions.\n\nCloses #1371

### DIFF
--- a/docs/adr/0100-agents-search-their-own-surface-through-the-channel-port.md
+++ b/docs/adr/0100-agents-search-their-own-surface-through-the-channel-port.md
@@ -1,0 +1,191 @@
+# 100. Agents search their own surface through the channel port
+
+Date: 2026-08-06
+
+Status: Draft
+
+## Context
+
+Claude Code is effective in a repository because the filesystem is a surface
+it can read and search: `Read` and `Grep` are cheap, bounded, and always
+available. A Curie agent's working surface is the channel it is bound to, and
+today it cannot read that surface at all. The channel port
+([ADR-0020](0020-message-port-rendering-free-channel-interface.md)) is
+write-only: the `SlackSink` protocol
+(`apps/worker/src/curie_worker/slack_sink.py`) exposes post, update, and
+status verbs and nothing else. There is no verb to fetch a message by
+permalink, read a window of history, or search the channel. An agent asked
+"what did we decide about this in March" has only its own memory document and
+whatever the current thread contains.
+
+This capability is independent of memory. ADR-0095 (Draft, in review) names a
+bounded escape valve into raw history as one consumer, but the primary
+consumer is the live turn: agentic search over the bound surface as part of
+the normal skill, the way Claude Code greps a repo mid-task.
+
+The permission argument is what makes a bounded version safe: anyone who can
+invoke the agent in a channel can already read that channel, so the agent
+reading the same history back adds no access. Crossing the channel boundary
+is what would add access, so the channel boundary is the security line, and
+it must be the platform's line, because (researched 2026-08) Slack does not
+draw it: Slack's search APIs are workspace-wide by default,
+`search:read.public` explicitly does not require channel membership, and the
+Real-time Search API's `context_channel_id` is a scoping hint, not a
+boundary. No surveyed system (Slack's own APIs, MCP Slack servers, enterprise
+search vendors) implements a true single-channel search primitive; the only
+hard fence found in the ecosystem is a config-level channel allowlist with no
+search verb at all.
+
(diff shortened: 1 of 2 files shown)


jw3329 (approved):

yichenzhang-curietech (commented):
Reviewed alongside #1370. My first two points come from reading them together, and I do not think either is visible from this ADR alone.

The zero-copy stance is well argued, and I want to call out the honesty of "federated paging has worse recall than an index, and the vendor that ships both says so. Curie accepts that." That is the right way to record a trade-off.

**1. The invoker-visibility argument does not cover hook-started turns, and this ADR extends the verbs to them.**

The safety case here rests on one sentence: "anyone who can invoke the agent in a channel can already read that channel, so the agent reading the same history back adds no access." That argument is only available when there is an invoker.

ADR-0099 (#1370) introduces turns that no user started, and this ADR says "Background turns get the same tools under the same flag." For those turns there is no invoker whose visibility bounds anything, so the justification is simply not applicable, and no replacement is offered. The bundle author's standing prompt may well be adequate authorization, but that is a different argument from the one the ADR makes, and it should be made rather than inherited.

**2. The injection mitigation assumes somebody reads the output. A hook turn has no reader.**

The 2024 Slack AI exfiltration is named as the threat model, and the v1 mitigation is provenance framing, with a classifier screen explicitly deferred as "future hardening". For a user-started turn that is a defensible v1: the turn produces a reply a human reads, so a steered turn has a witness and a wrong answer is noticed.

ADR-0099 states that "a hook turn has no reply handle" and that nothing appears in the channel unless the turn's own tools post. So the combination this ADR authorizes is an unattended turn that reads channel history, may act through tools, and produces no reply anybody sees. That is the exfiltration shape with the witness removed, and it is reachable in v1 by enabling one bundle flag.

Two ways out, either fine by me: hold the read verbs off for hook-started turns in v1, or move the classifier screen from future hardening into v1 for that path specifically. What I do not think works is shipping both defaults together while the stated mitigation quietly assumes a reader.

**3. The load-bearing distribution assumption needs a stated failure behaviour, not just a stated consequence.**

Recording "each agent is its own workspace-internal Slack app" as an architectural commitment is exactly right, and so is naming what happens to the cost model without it. What is missing is what the adapter does when the assumption is false. Under the constrained limits the ADR cites, a paginated search would not fail; it would return whatever it managed to page before the budget ran out, and a partial result presented as a search result is worse than a refusal, because the agent cannot tell the difference.

Suggest the adapter detect the constrained case and refuse the verb, so a broken deployment reports a missing capability instead of quietly answering from a fraction of the history.

**4. Small: is a refused `resolve_reference` distinguishable from a miss?**

The verb "refuses references outside the bound scope". If that refusal is distinguishable from "no such message", it confirms the existence of out-of-scope messages to anyone who can get the agent to dereference a guessed permalink. Probably minor given who can invoke an agent, but it costs one sentence now and is awkward to change later.

Points 1 and 2 are the ones I would want resolved before Accepted. 3 and 4 I would happily see as follow-ups.

TheConnMan:
Rebased onto main and pushed. All four land; 1 and 2 changed the document, 3 and 4 are in as written.

**1.** Correct, and the argument is replaced rather than inherited now. A hook turn's authorization is the bundle author's declared standing prompt, deployed and reviewed as part of the bundle, running against a channel an operator deliberately bound the agent to. The invoker-visibility argument is explicitly scoped to user-started turns, and Consequences says a path reaching these verbs with no invoker carries the standing-prompt argument instead.

**2.** Taken, but resolved differently from either of your two exits. Reading the bound surface is most of what background work is for, and the same author writes the hook, its standing prompt, and this flag, so v1 does not fence one against the other. What the ADR stops doing is claiming the mitigation covers the case. It now records that a hook turn has no reader, that an unattended turn reading planted history and acting through tools is the 2024 shape with the witness removed, that it is reachable by enabling one flag, and that this is a known accepted exposure owned by the author who enabled both surfaces. It also names that case as the concrete trigger for promoting the classifier screen out of future hardening. If that is not enough for Accepted, the classifier is the thing worth arguing about rather than the flag.

**3.** Taken. The adapter detects the constrained case and refuses the verb through the existing `ChannelCapability` path, so a broken deployment reports a missing capability instead of quietly answering from a fraction of the history.

**4.** Taken, one line in Bounds: an out-of-scope reference and one that does not exist return the same result.

Issue #1278, opened 2026-08-04
Attest the evaluated bundle, resolved policy, and deployed sandbox as one promotion record


TheConnMan:
## Problem

Curie versions immutable bundles and evaluates them before promotion, but it does not yet expose one verifiable record proving that the exact evaluated bundle reached a runner under the expected policy.

#62 covered image provenance and signing. This is the separate application level chain from source commit to bundle to evaluation to deployment.

## Scope

Create a canonical promotion record with:

1. Source commit and repository identity.
2. Bundle SHA256.
3. Eval suite identity and result reference.
4. Resolved policy digest, covering egress, model, tool, approval, trigger, and secret name requirements without any secret values.
5. Runner image digest and harness package identity.
6. A signature or attestation that the runner can verify before executing the fetched bundle.

The dev deployment produces the record. Production promotion references the same immutable record and does not rebuild it.

## Constraints

1. Do not change the frozen ACI or plugin format.
2. Reuse the image provenance work from #62 rather than creating a second image signing mechanism.
3. Keep the public shape diagnostic first. #1040 should expose the record under JSON without secret values.
4. Fail closed when a fetched bundle or resolved policy does not match the promotion record.

## Acceptance criteria

1. A dev deployment emits a record linking one Git commit, bundle digest, runner image digest, effective policy digest, and passing eval result.
2. Promoting that version to production reuses that record and bundle digest without rebuilding either.
3. Altering the stored bundle after evaluation causes sandbox startup to refuse execution with a recoverable diagnostic.
4. curie info JSON identifies the deployed source commit, bundle digest, policy digest, runner identity, and evaluation reference without exposing credentials.
5. A test proves that a policy change after evaluation cannot be reported as the evaluated deployment.

## References

Forge is the external reference for signed build artifacts and policy aware runtime packaging: https://github.com/initializ/forge

Related: #62, #1040, #844, #1095, #1226

Issue #1226, opened 2026-08-03
Name a single org-level security posture that narrower scopes can only tighten


TheConnMan:
Related to #515 (admin-force config tier) but a different axis: #515 is about which knobs a tenant cannot loosen, this is about giving the whole approval and screening surface a single name an operator picks once.

## The problem

Curie has the machinery. ADR-0010 defines approval gates, ADR-0050 says a declared approval policy is armed exactly or not at all, ADR-0056 makes gate grantability an operator opt-in, ADR-0035 bounds a one-shot post-approval allowance, ADR-0077 keeps skill-tier durable approvals unavailable. Each is right on its own.

What is missing is the sentence an evaluator needs. Someone asking "how do I lock this down before I point it at anything real" currently gets routed through five ADRs and has to assemble the posture themselves. That is a real adoption cost and it lands on exactly the security-conscious buyer the self-host story is aimed at.

## The shape worth stealing

YC's QM (MIT) ships three org-level postures a narrower scope can only tighten. From its README:

> - **Strict** - every harness tool call pauses for human approval, except the two no-effect turn enders.
> - **Auto** (default) - a classifier screens provenance-labelled external data and tool results before they reach the model; a deployment can point that at its own screening proxy.
> - **Dangerous** - no content screening, no pauses between tool calls.

and underneath all three:

> The predeclared command policy - approval rules and hard denials for things like recursive deletes or destructive SQL - applies in every posture, Dangerous included.

Two properties are what make it work and both are borrowable independent of the specific three names. First, the floor is unconditional: there is a policy no posture can turn off, so "Dangerous" is bounded rather than open. Second, the lattice only goes one way: a scope can tighten, never loosen, which is the same monotonicity #515 wants for config.

## Scope

- Name the postures and define each as a specific set of the defaults Curie already has, not as new mechanism.
- Define the unconditional floor: the denials that hold in every posture. Curie has candidates already in the policy gate.
- Enforce tighten-only composition from org down to agent down to scope.
- One documented line per posture in `SECURITY.md` and the README, since the point is that an evaluator can read it in ten seconds.

Mostly a naming and defaults exercise over shipped machinery. If it turns out to need new enforcement, that is the signal it wants an ADR rather than an issue.

## Acceptance

An operator sets one value at the org level, and the resulting approval and screening behavior at every tier is derivable from that value plus the documented floor, with no per-gate configuration required to reach a defensible default.

TheConnMan:
## Make posture a calculated result

[...]

Issue #1040, opened 2026-07-28
curie info: one command that prints the resolved bundle plus discovery diagnostics


TheConnMan:
## Why

An agent driving Curie has no single command that answers "what did the harness actually resolve from this bundle, and what did it look for and fail to find." The pieces exist scattered across `curie schema`, the `skill up` env summary, and per-verb `--json`, so an agent has to stitch them together. Worse, it cannot distinguish "not declared" from "declared but did not register", which is exactly the failure that makes a bundle look fine and behave wrong.

Vercel's eve ships `eve info --json` for this, and its published agent instructions name it as the command an agent must run before reporting success. Our `llms.txt` already declares that this CLI's primary user is a coding agent driven by a developer, so the gap is on our critical path, not a nice to have.

## What

Add `curie info` printing the resolved bundle:

- discovered skills (name, path)
- declared MCP servers and whether they load
- declared secrets and boot env, satisfied vs missing (ADR-0049)
- declared approval gates (ADR-0050)
- eval cases discovered (count, path)
- channel and comms connectedness
- artifact paths (bundle root, evals, plugin manifest)
- resolved model and the credential source by name

Plus a `diagnostics` block: for every candidate the harness looked at and did not register, what it looked for, where it looked, and why it did not count. That block is the whole point. A plain inventory is already obtainable; the rejection reasons are not.

## Constraints

- `--json` shaped per ADR-0074 (versioned result schemas) and the ADR-0021 agent-facing CLI contract (exit 0/1/2/3, `{error, fix}`).
- Answered at every tier per ADR-0041 (`skill`, `local`, `cluster`). Where a field is meaningless at a tier, say so explicitly rather than omitting it.
- Never print a secret value. Names only, consistent with the existing secrets surface.
- Register in `cli/command-manifest.json` and the schema index so the existing drift gates cover it.

## Done when

- `curie info --json` exits 0 against `examples/weather` and its payload lists the weather skill, the declared MCP server, and the eval case count.
- Deleting `examples/weather/evals/cases.json` makes the diagnostics block name the missing file by path, and the command still exits 0. An incomplete bundle is a diagnosis, not a crash.
- Renaming `skills/weather/SKILL.md` to a non-conforming name makes `info` report that directory as looked at and rejected, with the reason. It must not simply vanish from the inventory.
- The verb is answered at all three tiers per ADR-0041, including any explicit not-available-at-this-tier answers.
- Schema committed; field-parity, emit-parity, and docs-lint gates pass.

[...]

Issue #515, opened 2026-07-16
Admin-force config tier: operator guardrails a tenant cannot loosen


TheConnMan:
From the Grok Build review. **Worth-it-later**.

Grok's config merge has an intentional inversion: "requirements" layers sit **above** the user's own config, so an org-pinned `requirements.toml` or macOS MDM policy WINS over what the user wrote (`crates/codegen/xai-grok-config/src/loader.rs:288-299` `reapply_requirements`, test at :490). Permission rules separately merge by severity, not order: `deny > ask > allow`, so a global `deny` beats a project `allow`. There is also a root-owned `/etc/grok/requirements.toml` with `disable_bypass_permissions_mode=true` that a user cannot edit away.

Curie is dev-first and self-host, but the multi-tenancy epic (#158) needs exactly this primitive: an operator-pinned layer that a tenant cannot loosen (egress allowlist, allowed models, telemetry endpoint). This is the proven shape for "the operator sets guardrails the agent or user cannot widen", and it maps onto our fail-closed egress and authorizer posture.

- [ ] Admin-force config tier that wins over user/project config
- [ ] Severity-based permission merge (`deny > ask > allow`) rather than precedence-based

Relates to #158.

Source: review of https://github.com/xai-org/grok-build

TheConnMan:
## Explainable monotonic policy resolution

Add an operator readable effective policy result alongside the enforced configuration layer.

For every declared requirement, deployment planning should report the final decision and the owner of the controlling constraint. Composition must be monotonic from organization to agent to execution scope. A narrower scope may deny or require approval, but cannot widen an upstream denial.

Model the decision at capability level where possible. Do not make broad command text matching the primary policy mechanism. The durable result should distinguish allowed, approval required, denied, and unavailable.

#1226 can name the resulting operator posture. #1040 can expose it as JSON. #1278 can bind its digest to the evaluated deployment.

Forge is a useful external reference for layered policy attribution: https://github.com/initializ/forge

PR #1423, opened 2026-08-10
Draft ADR-0102: a named security posture is computed, not configured

yichenzhang-curietech:
**This PR is a Draft ADR and implements nothing.** ADR-0085 is explicit: a Draft may merge for discussion but cannot authorize implementation, and implementation starts only after an Accepted status published with explicit maintainer approval. I stopped at the decision rather than writing the resolver.

Flagging that plainly because #1226 reads like a naming exercise. Under option B (a resolver, not a `SECURITY.md` paragraph) it closes real alternatives, is cross-cutting, and the same decision now spans four issues (#1226, #515, #1040, #1278) -- past the "promote on the third" bar in AGENTS.md. The issue's own test agrees: *"If it turns out to need new enforcement, that is the signal it wants an ADR rather than an issue."*

## The decision

**A posture is a NAME for a computed result, not a value an operator sets.**

A configured posture is a second policy engine: it has to be reconciled against the settings it claims to summarize, and reconciliation is where drift lives. A computed name cannot disagree with what is enforced, because it is derived from it.

That threads the two constraints that otherwise conflict:

> "a compact presentation of the effective monotonic policy, not a second policy engine [...] auditable rather than a documentation label"

> "this should not depend on adding multi-tenancy [...] the posture names could still be implemented without the org components"

Those pull apart **only if a resolver needs an org tier**. It does not -- deployment, agent and bundle are a lattice already, and the org tier joins as an outer layer later without changing shape.

## Three facts, re-measured today

| | |
|---|---|
| **No org tier** | `models.py` declares no `Org`/`Tenant`, no `org_id`; `org_name` is a `Settings` string. #158 open. |
| **No screening** | No screening, classifier, or provenance-labelling mechanism anywhere in the tree. |
| **No resolver** | Defaults scattered across the API authorizer, a chart value, the bundle manifest, and two `agents` columns. |

## The floor: five denials no posture can turn off

Each verified in place, not recalled: self-approval refused server-side, egress default-deny with ADR-0067 keeping NetworkPolicy additive, ADR-0050's exactly-or-not-at-all arming, ADR-0056's default-false grantability, ADR-0077's unavailable skill-tier approvals.

The floor is what makes the most permissive posture **bounded rather than open**.

## Two deliberate refusals

**Screening is not an axis.** The borrowed model defines its middle rung by it. Naming a rung Curie cannot enforce is exactly the documentation label this decision exists to avoid. When screening ships it becomes a fourth axis under the same rule.

**The third posture is Permissive, not Dangerous.** The floor makes the stronger word untrue, and a name that overstates how far a posture goes is worse than a duller accurate one.

## Two things recorded rather than resolved

`grantableViaPolicy` (ADR-0056) is a **loosening** switch declared in the *bundle*. ADR-0056 is coherent on its premise -- the operator authors the manifest -- which **inverts** under a tenant-facing lattice, where the bundle author is the tenant. This ADR changes nothing about 0056; it writes that premise down as load-bearing so the first tenant-facing change has to face it.

And the resolver **reads only**. No new denial, not on any request path. That boundary is what keeps this reversible; enforcement built on it later is a separate decision.

## What happens next

On acceptance I implement it under #1226: the resolver over the three layers, per-axis source attribution, the posture names, and one documented line each in `SECURITY.md` and the README. If any of the three postures or the floor's contents want changing, that is cheapest here, in review, before code exists.

`curie dev docs-lint` green; `docs/adr/README.md` regenerated by the script rather than by hand.

Refs #1226

### DIFF
--- a/docs/adr/0104-a-named-security-posture-is-computed-not-configured.md
+++ b/docs/adr/0104-a-named-security-posture-is-computed-not-configured.md
@@ -1,0 +1,143 @@
+# 104. A named security posture is computed from the layers that exist, never configured
+
+Date: 2026-08-10
+
+Status: Draft
+
+Implements [#1226](https://github.com/curie-eng/curie/issues/1226).
+
+Deliberately does NOT depend on [#158](https://github.com/curie-eng/curie/issues/158)
+(multi-tenancy) or [#515](https://github.com/curie-eng/curie/issues/515)
+(monotonic policy resolution). The Decision says why.
+
+## Context
+
+Curie has the machinery: ADR-0010 (gates), ADR-0050 (armed exactly or not at
+all), ADR-0056 (grantability is an operator opt-in), ADR-0035 (bounded one-shot
+allowance), ADR-0067 (NetworkPolicy is additive), ADR-0077 (no skill-tier durable
+approvals). Each is right on its own.
+
+What is missing is the sentence an evaluator needs. "How locked down is this
+before I point it at anything real" is currently assembled by hand from six ADRs
+and four code locations, which is a real adoption cost on exactly the
+security-conscious self-host buyer.
+
+Three facts about the tree, measured rather than assumed, bound what can be
+decided today:
+
+- **No organization tier.** `models.py` declares no `Org`/`Tenant` and no
+  `org_id`; `org_name` is a `Settings` string the console renders. The layers
+  that exist are **deployment** (chart values, process env), **agent** (a row),
+  and **bundle** (the manifest).
+- **No content screening.** No screening, classifier, or provenance-labelling
+  mechanism exists anywhere in the tree.
+- **No policy resolver.** Defaults are scattered by construction: the
+  self-approval block is in the API authorizer, default-deny egress is a chart
+  value, gates are in the bundle manifest, model and thinking are agent columns.
+
+## Decision
+
+**A posture is a NAME for a computed result, not a value an operator sets.** A
+resolver reads the deployment, agent and bundle layers, reports the effective
+answer per axis together with the layer that imposed it, and the posture name
+labels the shape of that answer.
+
+### 1. Computed, never configured
+
+There is no `posture:` setting. An operator changes a posture by changing a
+constraint, so the name can never disagree with what is enforced.
+
+This is the load-bearing choice. A configured posture is a second policy engine:
+it must be reconciled against the settings it claims to summarize, and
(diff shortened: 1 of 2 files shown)


TheConnMan (commented):
@yichenzhang-curietech overall I think this is OK. It's a dense ADR and I think can be condensed.

My understanding is overall security posture needs to be computed through some command so the user can understand it from across all the various locations it lives. Is that a fair summariziation?

yichenzhang-curietech:
Condensed: 202 lines to 143, which puts it alongside 0101 (144) and 0093 (146). Cut the prose, kept the rejected alternatives since AGENTS.md requires an ADR to record what it decided against.

Also renumbered **0102 to 0104** and rebased: 0102 and 0103 both landed while this was open, and the new 0102 is "Accepted alongside implementation with explicit approval". Only `docs/adr/README.md` conflicted, and it is generated, so `check-docs.sh` rebuilt it rather than my hand-merging.

## Your summary, and the half it leaves out

> security posture needs to be computed through some command so the user can understand it from across all the various locations it lives

Fair, and it is the larger half. Two refinements:

**"Computed" is doing sharper work than "aggregated".** There is deliberately no `posture:` setting anywhere. An operator changes a posture only by changing a constraint, which is what makes the name unable to disagree with what is enforced. Aggregating four locations into one view would still leave a name that could be wrong; deriving it means that state is not expressible. That is the alternative the ADR closes.

**The floor is a promise, not a view.** Five denials hold in every posture -- self-approval refused, egress default-deny, ADR-0050's exactly-or-not-at-all arming, grantability off unless a manifest opts in, no skill-tier durable approvals. That half is not "understand what you have"; it is "here is what you cannot turn off, whatever you set". It is also what makes the most permissive posture bounded, and why the third name is Permissive rather than Dangerous.

So: **a command that computes it, plus a floor that no posture can lower.** The command is the part you named; the floor is the part that makes the answer worth reading.

## One thing worth settling before acceptance

Screening is deliberately **not** an axis, because Curie has none -- the external model this borrows from defines its middle rung by screening, and naming a rung we cannot enforce is exactly the documentation label the ADR exists to avoid. Worth your eye because ADR-0100's channel-read verbs would give untrusted content a path into context, which is when its absence starts to matter.

Happy to trim further if it is still dense.

jw3329 (approved):

TheConnMan (approved):

PR #1467, opened 2026-08-12
Harden object store chart assertions

### LINKED ISSUE #1451: worker-object-store-assertions.sh reports health on four configurations that break an install
`charts/curie/ci/worker-object-store-assertions.sh` closes the exact gap it was written for, and the
mutations it advertises do go red (reverting `worker.yaml` names all four vars; a differing endpoint
is caught; an equal-but-`localhost` endpoint is caught). Four things it treats as healthy would still
break an install.

All of the below were executed at `4b3346d9` against the real render, then reverted.

**1. The key list is closed, so a fifth variable added to `api.yaml` alone is invisible.**
```bash
# worker-object-store-assertions.sh:31
KEYS=(S3_ENDPOINT_URL S3_ACCESS_KEY S3_SECRET_KEY BUNDLE_BUCKET)
```
Adding `S3_REGION` to `api.yaml` only leaves **all 17** chart assertion scripts at exit 0. The script
never enumerates the api Deployment's actual env; it checks membership and equality of the four names
it was handed. This is the same divergence shape the script exists to catch, one variable over.

There is already a fifth setting in this family: `s3_region` defaults to `"us-east-1"` independently
in `apps/worker/src/curie_worker/config.py:414` and `apps/api/src/curie_api/config.py:60`. Neither
Deployment sets it, so today they agree only because two separate code defaults happen to match.

**2. A `secretKeyRef` naming a Secret or key that does not exist passes green.**
Pointing both templates at `name: does-not-exist, key: nopeNotAKey` gives exit 0, and the script
prints the bogus reference as evidence of health:
```
  S3_SECRET_KEY      = <secretKeyRef does-not-exist/nopeNotAKey>
```
`:95-99` only checks that *a* `secretKeyRef` is present, and the equality loop compares api against
worker, so a reference identically wrong on both sides is indistinguishable from a correct one. The
pods would sit in `CreateContainerConfigError` and never start. The sibling
`dispatcher-api-wiring-assertions.sh:137-142` already validates the ref's name and key.
(The narrower one-sided case, renaming the key in `secrets.yaml` while leaving consumers alone, IS
caught incidentally by `render-assertions.sh`.)

**3. It omits the sandbox bundle-fetch init container**, which reads the same object store under
different names (`S3_ENDPOINT`, not `S3_ENDPOINT_URL`, plus `AWS_DEFAULT_REGION` and
`AWS_CONFIG_FILE`, at `agent-sandbox.yaml:285-300`). Changing its endpoint or bucket leaves the script
green while every sandbox fails at bundle fetch. On a Kubernetes install this is the container that
actually fetches the bundle on the turn path.

**4. It inspects `containers[0]`, not the named application container** (`:76`). Safe today, since
both Deployments have exactly one container, but a sidecar rendered first would be inspected instead
and the assertion would report "missing" or go quietly wrong.

**5. Minor: present-but-empty values pass the assertion and fail only by accident.** With
`agentSandbox.runner.bundleFetch.bucket: ""` the failure list stays empty, the script prints
`api and worker agree on all four object-store variables:`, and then crashes with `KeyError:
'valueFrom'` in the reporting loop (`:109-111`). Right exit code, actively misleading output.

## The generalization that would have caught more

The script encodes the instance rather than the class. The class-level form, every
`curie_worker.config` field whose default contains `localhost`, cross-referenced against the rendered
worker Deployment, is roughly 15 lines and immediately flags a second live gap in the same commit:
```
VALKEY_HOST      in worker: True     DATABASE_URL   in worker: True
S3_ENDPOINT_URL  in worker: True     LANGFUSE_HOST  in worker: False   (in api: True)
```

## Acceptance criteria

- The script derives the key set from the api Deployment's rendered env rather than a literal list,
  so a new object-store variable on one side alone is a failure. Adding `S3_REGION` to `api.yaml`
  only makes it red.
- Every `secretKeyRef` it reports is resolved against the rendered chart Secret; the
  `does-not-exist/nopeNotAKey` mutation makes it red.
- The sandbox bundle-fetch init container's endpoint and bucket are asserted consistent with the api
  and worker values, allowing for the different variable names.
- Containers are selected by name, not index.
- Each of the above is proven by a stated mutation that makes the script red.

TheConnMan:
Closes #1451

Summary

* Derive object store keys from the rendered API environment.
* Validate rendered Secret names and keys.
* Check named API, worker, and sandbox bundle fetch containers.

Done check

[x] Code Review approved.
[x] Scope Review approved.
[x] Security Review approved.
[x] Required mutation matrix passed and restored every template.
[x] Helm lint and all chart assertion scripts passed.

### DIFF
--- a/charts/curie/ci/worker-object-store-assertions.sh
+++ b/charts/curie/ci/worker-object-store-assertions.sh
@@ -16,13 +16,17 @@
 #
 # So this asserts AGREEMENT, not presence. Presence alone would still pass the
 # day someone points the worker at a different-but-populated endpoint.
+#
+# The key set is DERIVED from the API's rendered env rather than listed here, so
+# a fifth object-store variable added to api.yaml alone is a failure instead of
+# being invisible. Every secretKeyRef is resolved against the rendered Secret,
+# because a reference identically wrong on both sides passes an equality check
+# and leaves both pods in CreateContainerConfigError.
 set -euo pipefail
 
 CHART=${CHART:-charts/curie}
-DEFAULT_RENDERED=$(mktemp)
-EXTERNAL_RENDERED=$(mktemp)
 EXTERNAL_VALUES=$(mktemp)
-trap 'rm -f "$DEFAULT_RENDERED" "$EXTERNAL_RENDERED" "$EXTERNAL_VALUES"' EXIT
+trap 'rm -f "$EXTERNAL_VALUES"' EXIT
 
 cat > "$EXTERNAL_VALUES" <<'EOF'
 rustfs:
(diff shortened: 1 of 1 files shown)


TheConnMan:
Heads up from the post-merge review of the 2026-08-12 window: this PR no longer merges.

#1468 merged at 18:43Z on 2026-08-12, 67 minutes after this PR was opened, and rewrote
`charts/curie/ci/worker-object-store-assertions.sh` including the `env_of` function this PR reworks.

```
$ git merge-base origin/pr-1467 066fe1b8
ef3e4dcd8ef258dacad4f8c50e16e21518ff0c00
$ git merge --no-commit --no-ff origin/pr-1467
CONFLICT (content): Merge conflict in charts/curie/ci/worker-object-store-assertions.sh
Automatic merge failed
```

When rebasing, the resolution should keep this PR's label-based `env_of`, not `main`'s
`name.endswith("-worker")` version. Detail and the mutation evidence are in a comment on #1451.

Two things from the merged version to fold in while rebasing, since neither exists on this branch's
base: the merged script deleted the `localhost` / `127.0.0.1` endpoint guard, which is worth
restoring, and it indexes `api[k]` without an api-side presence check, so a missing api variable
raises a bare `KeyError` rather than a named message.

PR #1541, opened 2026-08-14
Forward-merge main into next (ADR-0096 acceptance)

TheConnMan:
Brings main (v0.6.x line) forward into next after #1513 merged, so next carries the accepted ADR-0096 and the v0.6.x fixes that landed since the last sync.

One conflict, in .github/workflows/helm-ci.yaml: both lines added the worker/api object-store render assertion; next already had it plus the eval-wiring assertion, so the merge keeps next's side and drops main's duplicate block.

### DIFF
--- a/.claude/skills/implement/SKILL.md
+++ b/.claude/skills/implement/SKILL.md
@@ -64,7 +64,11 @@ Classify from the work, not from a requested process size.
 | Build | Multiple streams, architecture, security, deployment, or unclear interactions | Written plan, separate test and implementation contexts, reviews, and real surface verification when applicable |
 
 Promote a direct or quick change when its real scope exceeds the selected path. Do
-not add a planning artifact for a trivial mechanical change.
+not add a planning artifact for a trivial mechanical change. If a quick change
+turns up an assumption that qualifies for the de risk gate below, promote it to
+build before its first edit. If that assumption only appears after an edit, stop
+and either escalate or restart as a build path change; do not carry those edits
+forward.
 
 ## Build rules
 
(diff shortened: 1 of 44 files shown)


jw3329:
Heads-up on something that landed after this PR was opened, and that this PR does not yet contain. **This PR's merge result is currently clean** — I checked `refs/pull/1541/merge` and it carries exactly one `0021`. Nothing here is blocked. The issue is the *next* sync.

`0021_console_sessions.py` merged to `main` at `a52f1341` on 2026-08-14T22:11Z — about five hours after this PR was opened, so it is outside this merge. `next` already carries `0021_agent_channels.py`. Both declare the same identifiers:

```python
# main:0021_console_sessions.py     # next:0021_agent_channels.py
revision      = "0021"              revision      = "0021"
down_revision = "0020"              down_revision = "0020"
```

Git will not flag this — they are different filenames, so a merge adds both without conflict. Alembic will, at load time, and `next` then has two revisions claiming `0021` with `0022_approvals_reply_kind` pointing `down_revision = "0021"` ambiguously.

It should fail loudly rather than ship: `apps/api/tests/test_migrations.py` runs a real `alembic upgrade head` through its `migrated` fixture, so CI goes red at the merge that brings both together. Not a silent landmine — just an hour lost by whoever hits it without context.

**Renumbering choice is not obvious, which is why I am raising it rather than sending a patch:**

- Renumbering `0021_console_sessions` (say to `0025`) touches the **stable v0.6.x line**, where it may already be applied as `0021` in a deployed database. Any environment that ran it would need its `alembic_version` reconciled.
- Renumbering the `next` chain means `0021_agent_channels` → `0022` → `0023` → `0024` all shift, and `next` is unreleased, so no deployed database has them yet — cheaper, but four files and their `down_revision` links.

The second looks cheaper from here, but I do not know whether `0021_console_sessions` has already been applied anywhere on the v0.6.x line, and that is the fact that decides it.

Context for why I was looking: we run `v0.7.0-rc.4` on a cluster-tier install and were evaluating whether to track `next` directly. Separately — per-commit images are published for `main` only (`release.yaml`), so `next` commits have no images in GHCR; that is presumably intentional given "keeping release publication on main only", but it does mean `next` cannot be run as-is without building locally.

TheConnMan:
I don't want to do this right now. We will do this later. I don't want to do it specifically for the purpose of an ADR acceptance. When we release the next deploy or when we're prepping for the next release of Next, we will do this.

Issue #1535, opened 2026-08-14
Two releases on one cluster: sandbox controller version skew and shared image tag collisions


TheConnMan:
Two installs of different chart versions on one cluster surfaced two shared resource problems.

## 1. Sandbox controller version skew

The agent-sandbox controller is cluster scoped and shared, but there is no version skew contract between it and per release chart templates. An rc.3 controller lacked the NetworkPolicy RBAC that rc.4 templates assume, so rc.4 claims silently never bound (no error surfaced; sandboxes just never appeared). Fixed by hand applying a namespace Role and RoleBinding granting the controller ServiceAccount networkpolicies CRUD:

```yaml
kind: Role
rules:
  - apiGroups: [networking.k8s.io]
    resources: [networkpolicies]
    verbs: [get, list, watch, create, update, patch, delete]
# bound to ServiceAccount agent-sandbox-controller in agent-sandbox-system
```

The real fix is chart owned RBAC that tracks template requirements per release, or a stated controller compatibility floor.

## 2. Shared image tag collisions

The documented image path imports fixed tags (`curie-*:local`, `curie-runner:latest`) into the shared cluster runtime, so a second install silently overwrites the first install's images. This broke a parallel install with an ACI 0.2.9 vs 0.3.0 protocol skew until the images were rebuilt and the tags restored. Chart and docs should default to per release tags.

PriorityClass conflicts between releases are already documented with a values workaround; these two are the remaining halves.

## Environment

Single node k3s, rc.3 and rc.4 era installs side by side. Observed 2026-08-14.

TheConnMan:
## Cycle 10 retest (2026-09-09, source-built origin/main `0d3f6d4a`)

QA10-02, DUP of QA9-01 / QA8-02 / QA7-02. Isolated release `curie-qa909`, `agentSandbox.controller.deploy=false`, unique PriorityClass names, no shared-controller deploy.

`curie cluster up --dry-run --json` still planned:

```
kubectl label namespace agent-sandbox-system \
  curietech.ai/created-by=curie-qa909 curietech.ai/created-in=curie-qa909 --overwrite
```

The live install did not execute that command. After install, `agent-sandbox-system` still had only `kubernetes.io/metadata.name=agent-sandbox-system`. Unique QA PriorityClasses were created and later removed. Soak and default PriorityClass UIDs were unchanged.

So the dry-run plan is still a lie relative to the live path, which is the honesty face this issue already tracks. No new live mutation of the shared controller namespace.

[...]

PR #1545, opened 2026-08-14
Draft ADR-0129: cluster singleton ownership and version skew

TheConnMan:
## Summary

Adds Draft ADR-0129 on the `next` release train. It defines one owner for cluster-scoped controller resources, consumer-owned namespaced grants, a fail-closed compatibility contract, release-specific imported image tags, and guarded owner lifecycle operations.

The draft was rechecked against the current `next` chart and the review findings:

- Consumer RBAC follows each namespace even when that release does not own the controller.
- Helm cannot value-gate `crds/`; supported consumer installs therefore use `--skip-crds`, enforced by the `curie` path and pinned in CI.
- Contract epochs and compatible revision ranges are verified against registered consumers; breaking transitions drain and migrate the whole set rather than mixing epochs.
- External controllers must publish a verifiable contract and pass live controller/RBAC/CRD probes; acknowledgement alone is insufficient.
- Owner uninstall refuses while consumers remain and requires explicit handoff or disaster-recovery force.
- The current consumer render still produces one `SandboxTemplate` and zero controller NetworkPolicy roles.

## Verification

- `bash scripts/check-docs.sh` passed at `e67b7179`.
- Static consumer render: `SandboxTemplate=1`, controller read `ClusterRole=0`, controller namespaced `Role=0`.
- Runtime tiers are not applicable because no chart template or runtime behavior changes.

Ref #1535

### DIFF
--- a/docs/adr/0105-one-release-owns-a-clusters-shared-singletons.md
+++ b/docs/adr/0105-one-release-owns-a-clusters-shared-singletons.md
@@ -1,0 +1,252 @@
+# 105. One release owns a cluster's shared singletons; every other release declares what it needs from them
+
+Date: 2026-08-14
+
+Status: Draft
+
+Implements [#1535](https://github.com/curie-eng/curie/issues/1535).
+
+Builds on [ADR-0023](0023-controller-networkpolicy-rbac-cluster-read-namespace-mutate.md)
+(the controller's NetworkPolicy RBAC split), [ADR-0059](0059-sandbox-is-a-bounded-resource-envelope.md)
+decision 5 (the two PriorityClasses), and [ADR-0067](0067-controller-networkpolicymanagement-unmanaged-for-rail-1.md)
+(`networkPolicyManagement: Unmanaged`). Each of those decided what a single
+install renders. None of them decided who owns the result when a second install
+lands on the same cluster.
+
+## Context
+
+`charts/curie` is a single umbrella chart, and the mental model behind it is one
+release per cluster. That model is already false in practice: issue #1535
+reports two installs of different chart versions side by side on one k3s node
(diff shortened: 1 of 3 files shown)


jw3329:
Heads-up on the ADR number: **0105 is already taken on `main` itself**, so this PR conflicts on its own base today, not just after a train merge.

- This PR adds `docs/adr/0105-one-release-owns-a-clusters-shared-singletons.md`.
- `origin/main` has `docs/adr/0105-external-native-session-checkpoints.md`, added 2026-08-12 in `14089ee2` — two days before this PR was opened, so the number was never free on this base.

Numbers in flight across both trains: `main` is allocated through 0123, `next` through 0124, 0125 is claimed by open #1934, and I've suggested 0126 for #1847 and 0127 for #1846. So **0128** is clear here.

Worth noting the cluster-singleton decision itself may need a re-read against what landed since — ADR-0105 (external native session checkpoints) and the v0.7.0 records now occupy some of the same ground.

TheConnMan:
Rebased onto `next`, renumbered to ADR-0129, regenerated the index, and refreshed the PR metadata.

The draft was reread against the current chart rather than only renumbered. It now handles Helm's ungateable `crds/` path through explicit consumer `--skip-crds`, defines first-owner bootstrap and verified external-manager paths, checks compatibility ranges against registered consumers, separates ordinary owner-first upgrades from drained epoch transitions, and refuses owner uninstall while consumers remain. The consumer render was rerun and still shows one `SandboxTemplate` with zero controller NetworkPolicy roles.

`bash scripts/check-docs.sh` passes at `e67b7179`. The cross-train numbering process gap is tracked in #1957.

Issue #1525, opened 2026-08-14
Let one agent bind multiple channels and respond in kind


TheConnMan:
Today an agent's channel binding is singular (ADR-0089): one agent, one (kind, address) pair. Running the same brain on Slack and email means deploying two agents sharing a bundle, with separate identity, memory, and approval routing.

Allow one agent to subscribe to multiple channel bindings. Default reply behavior: respond in kind — the reply goes back out on the channel the turn arrived on (the (kind, address) pair on the queued turn already carries everything needed to route it). Cross-channel behavior beyond that default is out of scope here.

Notes:
- The agent_channels table already models bindings as rows separate from the agent, so this is largely an API-contract and identity/memory/approval-routing decision, not a schema rewrite. The (kind, address) uniqueness constraint stays: one agent per channel, but many channels per agent.
- Supersedes part of ADR-0089's singular-channel decision, so it needs a superseding ADR before implementation.

Follow-up to ADR-0096 phase 2 (#1459).

Issue #1459, opened 2026-08-12
Promote the channel port: neutral agent binding and a post-first sink


TheConnMan:
The follow-up ADR-0096 decision 5 names first. Replace the Slack-typed binding surface (agents.slack_channel, its validators, the unique constraint from migration 0017) with a channel-neutral binding, and land a channel-neutral egress contract in which edit-in-place is a negotiated capability rather than an assumption, so a second channel adapter can deliver without emulating Slack semantics.

A draft ADR proposing the shape (agent_channels table keyed on channel kind and address, post-first sink, verified author carried structurally) is in review; this lands behind ADR-0096.

Done when an agent binds a non-Slack channel kind through the API without schema changes, and the existing Slack path passes unchanged.

TheConnMan:
Shipped on `next` in #1514 plus the #1542 follow-ups, milestoned v0.7.0. Closing by hand: a closing keyword in a PR targeting `next` never fires, because GitHub only honours it on a merge into the default branch (`main`). This will ship with the v0.7.0 cut.

Issue #1060, opened 2026-07-29
No supported way for an agent to read its own Langfuse traces


jw3329:
## Ask

An agent should be able to answer questions about its own behaviour in the channel it lives in:

- "why did you say that an hour ago?"
- "which tool call failed on that turn?"
- "what is this bot costing us per question?"

Curie already has the data. Langfuse is the observability and eval backbone (ADR-0004), every turn is traced, and traces carry the prompt, each tool call and result, timings, and token cost. What is missing is a supported way for the agent to *read* it.

## Current state

There is no first-party Langfuse connector, and no documented pattern for one. The two published community MCP servers are not viable:

- `langfuse-mcp` — 1 star, single maintainer, stdio only (no HTTP transport)
- `mcp-langfuse` — repository URL is still the unedited `github.com/yourusername/...` template

Neither is a reasonable dependency for a connector that can read **every prompt and tool result the agent has ever produced**.

## Why this belongs in the platform

The obvious workaround is for each bundle to ship its own connector. I started doing exactly that and stopped, because it is the wrong shape:

- Every bot on Curie would rebuild the same thing
- Each copy independently gets the credential handling, read-only scoping, and truncation right or wrong
- The runner's read-only rootfs forces a service rather than a stdio subprocess, so it is real infrastructure, not a config line
- Trace access is a genuine security surface — it should be scoped once, centrally

Langfuse is Curie's, the traces are Curie's, and the credentials are already in the chart (`langfuse.init.projectPublicKey` / `projectSecretKey`). A bundle author should not be assembling this.

## Sketch

A first-party read-only connector, opt-in per agent, exposing something like `list_traces`, `get_trace`, `list_sessions`, `get_daily_metrics`. Points at the in-chart Langfuse by default, uses the bootstrap keys already present, and truncates output so one `get_trace` cannot blow the model's context.

Worth deciding deliberately whether an agent may read traces from threads other than the current one — in a single team channel that is fine, across channels with different audiences it is not.

## Context

Found building a real bot on Curie and using only the public surface. Related friction from the same session in #1058.

Issue #521, opened 2026-07-16
ADR number collisions on main (0029, 0038) + CI gate


TheConnMan:
Found while adding ADR-0040. `docs/adr/` on `main` has genuine ADR **number collisions**:

- `0029-conversation-history-port-and-first-loader.md` and a second `0029-*`
- `0038-generated-interface-catalog-and-doc-lint-gate.md` and `0038-observability-cli-helper-for-the-agent-dev-loop.md`

Two branches each adding "the next ADR number" merge **clean in git** but collide in the tree, because git sees two unrelated new files, not a conflict. This is the same failure shape as the concurrent alembic-head fork: only a live check catches it, never the merge.

ADRs are the system of record per CLAUDE.md ("`docs/adr/` is the system of record for architecture decisions", immutable once Accepted, superseded rather than edited). A duplicate number breaks the supersession chain and makes "ADR-0038" ambiguous in every future citation, including the ones already written into other ADRs and issues.

- [ ] CI gate: fail if any ADR number appears twice in `docs/adr/`
- [ ] Renumber the existing 0029 and 0038 duplicates and fix inbound citations

Note: renumbering a merged ADR conflicts with immutability, so the fix needs a call on whether to renumber or to record the collision explicitly. Flagging rather than assuming.

PR #1774, opened 2026-08-20
Draft ADR-0115: agents call each other with no third party in the path

alexrao2000:
## What this is

A **Draft ADR** proposing how one Curie agent calls another. Draft only: per
[ADR-0085](https://github.com/curie-eng/curie/blob/main/docs/adr/0085-acceptance-not-implementation-authorizes-an-adr.md)
it authorizes no implementation until a maintainer accepts it. No code changes --
the ADR plus the regenerated `docs/adr/README.md` index.

## Why now

Pillar 4 in `docs/pillars.md` already says agents communicate "with people and
with other agents", and nothing implements the second half.

**To be unambiguous, since turn-minting machinery landed recently and this could
be misread as a refinement: agents cannot communicate today, at all.** A sandbox
is handed exactly one API credential, `CURIE_STATE_TOKEN` (scope `state.app`),
and the API accepts a scoped token on the state router and nowhere else -- "a
scoped token is rejected everywhere else" (`apps/api/CLAUDE.md`). The platform
key is never placed in the sandbox. An agent therefore holds no credential that
any turn-minting route will accept. There is no partial path and nothing to widen
by configuration.

What recent work changed is the *material* a solution can be built from, not the
reachability. Two ingress paths now mint turns, and both are platform-key gated
(`POST /channels/token` sits under `require_api_key`; a hook signature derives
from `settings.api_key`). Handing a sandbox either credential would not be a
shortcut, because each surface also has properties that make it the wrong one to
open up:

- The hook ingress (#269, ADR-0079) pins `author` to `hook:{hook}` precisely so
  an upstream cannot supply an identity, routes the reply to the target's channel
  so a caller never sees the answer, and shares one conversation per hook so
  concurrent calls serialize. A caller, a private answer, and concurrency are the
  three things a call needs; it refuses all three deliberately.
- The channel ingress (#1459, ADR-0096) scopes a `chn` token to a binding row, so
  it answers "which binding may this enqueue for" and never "who is asking".
- Slack is closed on purpose: `is_user_message` drops any event with a `bot_id`,
  and it has to, or the dispatcher's own placeholder would re-trigger it.

Every one of those is deliberate, which is why this needs a decision rather than
a patch. Two consumers are already waiting: ADR-0044 (Accepted) committed to a
"callable step" as its phase-1 build and assumes an API call that runs a turn and
returns a result, which does not exist; and discussion #1372 (LangChain) wants
the same primitive.

## What you have to do today, and why removing it is the point

There is a workaround, and naming it is the clearest statement of what this ADR
is worth. To make two Curie agents collaborate today you must stand up something
outside Curie: a process holding the **platform key** -- the most privileged
credential in the system -- that mints turns for each agent through
`/channels/turns` or `/hooks/{agent}/{hook}`, then recovers each answer by
watching the agents' channel bindings, because neither route returns a result.

That relay is the third party. It costs you the platform key in a new place, an
extra service to run and secure, attribution that stops at "the platform"
(`author=hook:{hook}`), no loop or depth bound, no approval integration, and
answers scraped out of human channels. Every one of those is a direct consequence
of using surfaces built for external upstreams to carry an internal call.

So the notable claim is not "agents can now talk" in the abstract. It is that
they talk with no third party in the path, no additional distribution of the
platform key, a caller and an accountable principal carried on the turn, and the
approval and loop controls the relay cannot provide.

## The decision, in one line

**One agent may ask another agent to do something, over a surface the platform
owns end to end, with no third party in the path.** No interop protocol to adopt,
no external broker, no orchestration framework, no bundle-shipped MCP server, no
relay holding the platform key, and no round trip through a human channel. It
reuses machinery the platform already has -- the `curie:runs` stream, the resume
path, the ADR-0033 scoped token, the runner's existing MCP mounting -- so a
second agent becomes reachable with no new infrastructure and no new dependency
to trust.

The ADR treats that as a constraint as well as a property: if this decision ever
requires a third party, it is the wrong decision and should be revisited rather
than extended.

Everything else in the ADR is how that is made safe, under one governing rule:
**a call must never widen what the target may do, and must never lose who
asked.** Concretely that means the target keeps its own credentials, model,
egress, and approval policy; the turn carries both the immediate caller and the
accountable human principal; a call can never satisfy an approval gate; and the
surface is default closed with bounded depth and refused cycles.

The three things a naive version would quietly do -- launder authority, lose
accountability, and loop -- are each addressed explicitly, with #410
("sandboxed agent can self-resolve its own approval") as the precedent for why
authority must not be assertable from inside a sandbox.

## Sequencing

- **Sequences behind ADR-0112 / #1672.** Delegation adds a fifth `QueuedTurn`
  producer, and 0112 states every producer added before its rename widens the
  blast radius. 0112 also already settled the enum question here: it rejects
  `human` precisely because "ADR-0079's own agent-to-agent case ... [produces]
  turns caused by an agent speaking as a participant".
- **Does *not* sequence behind ADR-0099** (hooks as bundle-declared turns). A
  hook fire is a job with no caller and no audience; a call has both. The ADR
  argues against merging them.
- **Needs discussion #1049** for the identity half. The ADR specifies attribution
  at the payload and run-record level, which holds today, and does not claim the
  IdP-issued agent identity #1049 asks for.

## Referenced and prerequisite issues

**Prerequisites / open questions**
- Discussion [#1049](https://github.com/curie-eng/curie/discussions/1049) -- agent identity, invocation authorization, caller/owner attribution. The authority prerequisite; its requirement 3 (accountability) and requirement 1 (invocation authorization) are what the Decision implements, and its on-behalf-of non-goal stays a non-goal here.
- [#1672](https://github.com/curie-eng/curie/pull/1672) -- ADR-0112, turn source naming. Must land first.
- Epic [#158](https://github.com/curie-eng/curie/issues/158) -- multi-tenancy. Cross-tenant calls are out of scope until a tenant object exists; #158 must revisit cost attribution.
- [#82](https://github.com/curie-eng/curie/issues/82) -- per-tenant identity before any multi-tenant deployment.

**Foundations this builds on (all closed)**
- [#269](https://github.com/curie-eng/curie/issues/269) -- the HMAC hook ingress, `POST /hooks/{agent}/{hook}`.
- [#1459](https://github.com/curie-eng/curie/issues/1459) -- channel port promotion; the `kind`/`address` routing pair and the sink selection the return path extends.
- [#410](https://github.com/curie-eng/curie/issues/410) -- sandboxed agent could self-resolve its own approval via a forwarded platform key. The reason ADR-0033's scoped token exists, and the precedent for "authority is never asserted from inside a sandbox".
- [#244](https://github.com/curie-eng/curie/issues/244) -- durable Approval record and the suspend/resume lifecycle the caller's resume reuses.
- [#411](https://github.com/curie-eng/curie/issues/411) / [#419](https://github.com/curie-eng/curie/issues/419) -- the approval seam and its resume-enqueue reconciler.
- [#248](https://github.com/curie-eng/curie/issues/248) / [#249](https://github.com/curie-eng/curie/issues/249) (epic [#23](https://github.com/curie-eng/curie/issues/23)) -- the durable state store and the auto-mounted MCP server pattern (ADR-0073) this surface copies.
- [#19](https://github.com/curie-eng/curie/issues/19) -- per-thread reply routing; why `ReplyHandle` carries a per-turn endpoint at all.
- [#1060](https://github.com/curie-eng/curie/issues/1060) -- an agent reading its own runs (ADR-0107), the sibling agent-facing scoped surface.

**Related / where the work would land**
- Epic [#29](https://github.com/curie-eng/curie/issues/29) -- triggers beyond chat. The hook half is done; this is the caller half.
- Epic [#30](https://github.com/curie-eng/curie/issues/30) -- bundle format and authoring extensions, where the `delegatesTo` declaration would land.
- Discussion [#1372](https://github.com/curie-eng/curie/discussions/1372) -- LangChain support; second consumer of the same primitive.
- [#1525](https://github.com/curie-eng/curie/issues/1525) -- one agent binding multiple channels, adjacent to the return-path `kind`.
- Discussion [#1176](https://github.com/curie-eng/curie/discussions/1176) -- the pillar taxonomy whose pillar 4 names this gap.
- [#521](https://github.com/curie-eng/curie/issues/521) -- ADR number collisions. See the numbering note below.

## Numbering note

Numbered **0115**. 0112 (#1672), 0113, and 0114 are claimed; 0114 is currently
claimed by *two* unmerged branches, which is the #521 collision class the
`check-docs.sh` gate exists to catch. Happy to renumber if a maintainer is
sequencing these.

## Verification

`bash scripts/check-docs.sh` passes: ADR numbers unique, no catalog drift, every
citation resolves. The regenerated `docs/adr/README.md` is committed rather than
hand-edited, per `docs/adr/AGENTS.md` step 6.

ADR-0112 is cited by number rather than relative link because its file is not on
`main` yet and a link would fail the citation lint; worth converting to a link
once #1672 merges.

## Review asks

1. **Is a new first-party surface the right call**, versus widening one of the
   two existing platform-key-gated ingresses? The ADR argues those surfaces
   refuse a caller, a private answer, and concurrency on purpose, so widening
   them would leave a different route wearing their name.
2. **Two attribution fields** (immediate caller + accountable principal) rather
   than one -- is that the right split given #1049 is still open?
3. **Suspend/resume rather than a blocking call.** Rejected blocking because a
   target can suspend on an approval gate for a human-shaped interval.
4. **Default closed, bundle declares / operator arms** -- consistent with
   ADR-0056 and ADR-0098, but it is a new authoring surface in epic #30.
5. The kernel steering rule needs a second predicate (a delegated turn is a
   `message`, so `is_job` is False). That is single-owner kernel territory and
   flagged in Consequences rather than decided here.

### DIFF
--- a/docs/adr/0115-agents-call-each-other-with-no-third-party.md
+++ b/docs/adr/0115-agents-call-each-other-with-no-third-party.md
@@ -1,0 +1,413 @@
+# 115. Agents call each other directly, with no third party in the path
+
+Date: 2026-08-20
+
+Status: Draft
+
+## Context
+
+Curie has no way for one agent to ask another agent to do something. Pillar 4
+in [`docs/pillars.md`](../pillars.md) already says agents communicate "with
+people and with other agents", and nothing implements the second half.
+
+This is not a missing feature so much as a missing decision. Turn-minting
+machinery landed recently, which makes it tempting to describe the gap as almost
+closed. It is not: **an agent cannot reach any of it.** The gap is total, and the
+recent work changes only what a solution can be built out of.
+
+### What exists today, and why none of it is reachable
+
+**The blocking fact first.** A sandbox is handed exactly one API credential:
+`CURIE_STATE_TOKEN`, a `state.app`-scoped ADR-0033 token minted per claim
+(`apps/worker/src/curie_worker/binding.py`). The API accepts a scoped token on
+the state router and nowhere else -- "a scoped token is rejected everywhere else"
+(`apps/api/CLAUDE.md`) -- and the platform key is deliberately never placed in
+the sandbox. So an agent holds no credential that any turn-minting route will
+accept. There is no partial path, no lossy path, and nothing to widen by a
+configuration change. Everything below is about which surface a *new* credential
+should authorize, not about a surface an agent can use today.
+
+Since [#269](https://github.com/curie-eng/curie/issues/269) closed,
+[ADR-0079](0079-inbound-triggers-as-a-new-event-kind.md) is implemented. A turn
+can start from something other than a person: `QueuedTurn.source` distinguishes
+a person's message from a job, `ReplyHandle.placeholder` is nullable, and
+`POST /hooks/{agent_id}/{hook}` (`apps/api/src/curie_api/routers/hooks.py`)
+turns a signed external event into a turn on the same stream a mention feeds.
+[#1459](https://github.com/curie-eng/curie/issues/1459) then landed the channel
+port from [ADR-0096](0096-port-adapters-are-deployed-services.md), so
+`POST /channels/turns` mints a turn for a binding named in the body, holding a
+`chn` token scoped to one binding row rather than the platform key.
+
+So two ingress paths mint turns today, and both are gated on the platform key:
+`POST /channels/token` mints a `chn` token under `require_api_key`, and a hook
+signature derives from `settings.api_key`. Handing either credential to a sandbox
+is not a shortcut to this feature, because each surface also has properties that
+make it the wrong thing to open up -- properties that are deliberate, and load
+bearing where they sit:
+
+- **The hook ingress cannot carry a caller, by construction.** `_mint_turn` sets
+  `author=f"hook:{hook}"`, and the code says why: an upstream-supplied identity
+  there "would let a hook impersonate one to anything downstream that reads the
+  field". The reply route comes wholly from the target's binding row, so the
+  answer goes to the target's channel and a caller never sees it. And
+  `_conversation_id` is per hook, so two calls would share one thread and the
+  second would defer behind the first. A caller, a private answer, and
+  concurrency are the three things a call needs, and this surface refuses all
+  three on purpose.
+- **The channel ingress is for adapters, not agents.** A `chn` token is scoped
+  to a binding row so an ingress adapter can enqueue for that binding. It says
+  nothing about who asked, and the same reply-routing property applies: the
+  platform mints the route from the binding, never from the request.
+- **Slack is closed on purpose.** `is_user_message` in
+  `apps/dispatcher/src/curie_dispatcher/handlers.py` drops any event carrying a
+  `bot_id`. An agent posting into a channel cannot wake another agent. Without
+  that filter the dispatcher's own placeholder would re-trigger it, so the filter
+  is load-bearing and not an oversight to remove.
+
+Two other things exist and are worth naming because they are often mistaken for
+this feature. Subagents inside a single turn already work: the bundle manifest
+carries an `agents` field (`packages/plugin-format/src/plugin_format/models.py`)
+that the harness picks up. That is one agent identity, one pod, one credential
+set, one approval policy -- decomposition, not two agents talking. And the
+`curie-state` MCP server ([ADR-0073](0073-agentos-state-mcp-server-and-state-boot-env.md))
+is namespaced per agent with a token bound to that agent, so two agents cannot
+even use it as a shared blackboard.
+
+### Two decisions already made that this one has to fit
+
+ADR-0112 renames the person-message value of `source` from `slack` to `message`.
+It is a Draft and is not on `main` yet, so it is cited here by number rather than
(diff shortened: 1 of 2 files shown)


docs/adr/0115-agents-call-each-other-with-no-third-party.md
```diff
@@ -0,0 +1,413 @@
+# 115. Agents call each other directly, with no third party in the path
```
TheConnMan, line 1: @alexrao2000 do you have a demo spike I can review of this?

Today an agent could tag another agent in Slack to kick off additional work, right? Are you saying that would not work today? It should work in the same way that a user calling the bot would.

Or are you describing having agents call other agents under the hood, not through Slack?

I don't fully understand this because I think it would already work. Help me understand what I'm missing.

alexrao2000: Yeah I'm describing them calling other agents through our platform rather than slack. It's not that it wouldn't work just a different path. There are both pros and cons I can see to this method but it's not as though it blocks the tagging method from working.

PR #1778, opened 2026-08-20
Draft ADR-0116: session identity over the ACI so a sandbox can be pre-bound

jw3329:
> **Renumbered 0114 -> 0116.** `spike/tier-diff` dated its ADR a day earlier and also
> claimed 0114; 0112 and 0115 are claimed by other open branches and 0113 is merged,
> so this one moved to the next unused number.

## Demo

ADR-0116 demo

Recorded on a dedicated cluster with a controlled neighbour: busy-loop pods
requesting `200m` each, matching what ClickHouse requests, against the runner and
bundle containers' `50m`. That 4:1 request ratio is the amplifier the chart names.

|                      | quiet node | under contention |
| -------------------- | ---------- | ---------------- |
| today (cold create)  | 4.72s      | **never ready**  |
| pre-bound            | **0.17s**  | 7.79s            |

Under contention today's path crossed `claimTimeoutSeconds` at **91.02s** and was
still not ready when the harness stopped at 110s. In production the worker gives
up at 90s and three of those escalate as an opaque `runner-error`, so this is the
recorded incident reproduced on demand rather than waited for. The pre-bound arm
took 7.79s against the same load: it degrades by a large factor and still leaves
the ceiling unreachable, which is the whole claim.

Two full runs agreed (4.66 / timeout / 0.14 / 7.86 and 4.72 / timeout / 0.17 / 7.79).
The harness is in [`prototypes/adr-0116-residency/`](prototypes/adr-0116-residency/)
with a README recording what it does **not** establish.

The cold baseline is cluster-shaped, and this cluster is generous to it: fake
model, no observability stack resident, a 6,961-byte bundle, gVisor off, so 4-5s
is its floor against the 17.39s measured on a real-model install with Langfuse and
ClickHouse. The column that carries the argument is the second one, where the
shape of the failure does not depend on the baseline.

## What this is

A **Draft** ADR plus its companion design doc, for the bonus architecture
assignment. Docs only, no code. Per
as amended by
**a merged Draft authorizes no implementation.** This branch carries the decision
and the evidence behind it for review, nothing more.

No issue exists yet; one should be filed against the ADR if it moves toward
Accepted.

## The fix was tested, not just proposed

The sub-second bind is measured, not assumed. On the same release, against an
isolated `SandboxTemplate` and `SandboxWarmPool` created alongside the shipped
ones so no live claim was disturbed, with the bundle ref and
`CURIE_PLUGIN_DIR` baked in **per pool** rather than injected per claim:

- Two warm pods pre-fetched, pre-extracted, and pre-booted a real bundle to
  `readyReplicas: 2` in **~28s**. This is a real-model install, so the chart's
  "only a fake-model pool boots cleanly" caveat turns out to follow from a pool
  being *generic*, not from it being warm.
- A claim carrying **no env** bound one of those pods and reached a ready runner
  in **0.19s**, against the 17.39s cold baseline. That is **91x**, and it is the
  sub-second bind the whole decision rests on.
- The bound pod was usable rather than an empty shell: `/bundles/current` held
  the bundle's `AGENTS.md`, `evals`, and `skills`. The pool refilled itself.
- An otherwise identical claim carrying **one** env entry did not bind a pool
  pod. It created its own sandbox and attempted a new pod, which is the
  cold-create path. **Its wall clock is not established** -- that attempt failed
  on the namespace quota rather than completing, so only the behavioural
  difference is demonstrated here.

What remains untested is the ACI change itself. How a bind is authenticated once
the runner token no longer arrives as pod env is still an open question, and it
is now the only thing between the measured 0.19s and a usable one.

### That failure surfaced a third density cap, tighter than the two named below

```
exceeded quota: curie-sandbox-quota, requested: limits.cpu=1,
used: limits.cpu=8, limited: limits.cpu=8
```

ADR-0059 decision 4's namespace quota counts `limits.cpu` and is set to 8, while
each sandbox declares `limits.cpu: 1`. The namespace therefore admits **exactly
eight concurrent sandboxes**, while the same quota allows `pods: 50` and its
`requests.cpu` sat at 400m of an allowed 4000m. A ceiling chosen generously per
sandbox becomes a cluster-wide concurrency cap once a quota counts it, so
decision 6 has to move the limit and not only the request.

## The problem, measured

A turn costs almost nothing to serve. Against a real model, a turn consumed
**2.3% of one core** for a short reply and **8.8%** for a 354-token reply. An
idle runner consumed **0.043%**. Curie is an I/O broker.

Two things dominate its resource cost instead, and neither is the turn.

**A new conversation pays a cold sandbox create on its critical path.** Measured
claim-to-ready on a real `curie-0.7.0` release, on an idle 12-core node, with the
runner image already resident and a 6,961-byte bundle: **17.39 seconds**, against
a hard `claimTimeoutSeconds: 90`.

| Phase | Measured | Share |
| --- | --- | --- |
| claim admitted, pod created | 0.75s | 4% |
| pod scheduling, container starts | 1.93s | 11% |
| `bundle-fetch` (`aws-cli s3 cp`) | 4.11s | 24% |
| `bundle-extract` | 0.42s | 2% |
| container transition | 1.25s | 7% |
| runner boot to `/healthz` | 8.93s | 51% |

The Kubernetes machinery is 4% of it. Kubernetes events independently place
`bundle-fetch` at **~5 wall seconds to download 6,961 bytes** -- that is process
startup, not transfer, paid on every pod generation for an immutable
content-addressed artifact, and paid again on every resume.

The same runner on the `skill` tier, with no init containers, is ready in
**1.28s**.

**A finished conversation keeps its sandbox for up to an hour.** A live runner
against a real model measures ~505 MiB in cgroup, of which 233.9 MiB is private
(SDK-bundled Claude Code child 259.5 MiB RSS, Python runner 74.1 MiB) and
272.4 MiB is page cache shared per node. So an idle thread holds roughly
**334 MiB of marginal memory at 0.043% of a core** for the remaining ~59 minutes
of `routeTtlSeconds: 3600`.

## Why the warm pool does not absorb this

`agentSandbox.warmPool.replicas` defaults to **0**, and the chart records that
this is deliberate: a real-model warm pod has no bundle, raises
`PluginBundleError`, and crash-loops, and **per-claim env injection forces a cold
create anyway** (the `envVarsInjectionPolicy: Overrides` gotcha). Both halves
were reproduced while writing this: `PluginBundleError: invalid plugin bundle at
/unused`, twice.

The mechanism is the point. A sandbox's capability is decided by pod environment
injected per claim, so a pod that does not yet know its claim cannot be useful.
The fast-bind path exists in the substrate and is architecturally unreachable.

## Why this is an availability decision

The chart records an incident twice: a Langfuse merge burst took ~175% CPU with
0.0% idle on a 2-vCPU node, every new sandbox pod missed the 90s claim deadline,
**three attempts timed out and the turn escalated to a human as an opaque
`runner-error`**, and nothing flagged it because CPU starvation sets no node
condition. Capping ClickHouse at one core "halved the blast radius and left the
cause untouched", and the same escalation returned.

The chart also names the amplifier: under contention the kernel divides CPU by
**requests**, and the critical-path containers request 50m each. Confirmed on the
live release -- runner, `bundle-fetch`, and `bundle-extract` all request `cpu:
50m`, while ClickHouse requests `200m` and may burst to a full core.

The general defect: **an absolute wall-clock deadline sits on the path with the
smallest relative resource share in the cluster.** 17.39s starved 4:1 lands near
70 seconds against a 90-second ceiling. Making the bind sub-second removes the
deadline from reach; 0.5s starved 4:1 is 2 seconds.

## The decision

Session identity is delivered over the ACI after a sandbox is bound, not through
pod environment at claim time. Pod env carries version-scoped capability;
`CURIE_SESSION_ID`, `CURIE_HISTORY_REF`, and the runner token become an ACI call
against a bound runner. That is what makes a version-keyed warm pool bindable,
which is what makes short residency affordable.

Full decision list, business case, evidence, and rejected alternatives are in the
ADR. The design doc carries the measurement method, the seams each workstream
touches, and the ordering constraint that matters most: **version-keyed pools do
not work before session identity moves off pod env.**

## A spike killed an alternative before it landed here

An earlier draft of this reasoning claimed many-sessions-per-runner-process was
worth 6.6x on memory. Measured: each session is served by its own SDK-bundled
Claude Code child at 259.5 MiB RSS, so sharing the process amortises only the
74.1 MiB Python runner -- **1.25x at ten sessions**. It is recorded as a rejected
alternative rather than a proposal.

## Declared versus measured, per sandbox

| Axis | Chart request | Measured | Note |
| --- | --- | --- | --- |
| CPU | 50m | 0.43m idle, ~90m active | **116x over-reserved at idle** |
| memory | 192Mi | ~505 MiB cgroup | **under-reserved; the 768Mi limit carries it** |

The CPU over-reservation is what caps density on a small node: 1735m of platform
requests plus 50m per sandbox leaves room for about five sandboxes on two cores.

## Reviewing this

Every figure is reproducible; the design doc's "What was measured, and how to
re-measure it" section carries the exact commands, including how to drive a real
streaming turn through a host Ollama at no API cost, and why any residency figure
taken under `--fake-model` should be discarded (it understated the runner by
roughly 5x).

## Still outstanding for the assignment

- A demo recording. The design doc specifies what it has to show: not "faster",
  but that the deadline stops being reachable under reproduced contention.
- An implementation slice, which needs the ADR Accepted first.


docs/adr/0114-session-identity-arrives-over-the-aci-so-a-sandbox-can-be-pre-bound.md
```diff
@@ -0,0 +1,553 @@
+# 114. Session identity arrives over the ACI so a sandbox can be pre-bound
```
TheConnMan: This looks good. Not a wild moonshot idea, but a true performance issue. Please strip the prototype/demo/design out, we'll approve this PR, and you can fully implement it.

jw3329:
Continued in #1804. GitHub closed this one when the branch was renamed `task/adr-0114-sandbox-residency` -> `task/adr-0116-sandbox-residency`; the commits and head are unchanged (`d9367b16`). The ADR moved to 0116 because `spike/tier-diff` dated its ADR a day earlier and also claimed 0114.

Issue #1779, opened 2026-08-20
cluster up --local-model pulls weights implicitly via postStart, onto an emptyDir by default


alexrao2000:
## Summary

At the `cluster` tier, `--local-model` downloads the model weights inside the
cluster with no preflight, no size announcement, and no way to distinguish the
wait from a wedge. On the chart's default settings the weights land on an
`emptyDir` and are re-downloaded on every container restart.

This is the same failure shape as #1183, which [ADR-0093](https://github.com/curie-eng/curie/blob/main/docs/adr/0093-local-model-assets-are-pre-provisioned-never-implicitly-downloaded.md)
fixed for the `skill` and `local` tiers. ADR-0093 excluded the cluster tier on the
stated premise that "`cluster up` never faces the question: it does not run
Ollama." That premise no longer holds, and per ADR-0045 the ADR text correctly
stays as written -- which is why this is filed as an issue rather than an edit.

## Mechanism

`cluster up --local-model <model>` sets `inference.deploy=true` and
`inference.model=<model>` (`cli/src/ops.rs`, `up_value_plan`). The chart then
renders the Ollama Deployment, and with `inference.pullModel` (default **true**)
it pulls the weights from a **`postStart` lifecycle hook**
(`charts/curie/templates/inference.yaml`):

```sh
for i in $(seq 1 60); do
  if ollama list >/dev/null 2>&1; then
    ollama pull "<model>"
    exit 0
  fi
  sleep 2
done
exit 1
```

Four consequences, none of them announced:

1. **The pull is unbounded and invisible.** kubelet does not mark a container
   started until `postStart` returns, and the readiness probe does not begin
   until then either, so the pod sits not-ready for the whole download. The probe
   settings (`initialDelaySeconds: 5`, `failureThreshold: 12`) never apply to it.
2. **A failed pull kills the container.** A non-zero `postStart` causes kubelet to
   kill the container, so a network error or a bad model name becomes
   `CrashLoopBackOff` with the cause buried in a `FailedPostStartHook` event --
   and each restart retries the whole download.
3. **The default loses the weights on restart.** `inference.persistence.enabled`
   defaults to **false**, so the template mounts `emptyDir: {}`. `cluster up`
   sets only `inference.deploy` and `inference.model`, so the default path gets no
   PVC: any restart, eviction, or node drain re-downloads the model in full.
4. **`set -eu` conflates two failures.** A server that never came up and a pull
   that failed exit the same way with no distinguishing message.

[...]

Issue #1666, opened 2026-08-18
Verb parity: the same verb takes different flags and defaults at different tiers


TheConnMan:
The parity ladder promises the same bundle behaves the same at every rung. The bundle does. The verbs do not, and the divergences are silent.

## Observed

| Verb | `skill` | `local` / `cluster` |
| --- | --- | --- |
| `message` conversation default | always continues; no `--continue` or `--thread` exist | starts a new thread; `--continue` reuses the last turn |
| `message` output | streamed tokens, per-tool lines | transport narration, then one final block |
| starting a fresh conversation | not possible without `up --replace` | `reset-thread`, or omit `--continue` |
| `--model` | flag on `up` | no flag, `CURIE_MODEL` env only |
| model credential | shell env, vault, or `--env-file` | shell env only |

The conversation default is inverted between rungs, so a promoted bot silently stops remembering. Reported during a hands-on pass as: "we have to do a scrub for consistency about what the default flags and behavior are for each of the verbs, it needs to be identical."

Output volume also differs sharply with no warning: `skill message` narrates almost nothing, `local message` narrates the queue, `cluster message` narrates port-forwards and helm. The operator never asked for plumbing and cannot dial it per-command except with a blanket `-q`.

## Proposed rule

A verb that exists at more than one tier takes the same flags and defaults at every tier where it exists, or it is named differently. Where a tier genuinely cannot support a flag, exit 4 (unsupported-at-this-tier) naming the alternative, the pattern already used for `skill versions` and `skill memory` per ADR-0041.

Default output should be quiet, with `--debug` opting into transport narration.

## Mechanically checkable

`curie dev field-parity` and `emit-parity` already exist. The same shape would catch flag and default divergence across tiers, so this does not have to rely on review attention.

Related: #112 introduced `--continue` for cluster/local and never considered parity with the skill tier's implicit continue; #361 settled the model question on env only.

TheConnMan:
Completed by merged PR #1783. Closing manually because the implementation landed on the non-default release-train branch or the closing reference did not auto-fire.

Issue #1233, opened 2026-08-03
Consider letting --local-model target a host-installed Ollama instead of always containerizing it


yichenzhang-curietech:
## The idea

`--local-model` currently always runs Ollama **in a container**. `local up --local-model` starts the compose `ollama` service and points the worker at `http://ollama:11434` (hardcoded in `cli/src/local.rs`); `skill up --local-model` starts its own `ollama` container and points the runner at `http://<container>:11434`. Neither tier can be told to use an Ollama the operator already has running on the host.

That matters because of what the container path costs on a cold machine, measured while investigating #1183:

| asset | size |
|---|---|
| `ollama/ollama:0.24.0` docker image | **8.93 GB** |
| default `qwen3:4b` model weights | **2.5 GB** |

Anyone who actually wants a local model very likely already has Ollama installed and models pulled into `~/.ollama`. For them, both downloads are pure duplication: the same weights, fetched a second time, into a docker volume, so a container can serve them over the same port the host is already serving.

Letting `--local-model` target a host Ollama (`http://host.docker.internal:11434`, or an explicit `--model-base-url`) would skip **all ~11.4 GB** and make the first `--local-model` run essentially instant on such a machine.

## Why this is a Discussion-shaped question, not a ready task

It is a feature, not a bug fix, and it pushes on things this project deliberately constrains:

- **The egress posture.** Sandboxes run default-deny outbound. A runner reaching the host's Ollama is a new hole in that, and the chart's guardrails are defaults rather than options on purpose. What exactly gets opened, and at which tier, needs deciding rather than assuming.
- **The parity ladder.** `skill`, `local` and `cluster` are supposed to differ only in the infrastructure underneath. A host-provided model is available at `skill` and `local` and meaningless at `cluster`, so this either becomes a tier-specific escape hatch (and the ladder gets a documented asymmetry) or it does not ship.
- **Reproducibility.** The containerized Ollama pins `ollama/ollama:0.24.0`. A host Ollama is whatever version the operator happens to have, which is exactly the "works on my machine" gap the ladder exists to close. `docs/local-model.md` already warns that Ollama versions matter.
- **Discovery and failure modes.** Auto-detect a host Ollama, or require an explicit opt-in flag? Silent auto-detection would make the same command mean different things on two machines.

## Relationship to #1183

[...]

PR #1780, opened 2026-08-20
Correct the cluster tier's local-model download claim

alexrao2000:
## What this is

A docs-only correction. `docs/local-model.md` currently tells operators that the
cluster tier has nothing to download, and that is false today.

Filed alongside #1779, which carries the behaviour defects and the decisions they
need.

## The wrong claim

> `cluster up --local-model` is unaffected: it has no host-side Ollama to
> provision, so there is nothing to preflight and no `--pull-model` flag.

The first clause is true. The conclusion is not: there is no *host-side* Ollama
because the chart runs Ollama *in the cluster*, and it pulls the weights there
implicitly.

## What actually happens

`cluster up --local-model <model>` sets `inference.deploy=true` and
`inference.model` (`cli/src/ops.rs`, `up_value_plan`). With
`inference.pullModel` (default `true`) the chart pulls from a **`postStart`
lifecycle hook** on the Ollama container, which means:

- **the pod is not-ready for the entire download.** kubelet does not mark a
  container started until `postStart` returns, so the readiness probe has not
  begun and the wait is unbounded. This is the #1183 failure shape ADR-0093
  removed at the other two tiers;
- **a failed pull kills the container.** A non-zero `postStart` makes kubelet
  kill it, so a network error or an unknown model name becomes
  `CrashLoopBackOff`, cause buried in a `FailedPostStartHook` event, and each
  restart retries the whole download;
- **the default discards the weights.** `inference.persistence.enabled` defaults
  to `false` and `cluster up` requests no PVC, so the data directory is an
  `emptyDir` and any restart or eviction re-downloads in full.

`qwen3:4b` (~2.5GB) hides all of it. The documented upgrade `qwen3-coder:30b` is
~17-19GB, where it does not hide at all.

## What this PR changes

Only the prose. It corrects the claim, adds a **"The cluster tier downloads
implicitly"** subsection under "How it runs" describing the three behaviours
above, and gives the interim mitigation for a large model:

```bash
curie cluster up --local-model qwen3-coder:30b --set inference.persistence.enabled=true --set inference.persistence.size=40Gi
```

That routes through `curie cluster up --set` (`ClusterAction::Up`) rather than a
raw `helm upgrade`, per the one-entry-point rule in `CLAUDE.md`. My first draft
had the raw helm command; this is the corrected form.

## What this PR deliberately does not change

**ADR-0093's text.** It excluded the cluster tier on the premise that
"`cluster up` never faces the question: it does not run Ollama", which no longer
holds. Per ADR-0045 that stays exactly as written -- the status line, a
back-link, and pointer repairs are the only permitted edits, and "the gap between
what was decided and what was built is the record's whole point". The stale
premise is reported in #1779 instead.

**Any chart default or behaviour.** Whether this tier gets a preflight, whether
`pullModel` keeps defaulting to true, whether the pull moves out of `postStart`,
and whether `persistence` should default on when inference is deployed are four
separate calls that each change an operator-visible default. They are enumerated
in #1779 for a maintainer rather than picked here.

## Related

- #1779 the behaviour defects and decision points (filed with this PR)
- #1666 verb parity: same verb, different flags and defaults per tier -- this is
  a concrete instance, since `--pull-model` exists at two tiers and not the third
- #1183 the original silent-download report that produced ADR-0093
- #1233 letting `--local-model` target a host-installed Ollama
- ADR-0093, ADR-0045, and ADR-0041 (every verb answered at every tier)

## Verification

`bash scripts/check-docs.sh` passes: no catalog drift, every citation resolves.
`scripts/check-commit-messages.sh` clean. No code paths touched, so behaviour is
unchanged by construction.

### DIFF
--- a/docs/local-model.md
+++ b/docs/local-model.md
@@ -65,8 +65,13 @@ download from a wedged one (issue #1183). Once both assets are cached the flag i
 never needed again and `up` is unchanged — measured at **17.9s warm**, against
 **232s** for the same command on a cold machine over a fast link.
 
-`cluster up --local-model` is unaffected: it has no host-side Ollama to
-provision, so there is nothing to preflight and no `--pull-model` flag.
+`cluster up --local-model` has **no** such preflight and no `--pull-model` flag,
+because there is no host-side Ollama to provision. That is not the same as
+nothing being downloaded: the chart's inference Deployment pulls the weights
+*inside the cluster* instead, and does it implicitly. See
+[the cluster tier](#the-cluster-tier-downloads-implicitly) below before running
+this against a real cluster with a large model
+([#1779](https://github.com/curie-eng/curie/issues/1779)).
 
 ## How it runs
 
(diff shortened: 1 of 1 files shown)


TheConnMan (changes requested):
The correction is directionally right, but please amend the persistence bullet before merge. It says that "Any restart, eviction, or node drain re-downloads the model in full." The chart mounts Ollama data from an `emptyDir`; Kubernetes preserves that volume across container crashes, so a container restart within the same Pod does not lose the weights. Limit the claim to Pod removal/replacement (such as eviction or node drain), and avoid claiming a full re-download/retry unless demonstrated by Ollama.

Reference: https://kubernetes.io/docs/concepts/storage/volumes/#emptydir

alexrao2000:
Addressed in [f7285532](https://github.com/curie-eng/curie/commit/f7285532): the persistence bullet now says a container restart within the same Pod does not lose the weights (Kubernetes preserves the `emptyDir` across that), and limits the loss to Pod removal/replacement — eviction or node drain — where the replacement Pod's `postStart` pulls the weights again. Also updated the follow-on sentence ("deploy a large model with persistence... so an evicted or rescheduled Pod does not re-fetch it") to match, and dropped the unqualified "in full" language.

TheConnMan (approved):

Issue #411, opened 2026-07-15
Approval resolve strands the suspended session if the resume enqueue fails (no outbox/reconciler)


TheConnMan:
## Summary (🟠 MAJOR — correctness)

The approval resolve endpoint commits the compare-and-set claim **before** enqueuing the resume turn. If the enqueue fails, the approval is permanently resolved in Postgres but the suspended session is **never woken** — and there is no retry path or reconciler.

## Root cause

`apps/api/src/agentos_api/routers/approvals.py` (~L177-209): the CAS claim (`crud.claim_approval_resolution`) commits, then `resume_queue.enqueue(...)` runs. If `xadd` raises (Valkey blip/restart), the endpoint 500s but the record is already `approved`/`rejected`. Any retry hits the CAS-lost branch → 409 "already resolved by …" with no re-enqueue. `grep` confirms `enqueue` is called only here; no reconciler exists.

## Failure scenario

Human clicks **Approve** → DB flips to `approved` → Valkey drops the write → the agent stays suspended forever and the click can never be replayed.

## Fix direction

Make [claim → audit → enqueue] atomic via a **transactional outbox** (write the resume intent in the same transaction as the claim, drained by a relay), **or** add a **reconciler** that periodically re-enqueues resumes for records where `status != pending AND resolved_at IS set AND not-yet-resumed`. (Resume turns already carry a deterministic `event_id`, so a duplicate enqueue is absorbed by the worker's done-marker — re-enqueue is safe.)

## Provenance
- PRs: #406 (issue #244), #409 (issue #247)
- ADR-0010 frames the lifecycle as suspend → resume-on-resolution; this is a dead-end branch on that promise.

Found during post-merge review 2026-07-15.

Issue #410, opened 2026-07-15
Security: sandboxed agent can self-resolve its own approval (shared platform key forwarded into sandbox)


TheConnMan:
## Summary (🔴 HIGH — security)

A sandboxed agent can **resolve its own approval**, defeating the ADR-0010 tool-level permission gate. The hole emerges from the *composition* of three separately-merged PRs, so no single reviewer flagged it.

## Root cause

1. The worker forwards the **one shared platform API key** into the sandbox as `AGENTOS_MEMORY_TOKEN` / `AGENTOS_HISTORY_TOKEN` (needed for memory/history rehydration) — `apps/worker/src/agentos_worker/binding.py:227,239`. Comment admits "the state API has one shared key today."
2. `/approvals/{id}/resolve` and `/agents/{id}/state/...` are guarded by the **same** `require_api_key` (`apps/api/src/agentos_api/auth.py:17`) — one shared key.
3. The authorizer trusts caller-asserted `resolved_by` / `actor_channel` (`apps/api/src/agentos_api/authorizer.py`).

## Exploit

An agent reads `AGENTOS_HISTORY_TOKEN` (= the platform key) from its own env and POSTs `/approvals/{id}/resolve` with `X-API-Key: <that key>`, a fabricated non-author `resolved_by`, and `actor_channel = card_channel` → self-approves its own gated tool call. The egress allowlist does **not** mitigate: the sandbox must reach the API host for memory to work at all. Same key also grants read/write to **any** agent's memory + transcripts (cross-agent isolation break) and lets the sandbox resolve **any** pending approval platform-wide.

This directly undercuts ADR-0010's guarantee (rejected alt #3: a permission gate must "reliably block a model-initiated action").

## Fix direction

Issue a **scoped, least-privilege sandbox token** that can touch only that agent's memory/transcript namespace and **cannot** resolve approvals — or never place a resolve-capable credential in the sandbox env at all. The authorizer must not trust caller-asserted identity for any credential the sandbox holds.

## Provenance
- Approval gate: #406 (issue #244), #407 (#245), #409 (#247)
- Token forwarding: #402 (#20), #379 (#264)
- ADR: docs/adr/0010-approval-gates-and-human-in-the-loop.md; INTERFACE: docs/interfaces/approval/INTERFACE.md ("enforced server-side ... cannot be spoofed from inside the agent's box")

Found during post-merge review 2026-07-15. Must land before the approval e2e demo is shown with a real threat model.

Issue #269, opened 2026-07-10
Triggers: generic HMAC-verified inbound webhooks (POST /hooks/{agent}/{hook})


alexrao2000:
Part of #29 (triggers beyond chat epic). Same verification pattern as the existing GitHub webhook.

## Scope
- Generic **HMAC-verified inbound webhooks**: `POST /hooks/{agent}/{hook}` mapping payloads to events on the stream.

## Acceptance
A signed request to `/hooks/{agent}/{hook}` enqueues an event; an unsigned/bad-signature request is rejected.

alexrao2000:
Consumes the trigger declarations defined in the bundle manifest by #273 (schema + deploy-time validation).

TheConnMan:
## Secure webhook trigger contract

Treat this as an event producer for the existing stream, not a separate execution path.

Before enqueueing, verify the signature, timestamp or nonce, replay identity, body size, manifest declared route, and target version. Return success only after the durable append succeeds. Bad signatures, replayed payloads, and oversized bodies must have distinct rejection reasons without echoing sensitive payload data.

The queued event should retain an auditable source identity and version or bundle reference. Its eventual result must be visible through the same execution and trace surfaces as a Slack or CLI turn.

Related: #29, #1040, #1278

yichenzhang-curietech:
Decision of record: [ADR-0079: Inbound triggers as a new event kind, ingested by the API](https://github.com/curie-eng/curie/blob/next/docs/adr/0079-inbound-triggers-as-a-new-event-kind.md) (Accepted 2026-07-10). Linking it because the repo's rule is that an issue implementing a decision references its Accepted ADR, so whoever picks this up reads the intent first. In particular ADR-0079 already settles two things this issue does not restate: the API accepts the hook (not the Socket-Mode dispatcher), and a trigger is a new event kind with no placeholder.

Prerequisite: #1631 lands the `source` field and the kernel's placeholder-less output path, which both children build on. Per ADR-0079, no partial webhook code should merge before the kernel side.

TheConnMan:
Shipped on `next` in #1632, milestoned v0.7.0. Closing by hand: a closing keyword in a PR targeting `next` never fires, because GitHub only honours it on a merge into the default branch (`main`). This will ship with the v0.7.0 cut.

PR #1793, opened 2026-08-21
ADR-0115: agents call each other with no third party in the path

alexrao2000:
## What this is

A **Draft ADR** proposing how one Curie agent calls another. Draft only: per
[ADR-0085](https://github.com/curie-eng/curie/blob/main/docs/adr/0085-acceptance-not-implementation-authorizes-an-adr.md)
it authorizes no implementation until a maintainer accepts it. No code changes --
the ADR plus the regenerated `docs/adr/README.md` index.

## Why now

Pillar 4 in `docs/pillars.md` already says agents communicate "with people and
with other agents", and nothing implements the second half.

**To be unambiguous, since turn-minting machinery landed recently and this could
be misread as a refinement: agents cannot communicate today, at all.** A sandbox
is handed exactly one API credential, `CURIE_STATE_TOKEN` (scope `state.app`),
and the API accepts a scoped token on the state router and nowhere else -- "a
scoped token is rejected everywhere else" (`apps/api/CLAUDE.md`). The platform
key is never placed in the sandbox. An agent therefore holds no credential that
any turn-minting route will accept. There is no partial path and nothing to widen
by configuration.

What recent work changed is the *material* a solution can be built from, not the
reachability. Two ingress paths now mint turns, and both are platform-key gated
(`POST /channels/token` sits under `require_api_key`; a hook signature derives
from `settings.api_key`). Handing a sandbox either credential would not be a
shortcut, because each surface also has properties that make it the wrong one to
open up:

- The hook ingress (#269, ADR-0079) pins `author` to `hook:{hook}` precisely so
  an upstream cannot supply an identity, routes the reply to the target's channel
  so a caller never sees the answer, and shares one conversation per hook so
  concurrent calls serialize. A caller, a private answer, and concurrency are the
  three things a call needs; it refuses all three deliberately.
- The channel ingress (#1459, ADR-0096) scopes a `chn` token to a binding row, so
  it answers "which binding may this enqueue for" and never "who is asking".
- Slack is closed on purpose: `is_user_message` drops any event with a `bot_id`,
  and it has to, or the dispatcher's own placeholder would re-trigger it.

Every one of those is deliberate, which is why this needs a decision rather than
a patch. Two consumers are already waiting: ADR-0044 (Accepted) committed to a
"callable step" as its phase-1 build and assumes an API call that runs a turn and
returns a result, which does not exist; and discussion #1372 (LangChain) wants
the same primitive.

## What you have to do today, and why removing it is the point

There is a workaround, and naming it is the clearest statement of what this ADR
is worth. To make two Curie agents collaborate today you must stand up something
outside Curie: a process holding the **platform key** -- the most privileged
credential in the system -- that mints turns for each agent through
`/channels/turns` or `/hooks/{agent}/{hook}`, then recovers each answer by
watching the agents' channel bindings, because neither route returns a result.

That relay is the third party. It costs you the platform key in a new place, an
extra service to run and secure, attribution that stops at "the platform"
(`author=hook:{hook}`), no loop or depth bound, no approval integration, and
answers scraped out of human channels. Every one of those is a direct consequence
of using surfaces built for external upstreams to carry an internal call.

So the notable claim is not "agents can now talk" in the abstract. It is that
they talk with no third party in the path, no additional distribution of the
platform key, a caller and an accountable principal carried on the turn, and the
approval and loop controls the relay cannot provide.

## The decision, in one line

**One agent may ask another agent to do something, over a surface the platform
owns end to end, with no third party in the path.** No interop protocol to adopt,
no external broker, no orchestration framework, no bundle-shipped MCP server, no
relay holding the platform key, and no round trip through a human channel. It
reuses machinery the platform already has -- the `curie:runs` stream, the resume
path, the ADR-0033 scoped token, the runner's existing MCP mounting -- so a
second agent becomes reachable with no new infrastructure and no new dependency
to trust.

The ADR treats that as a constraint as well as a property: if this decision ever
requires a third party, it is the wrong decision and should be revisited rather
than extended.

Everything else in the ADR is how that is made safe, under one governing rule:
**a call must never widen what the target may do, and must never lose who
asked.** Concretely that means the target keeps its own credentials, model,
egress, and approval policy; the turn carries both the immediate caller and the
accountable human principal; a call can never satisfy an approval gate; and the
surface is default closed with bounded depth and refused cycles.

The three things a naive version would quietly do -- launder authority, lose
accountability, and loop -- are each addressed explicitly, with #410
("sandboxed agent can self-resolve its own approval") as the precedent for why
authority must not be assertable from inside a sandbox.

## Sequencing

- **Sequences behind ADR-0112 / #1672.** Delegation adds a fifth `QueuedTurn`
  producer, and 0112 states every producer added before its rename widens the
  blast radius. 0112 also already settled the enum question here: it rejects
  `human` precisely because "ADR-0079's own agent-to-agent case ... [produces]
  turns caused by an agent speaking as a participant".
- **Does *not* sequence behind ADR-0099** (hooks as bundle-declared turns). A
  hook fire is a job with no caller and no audience; a call has both. The ADR
  argues against merging them.
- **Needs discussion #1049** for the identity half. The ADR specifies attribution
  at the payload and run-record level, which holds today, and does not claim the
  IdP-issued agent identity #1049 asks for.

## Referenced and prerequisite issues

**Prerequisites / open questions**
- Discussion [#1049](https://github.com/curie-eng/curie/discussions/1049) -- agent identity, invocation authorization, caller/owner attribution. The authority prerequisite; its requirement 3 (accountability) and requirement 1 (invocation authorization) are what the Decision implements, and its on-behalf-of non-goal stays a non-goal here.
- [#1672](https://github.com/curie-eng/curie/pull/1672) -- ADR-0112, turn source naming. Must land first.
- Epic [#158](https://github.com/curie-eng/curie/issues/158) -- multi-tenancy. Cross-tenant calls are out of scope until a tenant object exists; #158 must revisit cost attribution.
- [#82](https://github.com/curie-eng/curie/issues/82) -- per-tenant identity before any multi-tenant deployment.

**Foundations this builds on (all closed)**
- [#269](https://github.com/curie-eng/curie/issues/269) -- the HMAC hook ingress, `POST /hooks/{agent}/{hook}`.
- [#1459](https://github.com/curie-eng/curie/issues/1459) -- channel port promotion; the `kind`/`address` routing pair and the sink selection the return path extends.
- [#410](https://github.com/curie-eng/curie/issues/410) -- sandboxed agent could self-resolve its own approval via a forwarded platform key. The reason ADR-0033's scoped token exists, and the precedent for "authority is never asserted from inside a sandbox".
- [#244](https://github.com/curie-eng/curie/issues/244) -- durable Approval record and the suspend/resume lifecycle the caller's resume reuses.
- [#411](https://github.com/curie-eng/curie/issues/411) / [#419](https://github.com/curie-eng/curie/issues/419) -- the approval seam and its resume-enqueue reconciler.
- [#248](https://github.com/curie-eng/curie/issues/248) / [#249](https://github.com/curie-eng/curie/issues/249) (epic [#23](https://github.com/curie-eng/curie/issues/23)) -- the durable state store and the auto-mounted MCP server pattern (ADR-0073) this surface copies.
- [#19](https://github.com/curie-eng/curie/issues/19) -- per-thread reply routing; why `ReplyHandle` carries a per-turn endpoint at all.
- [#1060](https://github.com/curie-eng/curie/issues/1060) -- an agent reading its own runs (ADR-0107), the sibling agent-facing scoped surface.

**Related / where the work would land**
- Epic [#29](https://github.com/curie-eng/curie/issues/29) -- triggers beyond chat. The hook half is done; this is the caller half.
- Epic [#30](https://github.com/curie-eng/curie/issues/30) -- bundle format and authoring extensions, where the `delegatesTo` declaration would land.
- Discussion [#1372](https://github.com/curie-eng/curie/discussions/1372) -- LangChain support; second consumer of the same primitive.
- [#1525](https://github.com/curie-eng/curie/issues/1525) -- one agent binding multiple channels, adjacent to the return-path `kind`.
- Discussion [#1176](https://github.com/curie-eng/curie/discussions/1176) -- the pillar taxonomy whose pillar 4 names this gap.
- [#521](https://github.com/curie-eng/curie/issues/521) -- ADR number collisions. See the numbering note below.

## Numbering note

Numbered **0115**. 0112 (#1672), 0113, and 0114 are claimed; 0114 is currently
claimed by *two* unmerged branches, which is the #521 collision class the
`check-docs.sh` gate exists to catch. Happy to renumber if a maintainer is
sequencing these.

## Verification

`bash scripts/check-docs.sh` passes: ADR numbers unique, no catalog drift, every
citation resolves. The regenerated `docs/adr/README.md` is committed rather than
hand-edited, per `docs/adr/AGENTS.md` step 6.

ADR-0112 is cited by number rather than relative link because its file is not on
`main` yet and a link would fail the citation lint; worth converting to a link
once #1672 merges.

## Review asks

1. **Is a new first-party surface the right call**, versus widening one of the
   two existing platform-key-gated ingresses? The ADR argues those surfaces
   refuse a caller, a private answer, and concurrency on purpose, so widening
   them would leave a different route wearing their name.
2. **Two attribution fields** (immediate caller + accountable principal) rather
   than one -- is that the right split given #1049 is still open?
3. **Suspend/resume rather than a blocking call.** Rejected blocking because a
   target can suspend on an approval gate for a human-shaped interval.
4. **Default closed, bundle declares / operator arms** -- consistent with
   ADR-0056 and ADR-0098, but it is a new authoring surface in epic #30.
5. The kernel steering rule needs a second predicate (a delegated turn is a
   `message`, so `is_job` is False). That is single-owner kernel territory and
   flagged in Consequences rather than decided here.

## Why this PR exists

#1774 merged to `main` by accident and was reverted by #1792 (now merged). This
restores the same Draft ADR the revert removed, content unchanged, held as a draft
PR so it lands when a maintainer chooses to merge it rather than by accident. The
review asks and the numbering note above still stand.

### DIFF
--- a/docs/adr/0115-agents-call-each-other-with-no-third-party.md
+++ b/docs/adr/0115-agents-call-each-other-with-no-third-party.md
@@ -1,0 +1,413 @@
+# 115. Agents call each other directly, with no third party in the path
+
+Date: 2026-08-20
+
+Status: Draft
+
+## Context
+
+Curie has no way for one agent to ask another agent to do something. Pillar 4
+in [`docs/pillars.md`](../pillars.md) already says agents communicate "with
+people and with other agents", and nothing implements the second half.
+
+This is not a missing feature so much as a missing decision. Turn-minting
+machinery landed recently, which makes it tempting to describe the gap as almost
+closed. It is not: **an agent cannot reach any of it.** The gap is total, and the
+recent work changes only what a solution can be built out of.
+
+### What exists today, and why none of it is reachable
+
+**The blocking fact first.** A sandbox is handed exactly one API credential:
+`CURIE_STATE_TOKEN`, a `state.app`-scoped ADR-0033 token minted per claim
+(`apps/worker/src/curie_worker/binding.py`). The API accepts a scoped token on
+the state router and nowhere else -- "a scoped token is rejected everywhere else"
+(`apps/api/CLAUDE.md`) -- and the platform key is deliberately never placed in
+the sandbox. So an agent holds no credential that any turn-minting route will
+accept. There is no partial path, no lossy path, and nothing to widen by a
+configuration change. Everything below is about which surface a *new* credential
+should authorize, not about a surface an agent can use today.
+
+Since [#269](https://github.com/curie-eng/curie/issues/269) closed,
+[ADR-0079](0079-inbound-triggers-as-a-new-event-kind.md) is implemented. A turn
+can start from something other than a person: `QueuedTurn.source` distinguishes
+a person's message from a job, `ReplyHandle.placeholder` is nullable, and
+`POST /hooks/{agent_id}/{hook}` (`apps/api/src/curie_api/routers/hooks.py`)
+turns a signed external event into a turn on the same stream a mention feeds.
+[#1459](https://github.com/curie-eng/curie/issues/1459) then landed the channel
+port from [ADR-0096](0096-port-adapters-are-deployed-services.md), so
+`POST /channels/turns` mints a turn for a binding named in the body, holding a
+`chn` token scoped to one binding row rather than the platform key.
+
+So two ingress paths mint turns today, and both are gated on the platform key:
+`POST /channels/token` mints a `chn` token under `require_api_key`, and a hook
+signature derives from `settings.api_key`. Handing either credential to a sandbox
+is not a shortcut to this feature, because each surface also has properties that
+make it the wrong thing to open up -- properties that are deliberate, and load
+bearing where they sit:
+
+- **The hook ingress cannot carry a caller, by construction.** `_mint_turn` sets
+  `author=f"hook:{hook}"`, and the code says why: an upstream-supplied identity
+  there "would let a hook impersonate one to anything downstream that reads the
+  field". The reply route comes wholly from the target's binding row, so the
+  answer goes to the target's channel and a caller never sees it. And
+  `_conversation_id` is per hook, so two calls would share one thread and the
+  second would defer behind the first. A caller, a private answer, and
+  concurrency are the three things a call needs, and this surface refuses all
+  three on purpose.
+- **The channel ingress is for adapters, not agents.** A `chn` token is scoped
+  to a binding row so an ingress adapter can enqueue for that binding. It says
+  nothing about who asked, and the same reply-routing property applies: the
+  platform mints the route from the binding, never from the request.
+- **Slack is closed on purpose.** `is_user_message` in
+  `apps/dispatcher/src/curie_dispatcher/handlers.py` drops any event carrying a
+  `bot_id`. An agent posting into a channel cannot wake another agent. Without
+  that filter the dispatcher's own placeholder would re-trigger it, so the filter
+  is load-bearing and not an oversight to remove.
+
+Two other things exist and are worth naming because they are often mistaken for
+this feature. Subagents inside a single turn already work: the bundle manifest
+carries an `agents` field (`packages/plugin-format/src/plugin_format/models.py`)
+that the harness picks up. That is one agent identity, one pod, one credential
+set, one approval policy -- decomposition, not two agents talking. And the
+`curie-state` MCP server ([ADR-0073](0073-agentos-state-mcp-server-and-state-boot-env.md))
+is namespaced per agent with a token bound to that agent, so two agents cannot
+even use it as a shared blackboard.
+
+### Two decisions already made that this one has to fit
+
+ADR-0112 renames the person-message value of `source` from `slack` to `message`.
+It is a Draft and is not on `main` yet, so it is cited here by number rather than
+by relative link ([PR #1672](https://github.com/curie-eng/curie/pull/1672)). Its
+rejection of `human` names this case explicitly: "ADR-0079's own agent-to-agent
+case, and the Twin agent pattern, both produce turns caused by an agent speaking as a
+participant rather than by a job firing. Those are messages and are not human."
+That settles the enum question here before it is asked. An agent asking another
+agent something is a message, not a job, and needs no fourth `source` value.
+
+[ADR-0099](0099-hooks-are-bundle-declared-turns-the-system-starts.md) (Draft)
+covers the other half of "a turn no user started": bundle-declared background
+work on a schedule. It is easy to assume delegation belongs there. It does not.
+A hook fire is a job with no caller and no audience; a delegated call has both.
+This ADR is therefore not sequenced behind ADR-0099, and the two should not be
+merged into one mechanism.
+
+### Two consumers are already waiting
+
+[ADR-0044](0044-workflow-controlled-agents-callable-substrate.md) (Accepted)
+committed to shape C as its phase-1 build: a developer's LangGraph or CrewAI
+graph stays the outer loop and calls into Curie at the steps that need
+open-ended AI, folding the result back into graph state. That ADR assumes the
+developer can "call the existing Curie API to run a sandboxed, approval-gated
+turn". No such call exists. The hook and channel ingresses mint a turn and
+return a receipt; neither returns a result. Discussion
+[#1372](https://github.com/curie-eng/curie/discussions/1372) (LangChain support)
+wants the same primitive.
+
+So the missing piece is one primitive with two callers: an external workflow, and
+another agent. The authorization question is much sharper in the second case,
+which is why this ADR is scoped to it and the external caller inherits the answer.
+
+### The question this ADR has to answer
+
+Discussion [#1049](https://github.com/curie-eng/curie/discussions/1049) (agent
+identity, invocation authorization, and caller/owner attribution) is the
+prerequisite conversation, and it frames the hard part. Its requirement 3 is that
+every action be traceable to an accountable person -- the caller when a human
+asked, the registered owner when a schedule fired -- and "never to just 'the
+agent' with no accountable person behind it". Its requirement 1 is that not
+everyone may invoke every agent. It explicitly lists delegated, on-behalf-of
+impersonation as a non-goal.
+
+A call between two agents is where those requirements bite hardest, because a
+naive implementation quietly does three things nobody decided:
+
+1. **It launders authority.** If B's turn runs with no record of who asked, then
+   a caller with no permission to do something can ask an agent that has it. The
+   approval machinery ([ADR-0010](0010-approval-gates-and-human-in-the-loop.md),
+   [ADR-0034](0034-approval-authorizers-resolve-membership-in-the-api.md),
+   [ADR-0035](0035-one-shot-post-approval-allowance.md)) resolves approvers
+   server-side precisely so authority cannot be asserted from inside a sandbox.
+   A delegation path that drops the caller reopens that from the other end.
+2. **It loses accountability.** #1049's requirement 3 fails at the first hop
+   unless the accountable principal is carried, not recomputed.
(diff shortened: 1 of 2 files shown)


alexrao2000:
Accidentally merged and reverted

alexrao2000:
## Demo: what this ADR is proposing, running

A **working prototype** of this ADR's decision, recorded end to end. To be unambiguous up front: this ADR is still `Draft`, so per ADR-0085/0102 it authorizes no implementation. The prototype exists to make the proposal concrete and reviewable — it is **not** a candidate implementation and is not offered for merge as one.

ADR-0115 delegate-call prototype

The three beats, in order:

1. **Two agents deploy.** `alice` gets the auto-mounted `curie-delegate` MCP tool with no manifest change — it is platform-injected, per Decision part 1. `bob` is backend-only.
2. **The call is refused while unarmed** — `403`, *"alice is not armed to call bob; an operator must arm this pair first"*. Default closed, per Decision part 5.
3. **An operator arms the pair, and the same call succeeds.** `alice`'s turn ends immediately (*"Asking bob about that."*) — the call is asynchronous, per Decision part 3. `bob`'s turn runs, and the round trip lands back on `alice`'s conversation: `"what is 2 + 2?"` → `"4"`, `status: delivered`, resolved in **under a second**, with no third party anywhere in the path.
4. **A second call shows the answer is trustworthy.** `"who wrote Dune?"` comes back `idk` rather than a fabricated answer.

Everything in the recording is a real HTTP round trip through real code — the API router, the auto-mounted MCP server, the worker reply-sink adapter, and two new tables. Nothing is stubbed at the HTTP or database layer.

> **`bob` is not a model, and the `idk` beat is there to make that visible.** The recording runs with no model credential, so the target side is a deterministic responder: it solves a two-operand integer expression by regex and says `idk` for everything else. `4` is really computed from the delegated message and really carried back through the delivery path — but it is arithmetic, not reasoning, and the demo should not be read as evidence that an agent reasoned. The caller side is scripted too: a `[fake:delegate:bob]` marker stands in for the decision to call, so this also does **not** show that an agent will *choose* to delegate from its skill. Both halves stay inside the `CURIE_FAKE_MODEL` offline guarantee (ADR-0055 — the fake model is a plumbing fixture): no model call, no network. A real-credential recording remains the stronger artifact.

### What it reuses rather than invents

The point the ADR argues, made concrete: the prototype adds no new ingress mechanism. The scoped token is ADR-0033's pattern with a new `delegate` scope, verified the way `routers/state.py` verifies `state`/`state.app`. The MCP server is ADR-0073's auto-mounted pattern, modelled directly on `curie-state`. Turn minting and enqueue reuse `delivery.py`'s existing claim/XADD helpers — the same machinery `hooks.py` and `channels.py` use. The return path uses a new `ReplyHandle.kind`, which ADR-0096 already makes an open vocabulary, so no protocol change was needed there.

### Where it deliberately falls short of the ADR

Documented in full in [`docs/demo/ADR-0115-PROTOTYPE-NOTES.md`](https://github.com/curie-eng/curie/blob/task/adr-0115-reland/docs/demo/ADR-0115-PROTOTYPE-NOTES.md). The load-bearing ones for reviewing the *decision*:

- **Reuses `TurnSource.WEBHOOK`** instead of the message-lane source Decision part 2 argues for. A new source value is a breaking change to the frozen tri-language contract; the prototype skipped it, so it does not exercise the lane argument at all.
- **No formal suspend/resume.** The caller's turn simply ends and the reply arrives later as an ordinary turn on the same conversation. Asynchronous and non-blocking, but not the `Approval`-style suspend of Decision part 3.
- **No bundle declaration, and no chain/depth on the wire.** Authorization is a flat `delegate_grants` table armed by name; depth is capped at 1 by refusing any call whose own conversation is already a delegate call. That stands in for Decision part 6's chain/cycle bound rather than implementing it.
- **A delegate-callable agent can hold no other channel binding**, because `agent_channels` enforces one binding per agent. `bob` is backend-only for that reason — this is the constraint most likely to matter for a real design.
- **Neither side of the call is a model**, and the two halves fail differently: the caller's decision to delegate is scripted by a marker, and the target answers by regex arithmetic rather than reasoning. Detail in the notes doc.

### Two things for maintainers, independent of this ADR

- **Migration number collision.** `0028_delegation_calls` here and `0028_agent_channels_multi_binding` on `task/1525-multi-surface` both declare `down_revision = "0027"`. No per-branch gate catches this, so whichever lands second needs renumbering. Green CI here is not evidence the two are safe together.
- **A pre-existing CodeQL alert on `main`.** Getting this branch green required re-dismissing alert #4 (`py/weak-sensitive-data-hashing` on `sandbox_token.py`). It is the same finding already dismissed as a false positive in #2 — HMAC-SHA256 token signing, not password hashing — which reopened under a new number when the AgentOS→Curie rename moved the file, orphaning the dismissal. It was created 2026-07-24 and is unrelated to this branch, which never touches that file. Worth knowing that **any** future PR touching a `sandbox_token` call site will hit the same red check.

TheConnMan (changes requested):
Please remove the prototype files, then we're good to go.

TheConnMan:
## Maintainer decisions to incorporate

This is separate from the existing requested change to strip the prototype files; that request stands. This comment records the architectural decisions for the ADR.

1. **Authorization belongs to each bot.** A may call B only when A is authorized to call B. A user who can call A but cannot directly call B may still cause A to choose to call B when A is authorized to do so. That is an intended delegation path, not authority laundering.

2. **V1 identity may be payload-level agent identity.** Please explicitly reference discussion #1049 and retain the distinction between this v1 identity choice and a later IdP-backed identity model.

3. **Bot interactions use the core first-party surface.** When an authorized A @mentions or calls B, the interaction is mediated, authorized, attributed, and persisted through that surface.

4. **No private/internal bot-to-bot bypass.** A must not invoke B through an internal direct path. Authorized calls go through the core surface; A and B may then have a conversation there.

alexrao2000:
Addressed in [e86088df](https://github.com/curie-eng/curie/commit/e86088df) (force-pushed): dropped the two prototype commits (`60802636`, `a9fc2bfe`) that carried the demo implementation. The PR now only touches `docs/adr/0115-agents-call-each-other-with-no-third-party.md` and `docs/adr/README.md`, matching the description. `scripts/check-docs.sh` passes clean.

alexrao2000:
Incorporated in [e86088df](https://github.com/curie-eng/curie/commit/e86088df):

1. **Per-bot authorization.** Added to §5 (Default closed): the check is "may A call B", not "could this human do what B can do directly" — a user causing an authorized A to call B is the intended delegation path, not the laundering failure mode from Context.
2. **V1 identity.** Reframed the §4 gap paragraph as an explicit decision: v1 identity is payload-level pending #1049, not IdP-backed — same content as before, now stated as a decision rather than an apologetic gap.
3. **& 4. No bot-to-bot bypass.** Added to §1: any A↔B exchange, not just the initial call, stays entirely on the first-party surface — mediated, authorized, attributed, persisted — with no internal/direct path that skips it.

`scripts/check-docs.sh` still passes clean after the edits.

alexrao2000:
> Please remove the prototype files, then we're good to go.

I removed them. Why don't we want them in here? Wouldn't they be a good starting point for the actual build?

alexrao2000:
I preserved the prototype build in a different branch/PR

TheConnMan (approved):
Thanks for preserving on another branch. I don't want to merge prototype code into the core codebase. Keeping it on a branch for the real implementation to reference is perfect.

Issue #1801, opened 2026-08-21
Extend the CLI's preserved-values mechanism to the 8 new existingSecret/existingSecretKey field pairs from #1759


athusoo:
## What

#1759 adds a per-key `existingSecret` / `existingSecretKey` pair (mirroring `api.githubAppExistingSecret`) to the 8 direct-passthrough credential keys that previously had none: `agentSandbox.runner.credentials`, `worker.adapterCredentials`, `api.githubToken`, `sealing.privateKey`, `sealing.previousPrivateKey`, `dispatcher.slack.appToken`, `dispatcher.slack.botToken`, `dispatcher.slack.signingSecret`.

None of the 16 new fields (8 `*ExistingSecret` + 8 `*ExistingSecretKey`) are covered by the CLI's preserved-values mechanism in `cli/src/ops.rs`. A plain `curie cluster up` runs a full `helm upgrade --install` with no `--reuse-values` (`cli/src/ops.rs`, around the `up_commands_with_plan` helm invocation): anything not explicitly re-passed as `--set`/a values file resets to the chart default. Two prior issues are exactly this failure hitting a values key the CLI did not yet know to re-supply: #1067 (Slack tokens, fixed by `COMMS_MANAGED_KEYS`) and #1256 (the GitHub App identity, fixed by `GITHUB_APP_MANAGED_KEYS`, which also picked up `api.githubAppExistingSecret`/`api.githubAppExistingSecretKey`).

So today: an operator who sets, say, `dispatcher.slack.botTokenExistingSecret` by hand (there is no dedicated CLI flag for it) will have it silently reset to `""` by the next unrelated `curie cluster up` -- the BYO reference just stops being read and the render falls back to the chart's own Secret with no warning.

## Why deferred out of #1759

#1759 explicitly permits shipping the chart-level "shared mechanism" and handing CLI wiring off separately, the same way #1255 tracks CLI wiring for the one key (`githubAppPrivateKey`) that already had this escape before #1759. Pulling `cli/src/ops.rs`'s preserved-value resolvers (their own test/lint/clippy gates, and a change to `curie diff`'s reported output) into the chart-only ticket would have been disproportionate.

## Acceptance criteria

For each of the 8 keys, once its `*ExistingSecret`/`*ExistingSecretKey` values are set (by any means -- hand-edited values file, `--set`, or a future dedicated flag), a subsequent plain `curie cluster up` must not reset either field to empty:

[...]

Issue #1359, opened 2026-08-06
Sealing cannot be used by an author or rotated by an operator: no public-key publisher, no reseal, no rotate verb, no existingSecret path


TheConnMan:
ADR-0094 promises two operational properties that the implementation does not provide. The cryptography and the two key overlap are genuinely well built; these are the surfaces around them.

## 1. Nothing publishes the cluster public key, so an author cannot seal

ADR-0094: "The public half is published so authors can seal against it."

`cli/src/seal.rs`'s own docstring says the `--public-key` path "is not a convenience... Requiring kubectl would put every author in the operator's shoes and quietly undo the separation ADR-0094 is buying."

But `public_key_of` has exactly one production call site, and it is inside `curie seal`'s **cluster** branch, which requires kubectl read on the release Secret:

```
$ grep -rn "public_key_of" cli/src/ | grep -v "fn public_key_of"
cli/src/sealing.rs:226:   (test)
cli/src/seal.rs:152:      crate::sealing::public_key_of(current)     # inside the kubectl branch
cli/src/ops.rs:1498:      (test)
```

No verb emits the public key. `curie cluster --help` has no `sealing-public-key` or equivalent. So an agent author with a repo and no cluster access runs:

```
$ printf 'x' | curie seal --connector g TOK
Error: release curie in namespace curie has no sealing key, so nothing can be sealed to it. ...
```

and `--public-key` needs a value nobody can produce for them. Every seal therefore requires an operator, which is the separation the ADR set out to remove.

**Fix:** add `curie cluster sealing-public-key` (and surface it in `cluster status --json` and `curie doctor`), so an operator hands the value out once and authors seal offline.

## 2. No reseal and no rotation verb

ADR-0094 names rotation as "the strongest argument against the whole approach and the thing to get right in the implementation", and sets an explicit minimum:

> at minimum, support two active keys so a rotation can overlap, and make `curie seal` able to reseal a repository in one command.

The first half is delivered and well tested: `SEALING_PREVIOUS_PRIVATE_KEY` (`cli/src/sealing.rs:32`) is rendered at `charts/curie/templates/secrets.yaml:85`, preserved across `up` at `cli/src/ops.rs:1087`, and `open_with_any` tries both. Both languages implement the overlap.

[...]

Issue #1256, opened 2026-08-04
cluster up drops the GitHub App credential, silently reverting the platform to the PAT


TheConnMan:
A plain `curie cluster up` resets `api.githubAppId`, `api.githubAppPrivateKey` and `api.githubCloneBase` to chart defaults, silently disconnecting the GitHub App configured by `curie cluster github-app`.

`ops.rs:792`'s own comment says a plain `cluster up` "does a FULL upgrade and drops anything it does not re-pass." `fetch_existing_values` re-supplies only `REQUIRED_SECRETS`, `COMMS_MANAGED_KEYS` (the two Slack tokens) and `api.githubToken`. `grep -n "githubApp" cli/src/ops.rs` returns zero matches, and `--reuse-values` appears only in `comms.rs` and `github_app.rs`, never on the `cluster up` path.

## The chart's lookup guard does not cover these keys

This was the most likely refutation and it does not apply. The chart does have a `lookup`-based preserve pattern feeding `curie.managedSecret` at `charts/curie/templates/secrets.yaml:28`, but `secrets.yaml:73` renders the private key with no guard at all:

```
githubAppPrivateKey: {{ .Values.api.githubAppPrivateKey | default "" | quote }}
```

And `githubAppId` (`api.yaml:94`) and `githubCloneBase` (`api.yaml:116`) are plain Deployment env values, where no lookup guard is possible.

## Failure scenario

An operator runs `curie cluster github-app --app-id N --private-key app.pem`, confirms private-repo deploys work, then later runs a routine `curie cluster up` to change something unrelated. Two renders diff to:

```
githubAppPrivateKey: "FAKEKEY"   ->   githubAppPrivateKey: ""
GITHUB_APP_ID  value: "1234567"  ->   value: ""
```

`github_app.py:102` then reports not-configured and the platform falls back to the PAT, so every private clone 404s, with no error at upgrade time. This is the same defect class as #1067 (Slack tokens) and #1124 (`githubToken`), both fixed by adding to a preserve list.

One nuance: `githubCloneBase` reverts to the chart default `https://github.com` rather than to empty, so it breaks only GitHub Enterprise operators, silently and for a subset, which is harder to notice.

Render-only evidence is conclusive here precisely because no lookup path exists in the template, so cluster mode and template mode cannot differ. The live `helm upgrade` was not observed (no cluster available on the review box).

## Acceptance criteria

[...]

Issue #1255, opened 2026-08-04
The CLI cannot use the recommended BYO-Secret path, and misreports success when one is configured


TheConnMan:
The chart calls `githubAppExistingSecret` the recommended way to supply the App private key, but the CLI has no flag for it, and when one is configured the CLI reports success while the platform keeps using the old key.

## The recommended path has no CLI surface

`charts/curie/values.yaml:479` says "Prefer githubAppExistingSecret below for anything but a quick trial." `charts/curie/README.md:666` labels it "Recommended", and calls the CLI path "Quick trial" with "move to `githubAppExistingSecret` before you rely on it."

But `curie cluster github-app --help` offers only `--app-id`, `--private-key`, `--clone-base`, `--disconnect`, `--namespace`, `--release`, `--chart` and `--dry-run`. `grep -rn "ExistingSecret\|existing_secret" cli/src/` returns nothing. The CLI only ever emits `--set-file api.githubAppPrivateKey=<path>`, which is the path #1236's own commit message documents as leaking the PEM into retained helm release history, and which `github_app.rs:5-10` still argues is the secure choice.

This inverts the repo's one-entry-point rule: the CLI implements only the path the docs tell you not to use, and the recommended path is reached exclusively through hand-edited values plus `kubectl create secret`.

## Failure scenario: silent misreport during key rotation

With `api.githubAppExistingSecret` already in the release, rendering with the CLI's exact flags:

```
# no existing secret
key: githubAppPrivateKey   (name: curie-secrets)
# + api.githubAppExistingSecret=my-byo-secret
key: "privateKey"          (name: "my-byo-secret")
```

An operator rotating the key runs `curie cluster github-app --app-id N --private-key new.pem`, gets "GitHub App configured", `{"github_app_configured": true}` and a rollout, while the API keeps signing with the old key from the BYO Secret and the new PEM sits in helm release history for nothing.

## Two related defects in the same area

- `disconnect_commands` (`cli/src/github_app.rs:63-77`) clears `githubAppId` and `githubAppPrivateKey` but not `githubAppExistingSecret`, so `--disconnect` leaves a dangling reference.
[...]

PR #1802, opened 2026-08-21
Add existingSecret escapes to the remaining direct-passthrough credentials

### LINKED ISSUE #1759: Eight credential keys have no BYO Secret escape, so the model credential and the Slack tokens always land in helm release history
Found during a live v0.7.0 GA upgrade of a staging cluster on 2026-08-20.

## What

The chart-managed credential Secret (`charts/curie/templates/secrets.yaml`) carries 21 keys in `stringData`. They split into two very different classes:

**Class A, resolved through `curie.managedSecret` (11 keys):** `postgresPassword`, `valkeyPassword`, `clickhousePassword`, `rustfsSecretKey`, `langfuseSalt`, `langfuseEncryptionKey`, `langfuseNextauthSecret`, `langfuseInitProjectSecretKey`, `langfuseInitUserPassword`, `apiKey`, `githubWebhookSecret`. Left at their chart default these are generated once and then re-read from a live `lookup` of the existing Secret on upgrade, so an operator can omit them from values entirely and no value ever enters helm release history.

**Class B, direct `.Values` passthrough (9 keys):** `agentCredentials`, `adapterCredentials`, `githubToken`, `githubAppPrivateKey`, `sealingPrivateKey`, `sealingPreviousPrivateKey`, `slackAppToken`, `slackBotToken`, `slackSigningSecret`. Exactly one of these, `githubAppPrivateKey`, has a BYO escape (`api.githubAppExistingSecret` plus `api.githubAppExistingSecretKey`, consumed at `charts/curie/templates/api.yaml:102-104`). The other **eight have no escape hatch at all**, so the only way to set them is to put the plaintext value into helm values.

Consequence: the model credential and all three Slack secrets, among others, are copied into every retained helm revision and are printable with `helm get values`. The chart's own values comment for `githubAppPrivateKey` states the problem exactly ("a value set here is stored in the Helm release itself, so it is copied into every retained revision and `helm get values` can print it", `charts/curie/values.yaml:601-604`), and `charts/curie/templates/NOTES.txt:191-198` recommends the BYO Secret path "for the most sensitive value". That reasoning applies unchanged to the eight keys that cannot take it.

`charts/curie/templates/secrets.yaml` acknowledges the gap in its own header comment: the app services "read from this Secret unconditionally (they have no per-component existingSecret escape)".

## Why it matters

An operator who manages secrets out of band (External Secrets, SOPS, a sealed-secrets controller, a cloud secret manager) can do so for every backing store and for the GitHub App key, and cannot do so for the credential that authorizes the model provider or for the tokens that own the Slack workspace connection. On a shared or audited cluster that is the difference between a supported install shape and a hand-patched one.

## Suggested shape

Mirror the idiom the chart already uses everywhere else. Either a per-key `existingSecret` plus `existingSecretKey` pair on each of the eight, or a single `existingSecret` naming one operator-managed Secret that supplies any subset of the direct-passthrough keys, with the values entry winning only where the referenced Secret has no key. Whichever shape is chosen, the render must fail loudly rather than emit an empty credential when a referenced key is absent, following the precedent set for `otlpAuthHeader` in #1563.

## Related, and deliberately not duplicated

- #1132 asks for exactly this for `githubToken` alone, and is still open. This issue is the same request generalized to the remaining seven.
- #1359 item on sealing notes "no existingSecret path" for `sealingPrivateKey`, as one of several sealing gaps.
- #1255 covers the CLI side of the already-shipped `githubAppExistingSecret`, which is a different defect on the one key that does have the escape.

Closing this issue should either subsume #1132 and the sealing existingSecret item of #1359, or explicitly hand them the shared mechanism.

athusoo:
## Summary

Eight of the nine direct-passthrough credential keys in `charts/curie/templates/secrets.yaml` had no BYO `existingSecret` escape (only `githubAppPrivateKey` did, ADR-0092). This adds the same `existingSecret`/`existingSecretKey` shape to the other eight: the model credential (`agentSandbox.runner.credentials`), per-adapter egress secrets (`worker.adapterCredentials`), the outbound GitHub token (`api.githubToken`), the sealing keypair (`sealing.privateKey`/`sealing.previousPrivateKey`), and the three Slack credentials (`dispatcher.slack.appToken`/`botToken`/`signingSecret`). The two keys with more than one consumer template (`agentCredentials`, `slackBotToken`) get a shared `_helpers.tpl` helper so every consumer resolves identically; the other six get an inline if/else matching the existing `githubAppPrivateKey` shape. Also fixes two parity-seam gates (`curie.dispatcher.enabled`, the `agentCredentials` render gate) that would otherwise silently ignore a BYO-only credential.

**Deliberately out of scope, and disclosed rather than shipped silently:** none of the 16 new fields are yet covered by the CLI's preserved-values mechanism (`cli/src/ops.rs`), so a plain `curie cluster up` will still reset a hand-set field to empty on the next unrelated upgrade -- the same failure shape #1067 and #1256 hit before `COMMS_MANAGED_KEYS`/`GITHUB_APP_MANAGED_KEYS` existed. This is explicitly licensed by #1759's own text (the same split already exists for `githubAppExistingSecret` via #1255) and would otherwise pull a 10,800-line Rust file with its own test/lint gates into a chart-only ticket. Documented next to every new field, in the README, and in NOTES.txt. Follow-up: #1801.

## Related issue

Closes #1759

Related: #1132 (githubToken specifically, superseded by this), sealing item of #1359 (superseded by this), #1255 (the analogous CLI-side ticket for the one key that already had this escape), #1801 (the follow-up this hands off to)

## End-to-end verification

| Tier | Required / n/a | Reason | Mode | Command and observed outcome |
| skill | n/a | No plugin/skill packaging, runner turn loop, or ACI event touched | -- | -- |
| local | n/a | No compose service, dispatcher/worker/API source, or `curie` verb output change; only which Secret a chart `secretKeyRef` points at | -- | -- |
| local-release | n/a | No released binary/image identity or install-path change | -- | -- |
| cluster | required | Chart templates change (`_helpers.tpl`, `secrets.yaml`, `api.yaml`, `worker.yaml`, `dispatcher.yaml`, `agent-sandbox.yaml`) | live | Real `helm install`/`upgrade` against a disposable minikube namespace (`curie-1759ver`, since deleted). Positive: `kubectl exec` into the real dispatcher/api/worker pods after pointing `dispatcher.slack.{appToken,botToken,signingSecret}ExistingSecret` and `agentSandbox.runner.credentialsExistingSecret` at hand-created Secrets showed every consumer of both multi-consumer keys (`slackBotToken`: dispatcher/api/worker; `agentCredentials`: worker) resolving the exact same operator-supplied value. Negative: pointing `botTokenExistingSecret` at a Secret missing that key produced a new dispatcher pod in `CreateContainerConfigError` with kubelet event `Error: couldn't find key slackBotToken in Secret .../my-slack-secret-missing-bot`; a follow-up upgrade back to the good Secret returned the pod to Running. Full teardown confirmed (namespace deleted, no PVs/pods left, `helm list -A` empty). |
| live provider | n/a | Changes only where credential material is sourced from, not model routing/auth code; no real model call needed to prove secret provenance | -- | -- |
| external integration | n/a | Changes only chart secret wiring, not dispatcher/worker Slack client code; no real Slack workspace needed to prove render/gate behavior | -- | -- |

- [x] Each meaningful acceptance criterion has positive proof plus a falsifiable negative (see cluster row; also covered structurally by the new `charts/curie/ci/direct-passthrough-existing-secret-assertions.sh`, 30 render assertions including negative controls for both gate fixes).
- [x] No required tier is left unproved.
- [x] Behavior-bearing (chart templates change); cluster tier carries the proof above.

## Checklist

- [x] Tests pass: `helm lint` (default + dev + e2e-nogvisor overlays), `helm template` (both), the new `direct-passthrough-existing-secret-assertions.sh` (30/30), and every sibling `ci/*.sh` script touching the same templates (`github-app-credential-assertions.sh`, `sealing-key-assertions.sh`, `dispatcher-api-wiring-assertions.sh`, `otel-collector-langfuse-secret-assertions.sh`, `connector-secrets-assertions.sh`, `worker-object-store-assertions.sh`, `worker-eval-wiring-assertions.sh`, `sandbox-hardening-assertions.sh`).
- [x] Docs updated: `charts/curie/values.yaml` comments, `README.md` (`## Secrets at rest`), `NOTES.txt`.
- [ ] No ADR: this mirrors an existing chart idiom (`githubAppExistingSecret`, ADR-0092) rather than introducing a new architectural decision.


alexrao2000:
## Review

The chart-side work itself is solid — all 8 BYO escapes are correctly wired, `nindent` depths check out, both parity-seam gates (`curie.dispatcher.enabled`, the `agentCredentials` render gate) are fixed correctly, and the 30-assertion CI script genuinely covers the render behavior. A few gaps beyond what's disclosed, though:

### Correctness

1. **`cli/src/ops.rs:6181` `read_sealing_keys()`** and **`cli/src/ops.rs:6048` `discover_slack_bot_token()`** both read the plain value straight from the chart's managed Secret and never resolve the new `*ExistingSecret` fields. `curie seal` and `curie cluster message` will fail for anyone who adopts the BYO escape this PR ships, even though the workload itself resolves correctly in-cluster. This is a different, more acute bug than the disclosed "#1801 known gap" (which is only about `cluster up` resetting values on upgrade) — sealing keys are called out in this PR as "the most sensitive value most installs carry," and this breaks `curie seal` for exactly that BYO case.
2. **`charts/curie/templates/worker.yaml:116`** — the pre-existing `checksum/adapter-credentials` pod-restart annotation still only hashes `.Values.worker.adapterCredentials`, never the new `worker.adapterCredentialsExistingSecret`/`Key`. Rotating a BYO adapter-credentials Secret's contents in place (the normal BYO rotation flow) won't change the checksum, so the worker Deployment won't roll and workers keep the old credential in memory — silently narrowing a rotation guarantee that held before this PR for anyone who adopts the new escape.

### Scope/depth

3. Deferring CLI preserved-values coverage to #1801 is worth reconsidering given its own precedent: `githubAppExistingSecret`'s CLI-preservation companion (`GITHUB_APP_MANAGED_KEYS`) shipped one day after the escape itself (`25ad2d59` → `44186714`, "Stop cluster up from silently dropping the GitHub App"). For 6 of these 8 keys the plain values *were* already preserved by `cluster up`, so adopting the new escape is a regression in upgrade behavior, not just a "not yet" gap.

### Minor

4. `charts/curie/README.md:931` uses a literal em-dash ("—") instead of the repo's " -- " convention (AGENTS.md: "No dashes/emdashes in prose content") — the only occurrence in this PR's otherwise-consistent new prose.
5. `charts/curie/templates/_helpers.tpl:764` — only the 2 multi-consumer keys (`agentCredentials`, `slackBotToken`) got a shared `curie.secretRef.*` helper; the other 6 single-consumer keys got hand-copied inline if/else blocks instead of one generic helper parameterized by field name (matching the dict-argument pattern `curie.image`/`curie.managedSecret` already use in this file). Leaves the same drift risk the PR's own commentary warns about, just narrowed to keys that might gain a second consumer later.
6. `charts/curie/ci/direct-passthrough-existing-secret-assertions.sh:382` — assertions `3a`/`3b` duplicate `2b`/`2h` byte-for-byte; the `byo` render already proves "wins over stale value" as part of group 2, so these two add no independent coverage while inflating the advertised assertion count.

The `curie seal`/`curie cluster message` breakage (#1, #2) is the one I'd fix or explicitly disclose before merge, since it hits the exact use cases (sealing keys, connected Slack) this PR is built for.

TheConnMan (approved):

athusoo:
Thanks for the thorough review — addressed in 99c930bc.

**Correctness (1, 2): fixed, both verified live.**

1. `discover_slack_bot_token`/`read_sealing_keys` now resolve through a new `resolve_existing_secret_ref` (pure, unit tested) + `read_direct_passthrough_secret` pair that checks the release's BYO `*ExistingSecret` field before falling back to the chart's own Secret. Verified against a real cluster: `curie seal` against a release with `sealing.privateKeyExistingSecret` set derives the exact public key independently computed from that BYO private key, and — with the override cleared and the chart's own `sealing.privateKey` left at a deliberately garbage value — fails loudly (`the sealing private key is not valid base64`) instead of silently reading the wrong thing. `discover_slack_bot_token` shares the identical resolver (its own unit test uses the slack-token value path as its exact case); didn't re-run it through the much heavier `cluster message` live path (needs a real dispatcher + Valkey port-forward + channel resolution) since it exercises no code the sealing-key live run didn't already cover.
2. `checksum/adapter-credentials` now hashes the existingSecret name/key alongside the inline value — added assertion 7 covering it (switching `adapterCredentialsExistingSecret` now changes the checksum with the inline value held constant). Rotating a BYO secret's *content* in place still needs a manual rollout, same as every other existingSecret-sourced credential in this chart — noted in the comment.

**Scope/depth (3): investigated, not done here, sequencing recorded on #1801.**

You're right that this deserves reconsidering rather than a flat "later" — updated #1801 with your framing verbatim plus what I found investigating it: the scope is uneven across the 8 keys. `dispatcher.slack.{appToken,botToken}` is close to append-only (`COMMS_MANAGED_KEYS` already iterates generically). Sealing needs a new disposition branch so BYO doesn't trigger chart-side key generation — not just a list append. `api.githubToken`/`agentSandbox.runner.credentials` have bespoke resolution threaded through `UpOpts`, `--set` generation, and `curie diff` at several call sites each — the largest piece. `worker.adapterCredentials` has no existing preservation family at all. Given that spread, doing it properly (with its own list-parity test per key family) felt like it deserved its own PR rather than a rushed pass appended to this one — sequencing plan is on the issue.

**Minor (4, 5, 6): all fixed.**

4. Em-dash fixed.
5. Replaced both named `curie.secretRef.*` helpers and all six inline if/else blocks with one generic, dict-parameterized `curie.secretRef` (matching `curie.image`/`curie.managedSecret`'s own style) — every one of the eight keys now shares the same resolution logic, not just the two that already had a second consumer.
6. Removed 3a/3b (confirmed byte-identical to 2b/2h) and fixed the advertised count.

Full chart + CLI suites green: `helm lint` (×3 overlays), `helm template` (×2), the direct-passthrough script (29/29), every sibling `ci/*.sh` script, and `cargo fmt`/`clippy -D warnings`/`test --lib` (827/827, including 3 new unit tests for the resolver).

PR #1805, opened 2026-08-22
ADR-0117: a tool that changes the world reports what it changed, and the platform can put it back

jw3329:
## What this is

**ADR-0117, Accepted, and nothing else.** Per the review: prototype stripped, ADR
ready to merge, implementation to follow as real work.

The diff is three files — the ADR, the regenerated index, and the design pass
behind it. `bash scripts/check-docs.sh` passes.

## What came off

Everything the prototype added: the ACI `arguments`/`result` fields and the
regenerated Rust/TS bindings, the per-call frames in the runner, the `k8s-scale`
connector, the `agent_actions` migration and the actions API, the receipt card,
the probes, the demo and its recording. Thirty-two files, ~4.9k lines.

That is the process working rather than a loss. Under
as amended by
acceptance is what authorizes implementation, and none of that code had it. It
survives on **`spike/adr-0117-action-ledger`** as a reference for the slices, not
as code to land.

## What the ADR kept, and what it gave up

The findings stay; the claims go. The old "Evidence (built and measured)" section
is now **"What a prototype established"** — same findings, each stated because it
narrows a slice, and not one citation to a path this PR removes:

- a connector can report prior state on every path, and a failed read must refuse
  to write at all
- the scale tool belongs beside `k8s-write`, and that is also the narrower RBAC
  grant
- `arguments` and `result` on `SideEffectFlag` are a **patch** under ADR-0036, not
  a minor
- per-call framing leaves `saw_side_effect` latching on the first frame, so
  ADR-0013's no-retry rule reads exactly what it read before
- the ledger migration has to be hand-written, because autogenerate proposes
  dropping Langfuse's tables
- the round trip closes, and the refused undo is the step the feature lives on

Claims about code that is no longer changed go back to present tense: the frame
*is* built with a constant `detail` string, the v0.1 contract *drops* the tool
result, the flag *fires* once per turn. Status is `Accepted` with `Issue: #1861`.

## The implementation

Tracked in #1861, one PR per slice against `next`:

| | |
| --- | --- |
| #1862 | the ACI carries a call's arguments and result, one frame per call |
| #1863 | the ledger tables, and the kernel writes a record per call |
| #1864 | the actions API: record, list, and rule on an undo |
| #1865 | a write connector reports the state it read before writing |
| #1866 | the receipt: a per-action card with an undo control or a stated reason |
| #1867 | where the undo executor lives — **needs its own ADR first** |

#1867 is the decision this ADR deliberately does not make: nothing in the platform
can reach a connector today, so the API can rule on an undo but cannot perform
one.

Also filed: #1868, the latent migration-head pin in
`test_migration_0021a_v0_6_x_reconcile.py`. Its fix came off with the migration
that provoked it, so the defect is still live on `next` and will bite the next
migration author.

### DIFF
--- a/docs/adr/0117-a-tool-that-changes-the-world-reports-what-it-changed.md
+++ b/docs/adr/0117-a-tool-that-changes-the-world-reports-what-it-changed.md
@@ -1,0 +1,351 @@
+# 117. A tool that changes the world reports what it changed, and the platform can put it back
+
+Date: 2026-08-21
+
+Status: Accepted
+
+Issue: #1861
+
+Completes the pair [ADR-0010](0010-approval-gates-and-human-in-the-loop.md)
+opened. That decision answers a mutating action **before** it happens; this one
+answers it **after**. It changes no isolation boundary:
+[ADR-0006](0006-security-rails-as-chart-defaults.md)'s rails and
+[ADR-0008](0008-multi-tenancy.md)'s tenant compute stand as they are, and a
+sandbox reaches nothing new.
+
+## Context
+
+Curie already detects that a turn changed the world, and records one bit.
+
+[`runner/src/curie_runner/side_effects.py`](../../runner/src/curie_runner/side_effects.py)
+classifies every tool absent from a harness-declared read-only allowlist as
+side-effecting, deny-by-default, so an unknown tool is assumed to mutate. The
+runner emits a `side_effect_flag`, and the kernel reduces the whole stream to
+`saw_side_effect: bool`
+([`apps/worker/src/curie_worker/kernel.py`](../../apps/worker/src/curie_worker/kernel.py))
+for exactly one purpose: refusing to auto-retry a failed turn that already
+touched something (ADR-0013).
+
+That is the entire memory the platform keeps of an agent acting on production.
+Not which tool. Not with what arguments. Not against what resource. Not whether
+it worked. So when a bot changes something, the person who asked gets a sentence
+of prose from the model and no account from the platform, and there is nothing to
+act on afterwards.
+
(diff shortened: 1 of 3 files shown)


TheConnMan (changes requested):
I like this! An undo button would be very helpful. Thanks for the prototype. ADR approved. Please strip the prototype code out, merge the ADR, and make a real implementation.

jw3329:
Done — prototype out, ADR at `Accepted` with `Issue: #1861`. The PR is three files now: the ADR, the regenerated index, and the design pass behind it.

Two calls worth flagging:

**The prototype is preserved, not deleted.** It is on `spike/adr-0117-action-ledger` so the slices can crib from working code without any of it landing on the strength of a Draft.

**The ADR kept the prototype's findings and dropped its claims.** "Evidence (built and measured)" would have cited paths this PR removes, so it became "What a prototype established" — the patch-not-minor reasoning under ADR-0036, the hand-written migration, the RBAC argument for a separate scale connector, the refused-undo step. Claims about code that is no longer changed went back to present tense. Everything else in the body is untouched, and it is immutable from here under ADR-0045.

Implementation is #1861: six slices, one PR each, dependency-ordered. #1867 is the executor decision this ADR does not make and needs its own ADR before any code — until then an undo is authorized and recorded but not executed. Starting on #1862.

One thing to know regardless of this PR: #1868. `test_migration_0021a_v0_6_x_reconcile.py` pins the Alembic head as the literal `"0027"`, so the next migration anyone adds fails a test about the 0021 collision. The prototype's fix came off with the migration that provoked it, so it is still live on `next`.

TheConnMan (approved):

PR #1806, opened 2026-08-22
Draft ADR-0118: a version's golden checkpoint makes a refill a restore, not a boot

jw3329:
## Stacked on #1804

Base is `task/adr-0116-sandbox-residency`, not `next`, so this diff shows only
ADR-0118. **Read #1804 first** -- this ADR exists because of a number that one
measured.

Docs only. Per
as amended by
a merged Draft authorizes no implementation.

## Why this exists

ADR-0116 took a 17.39s boot off the 90-second claim deadline and measured the
bind at 0.18s, four times across three clusters. Then its own load test found
the ceiling:

| regime | claim p50 | claim p95 | throughput vs cold path |
| --- | --- | --- | --- |
| burst of 24, pool of 3 | 6.22s | **11.40s** | **1.0x** |
| burst of 4, pool of 4 | **0.21s** | 0.22s | 4 in 0.4s |
| arrivals 9s apart, pool of 3 | **0.14s** | 0.18s | all sub-second |

Only the first three conversations were served sub-second, one per warm pod.
**A refill is the same cold create the claim used to do inline**, so pre-binding
prepays a boot rather than removing one. Whether the pool keeps up is decided by
the cost of making one more ready runner, and today that cost is a full boot.

## What this decides

A version's deploy boots its runner **once**, quiesces it at a turn boundary with
no session established, and stores the checkpoint beside the bundle under the
same content digest. A warm pool then refills by **restoring** that image instead
of booting.

## Scoped against the measured phase table, not asserted

| phase | measured | can a restore skip it? |
| --- | --- | --- |
| claim admitted, pod created | 0.75s | **no** |
| pod scheduling, container starts | 1.93s | **no** |
| `bundle-fetch` | 4.11s | yes |
| `bundle-extract` | 0.42s | yes |
| container transition | 1.25s | partly |
| runner boot to `/healthz` | 8.93s | **yes, the largest** |

A restore skips **13.46 of 17.39 seconds** and cannot skip the pod lifecycle,
leaving a floor near **3.9s**. So:

| | | |
| --- | --- | --- |
| bind a running pool pod | **0.18s** | measured |
| restore a fresh runner | **~3.9s** | **projected, not measured** |
| boot a fresh runner | **17.39s** | measured |

**This does not make a claim faster and does not replace the pool.** A warm bind
is 0.18s either way. It makes refill ~4x cheaper, which raises the arrival rate a
given pool depth absorbs by about the same factor. The 3.9s is a projection and
the ADR labels it as one; its error bar is the `container transition` row and
whatever `runsc` restore itself costs.

## Two properties make this reachable rather than exotic

**gVisor is already the default sandbox runtime.** Verified live: `RuntimeClass
gvisor`, handler `runsc`, the chart's own gVisor preflight passing. `runsc` has
first-class checkpoint/restore, so the isolation cost the product already pays is
what makes this cheap to reach. It is also where the cost is: measured, gVisor
took the cold create from 4.72s to 7.06s and left the bind at 0.18s.

**The kernel already has turn boundaries.** ADR-0013's finish race, steer path,
and `/v1/reset` all key off a point with no in-flight request. Checkpointing a
socket-holding process is the classic hard case and quiescing first is the way
out; this system has the quiesce point already.

## The image holds no conversation and no credential

It is taken before any session exists, so no transcript, no history ref, no
session id. Credential and runner token are excluded from the checkpointed
address space and injected on restore. **An image that cannot be shown to exclude
them is not written**, and the ADR calls the exclusion test the
security-critical part of the decision.

That is what keeps this separable from the harder half of the idea --
checkpointing a **live** conversation to hibernate an idle thread, which inverts
ADR-0003 and puts a transcript plus a live credential at rest. #1804's deferred
item is now split along exactly that line: the refill half points here, the
hibernation half stays unowned.

## Restore has a boot as its fallback

`runsc` checkpoint/restore is the least proven component here. On any failure a
pool pod boots normally and records why, so the downside is bounded by today's
behaviour rather than by an outage.

## Reviewing this

The load-test harness behind the table is in
`prototypes/adr-0116-residency/load_*.py` on the base branch, with its results
committed. Its headline is negative and that is why it is kept.

### DIFF
--- a/docs/adr/0116-session-identity-arrives-over-the-aci-so-a-sandbox-can-be-pre-bound.md
+++ b/docs/adr/0116-session-identity-arrives-over-the-aci-so-a-sandbox-can-be-pre-bound.md
@@ -632,12 +632,20 @@ counts it, which decision 6 is what fixes.
   as hygiene; it does not remove a wall-clock deadline from a proportionally
   starved path.
 
-- **Process checkpoint/restore (`runsc checkpoint`) to preserve a live session
-  and its prompt cache across idleness.** Deferred, not rejected. It is the
-  natural extension of decision 5 and gVisor is already the default sandbox
-  runtime, but restoring a process holding an open ACI socket, and a checkpoint
-  image containing a live runner token, are each their own decision. This ADR
-  deliberately buys most of the benefit without them.
+- **Process checkpoint/restore (`runsc checkpoint`).** Deferred, not rejected,
+  and now split in two by the load test above.
+
+  The half that addresses this ADR's measured ceiling -- refill cost, which is
+  what caps the arrival rate a pool can absorb -- is taken up by
+  [ADR-0118](0118-a-versions-golden-checkpoint-makes-a-refill-a-restore-not-a-boot.md):
+  a version's deploy produces a golden checkpoint of a booted, *unbound* runner,
+  and a pool refills by restoring it. That needs no conversation and no
+  credential in the image, which is what makes it separable from the harder half.
+
+  The harder half is checkpointing a **live** conversation to hibernate an idle
+  thread and keep its prompt cache. That inverts
+  [ADR-0003](0003-stateless-first-rehydrate-on-resume.md), and it puts a
+  transcript and a live credential into an image at rest. It stays deferred.
 
 - **Replacing the adopted harness to shrink the 259.5 MiB per session.**
   Rejected under ADR-0007: `claude-agent-sdk` and the Claude Code plugin format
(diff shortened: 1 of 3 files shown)


TheConnMan (commented):
My understanding is this spike proved we _shouldn't_ do this, correct?

jw3329:
-

jw3329:
Closing per the above: blocked on an upstream restore path, not on complexity. Finding preserved in ADR-0116's alternatives (#1804).

Issue #473, opened 2026-07-16
Epic: model routing — opt-in binding hook + Pareto selection (ADR-0037)


TheConnMan:
Curie has no model routing today: model selection is a boot-time pin (CURIE_MODEL / per-agent override), one model credential resolves, and provider swap is a config-only seam. A customer engagement needs a router (phase 1: static task-type rules; phase 2: model-estimated difficulty), and the same mechanism enables cost optimization for any deployment.

ADR-0037 (PR to follow) decides the shape: an opt-in binding hook in the worker that routes at task-claim time (never an in-path proxy — see #253 and the prompt-cache rationale), a manifest-declared routing policy, an install-level model registry, deterministic cheapest-above-floor selection, and decisions that ride the typed ACI SessionConfig.

Stones (each an independent, fully-tested PR):
- [ ] #475: Promote model selection into typed SessionConfig (ACI)
- [ ] #476: Install-level model registry with N model credentials and egress
- [ ] #477: Manifest routing block + rules adapter (phase 1)
- [ ] #478: LLM judge floor estimator (phase 2)

Opt-out cost is zero: a bundle without a routing block behaves byte-identically to today.

Issue #24, opened 2026-07-07
Epic: bring your own model (OpenRouter and native Anthropic-format endpoints)


TheConnMan:
Run agents on non-Anthropic models, with evals as the way to learn which use cases tolerate which models. **OpenRouter is the primary path**: one key, hundreds of models, and an Anthropic-compatible Messages surface, so the existing harness works unchanged. The harness officially supports `ANTHROPIC_BASE_URL` gateways, and the runner's key mapping already fails loudly on foreign prefixes (`runner/src/curie_runner/sdk_auth.py`), so all of this is additive.

Three provider paths, in order:

1. **OpenRouter (primary).** Point `ANTHROPIC_BASE_URL` at OpenRouter's Anthropic-compatible endpoint; detect `sk-or-` key prefixes in the runner boot env and configure automatically. Known gotcha to encode: `ANTHROPIC_API_KEY` must be set to an empty string (not merely unset) or the harness can fall back to authenticating against Anthropic directly.
2. **Provider-native Anthropic-format endpoints.** Zhipu (GLM), Moonshot (Kimi), and DeepSeek each publish their own `/anthropic` endpoint, so a single-model agent needs no gateway at all: one format boundary, and the provider's automatic prefix caching applies.
3. **LiteLLM Anthropic-format sidecar** (values-gated in the chart) only for genuine multi-model routing/fallback needs.

Platform surface:

- [ ] `sk-or-` prefix detection and `ANTHROPIC_BASE_URL` plumbing in the runner boot env
- [ ] Native `/anthropic` endpoint support per provider (config, docs, key-prefix mapping)
- [ ] Values-gated LiteLLM sidecar in the chart
- [ ] Per-agent model selection surfaced in manifest/UI (`CURIE_MODEL` already exists in the boot-env path)
- [ ] Model dimension on the eval matrix: run the same suite across models, compare pass rate and cost per use case
- [ ] Prompt-cache smoke test: assert warm-thread cache hits on whichever path is configured

[...]

PR #1846, opened 2026-08-23
Draft ADR-0128: installation-owned model gateway and operator-selected aliases

TheConnMan:
## Summary

Adds Draft ADR-0128 on the `next` release train. It defines an optional installation-owned model gateway without replacing ADR-0037's direct-provider default or portable capability-intent boundary.

Review updates make the authority split explicit:

- Portable bundles declare capability intent; only operator-owned agent/install bindings select installation-specific aliases.
- The root gateway credential stays in the control plane. Sandboxes receive short-lived tokens scoped to installation, agent, session, alias, budget, and expiry.
- Gateway transport is authenticated and encrypted; plain HTTP bearer-token delivery is unsupported.
- A gateway may fall back only before the first concrete route is pinned. After pinning, unavailability fails the turn rather than switching models.
- Direct providers remain byte-identical when no gateway is configured.

## Verification

- `bash scripts/check-docs.sh` passed at `f8fe73f2`.
- Runtime tiers are not applicable because no runtime, API, schema, image, chart, or deployment behavior changes.

Ref #24
Ref #473

### DIFF
--- a/docs/adr/0112-install-owned-model-gateway-and-agent-aliases.md
+++ b/docs/adr/0112-install-owned-model-gateway-and-agent-aliases.md
@@ -47,19 +47,29 @@ protocol is Anthropic Messages because the current Claude harness speaks that
 protocol. A later OpenAI or Responses harness may add another declared protocol
 without changing this ownership boundary.
 
+Alias names are installation configuration, not a frozen Curie-wide vocabulary.
+An installation may use arbitrary aliases that its compatible gateway supports,
+including LiteLLM model aliases. Curie recommends the conventional names
+`fast`, `code`, `reasoning`, and `auto` as useful defaults, but does not require
+or reserve them. An installation may map those conventions, or any other
+allowed alias, to its own model pools and routing policy.
+
 Curie delivers only the gateway credential and gateway base URL to a gateway
 backed sandbox. Provider credentials remain on the gateway side. A direct
 provider credential and a gateway credential are distinct installation secrets;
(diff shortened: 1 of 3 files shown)


jw3329:
Heads-up on the ADR number before this one is picked back up: **0112 is taken on `next`.**

- This PR adds `docs/adr/0112-install-owned-model-gateway-and-agent-aliases.md`.
- `origin/next` already has `docs/adr/0112-a-turn-source-names-what-caused-the-turn.md`.

`main` has no 0112 yet, so nothing conflicts on this base today — the duplicate only appears when the v0.7.0 train merges down, at which point both records are Accepted and immutable. Cheaper to renumber here.

Numbers in flight across both trains: `main` is allocated through 0123, `next` through 0124, and 0125 is claimed by open #1934. I've suggested 0126 for #1847, so **0127** is clear for this one.

Same class of issue as #1545, which is holding an 0105 that has already landed on `main`.

TheConnMan:
Rebased onto `next`, renumbered to ADR-0128, regenerated the index, and refreshed the PR metadata.

The reread against ADR-0037 also tightened the decision: portable bundles declare capability intent and never installation aliases; operator-owned agent/install bindings select aliases; the root gateway credential remains outside sandboxes; session tokens are scoped and short-lived; transport is authenticated and encrypted; and fallback is allowed only before the first concrete route is pinned. Direct providers remain the default when no gateway is configured.

`bash scripts/check-docs.sh` passes at `f8fe73f2`. The cross-train numbering process gap is tracked in #1957.

Issue #1609, opened 2026-08-17
Add an ACP client path behind the harness port


TheConnMan:
## What

Add an inbound Agent Client Protocol (ACP) client behind the existing harness port, so one adapter admits every ACP-compatible harness instead of a per-vendor ACI server.

## Why now

ADR-0040 already establishes the ACP relationship and already concedes that fifty-plus agents including Claude Code, Codex, and OpenCode speak it. But it aims Curie *outward* as an ACP server for editor embedding. Meanwhile `docs/architecture-vision.md` concedes the harness port is entangled with the plugin format, which is what makes a per-vendor ACI server expensive.

Three separate projects made ACP their interop layer inside two weeks: Vercel eve shipped `@ai-sdk/harness-acp` with a `createACP()` meta-adapter where one integration admits any ACP harness ([changelog](https://vercel.com/changelog/use-acp-compatible-harnesses-with-the-ai-sdk-harness-layer)), and Block Buzz, xAI Grok Build, and Agentrove all treat ACP as the interop standard. An inbound ACP client is the cheapest credible route to a second and third harness.

This does not disturb ADR-0040's correct decision that `TurnEvent` stays the internal canonical type.

## Prerequisite

This wants an amendment to ADR-0040 before any code, because ADR-0040 currently decides the outbound direction. Per the ADR gate in CLAUDE.md, the amendment is a new ADR that supersedes rather than an edit, and it needs explicit maintainer acceptance before implementation.

## License

CODE-BORROWABLE. Vercel eve is Apache-2.0 and ACP itself is Apache-2.0. The borrow here is the direction of the adapter, not the TypeScript.

## Size

Large. Filed for sequencing, not dispatched for implementation.

Source: 2026-08-14 competitive scan, proposal A.

Issue #1515, opened 2026-08-14
Ship email as a first-party channel integration


TheConnMan:
ADR-0096 phase 2 (#1459, PR #1514) proved the neutral channel interface with an AgentMail email adapter running as an out-of-tree reference service: an external email reaches a bound agent and the reply threads back onto the mail conversation.

Promote that adapter to a first-party integration so email is a supported surface, not a test rig:

- A tracked home in the repo with its tests in CI
- Image built through the standard build path
- Chart wiring for deploy, credentials, and the egress secret pair (replacing hand-rolled manifests)
- Config surface for inbox, polling, and allow-listing
- Operator docs

Reference: docs/guides/building-a-channel-adapter.md describes the contract the adapter already satisfies. ADR-0096 is the governing decision.

TheConnMan:
Shipped in the v0.8.0 candidate via PR #1597. The implementing merge commit is an ancestor of origin/next; closing during the release milestone true-up.

PR #1847, opened 2026-08-23
Draft ADR-0127: an inbound ACP client admits ACP harnesses

TheConnMan:
## Summary

Adds Draft ADR-0127 on the `next` release train. The ADR adds the inbound ACP direction behind the existing harness port while preserving ACP as an edge projection rather than Curie's canonical turn type.

The review-driven safety boundary is explicit:

- Stable ACP v1 display fields are not grant authority. A generic peer is read-only unless a Curie-controlled permission identity contribution supplies the canonical tool and input.
- The approval card, durable provenance, canonical input digest, resumed grant, and execution check all bind to the same trusted operation. The separately reviewed frozen-contract prerequisite is #1956.
- Cancellation returns ACP's required `Cancelled` outcome before the prompt and process are released.
- The resumed grant has both a one-prompt boundary and an absolute expiry.
- Durable replay remains blocked on ADR-0119 implementation issue #1902.

## Verification

- `bash scripts/check-docs.sh` passed at `994c1c2a`.
- Runtime tiers are not applicable because no runtime, API, schema, image, chart, or deployment behavior changes.

Ref #1609
Ref #1902
Ref #1956

### DIFF
--- a/docs/adr/0120-an-inbound-acp-client-admits-any-acp-harness.md
+++ b/docs/adr/0120-an-inbound-acp-client-admits-any-acp-harness.md
@@ -1,0 +1,188 @@
+# 120. An inbound ACP client admits any ACP harness behind the harness port
+
+Date: 2026-08-23
+
+Status: Draft
+
+**Amends [ADR-0040](0040-adopt-acp-as-an-edge-projection.md)** by adding one
+decision, the inbound direction. ADR-0040's decision 1 (ACP is an edge
+projection, never the internal canonical type) is preserved verbatim and is
+extended to bind in both directions. Its decisions 2, 3, 4, 5 and 6 stand
+unchanged; this ADR adds a client path beside the server path of decision 2, it
+does not replace it. On acceptance, the back link ADR-0045 requires is added to
+ADR-0040's header; a Draft amends nothing yet, so no edit to ADR-0040 is made
+here.
+
+Tracked by [#1609](https://github.com/curie-eng/curie/issues/1609). Per
+[ADR-0085](0085-acceptance-not-implementation-authorizes-an-adr.md) as amended
+by [ADR-0102](0102-accepted-alongside-implementation-with-explicit-approval.md),
+this Draft authorizes no implementation. No ACP client, harness package, or
+runner entry point is written until it is Accepted with explicit maintainer
+approval.
+
+## Context
+
+ADR-0040 adopted ACP and aimed it **outward**: `curie runner acp` speaks ACP over
+stdio so Zed, the JetBrains IDEs, Neovim and Emacs can embed Curie. That is the
+editor-embedding problem, and it is the only direction ADR-0040 decided.
+
+The reverse direction is a different and currently more expensive problem. Adding
+a second harness today means writing an ACI server per vendor.
+`docs/architecture-vision.md` states the cost honestly: a candidate harness must
+implement an HTTP process serving the ACI wire, and "the harness port is
+entangled with the plugin format", because `packages/plugin-format` is the Claude
+Code plugin shape verbatim. ADR-0060 withdrew OpenCode-as-second-harness after
+its steer spike *succeeded*, precisely because of that unpriced per-vendor tail:
+the synthesis tax of forging another harness's dataclasses, plus the installer
+and the bundle compiler. One harness has ever crossed this port.
+
+Meanwhile the same fifty-plus ACP agent registry ADR-0040 already cites —
+Claude Code and Codex among them, with OpenCode shipping native ACP — is a set
+of engines that already speak a normalized turn and permission wire. Curie
+reimplements per vendor what those engines already export once.
+
+**Prior art, and the precise shape of the borrow.** Vercel's AI SDK harness layer
+ships a `createACP()` meta-adapter: one integration written against ACP admits
+any ACP-compatible harness, rather than one integration per vendor
+([changelog](https://vercel.com/changelog/use-acp-compatible-harnesses-with-the-ai-sdk-harness-layer)).
+Both that work and ACP itself are Apache-2.0. What is borrowed is the **direction
+of the adapter** — inbound, one adapter parameterized by which ACP agent it
+launches — and nothing else. No TypeScript is copied, and the surrounding AI SDK
+model is not adopted. Block's Buzz, xAI's Grok Build, and Agentrove independently
+converged on ACP as the interop layer in the same window, which is the signal
+that this is an ecosystem standard rather than one vendor's convenience.
+
+Two constraints from the existing corpus bound the answer. ADR-0060 makes a
+harness a declared package registering a `HarnessContribution`
+(`runner/src/curie_runner/harness/contribution.py::HarnessContribution`) through
+a guarded entry point. ADR-0061 is a **Draft**, gated on a spike, proposing that
+the harness boundary become an out-of-process `create_app() -> FastAPI` adapter;
+until that spike answers, the port in the tree is the in-process `ModelSession`
+Protocol (`runner/src/curie_runner/adapter.py::ModelSession`), whose values are
+mapped by `translate_message`
+(`runner/src/curie_runner/translate.py::translate_message`) into the ACI outbound
+union.
+
+## Decision
+
+**1. Add one inbound ACP client adapter behind the harness port.** Curie may act
+as an ACP **client**, driving an ACP-speaking agent as its engine, in addition to
+acting as an ACP server for editors. The adapter launches the configured ACP
+agent, performs `initialize`, opens a session, sends prompts, and consumes
+`SessionUpdate` notifications, mapping them into whatever internal turn
+representation the harness port carries.
+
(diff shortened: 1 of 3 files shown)


alexrao2000:
## Review

The decision itself is sound and well argued — one adapter parameterized by which ACP agent it launches, rather than a harness package per vendor, is the right shape, and it's aimed squarely at the cost ADR-0060 actually recorded. I checked every code path and symbol the ADR cites and they all resolve (`HarnessContribution`, `ModelSession`, `translate_message`, `format_conversation_preamble`, `run_conformance`, `packages/plugin-format`), and ADR-0040's decisions 1–6 are represented accurately, including the parts this ADR claims to leave standing.

The blockers below are all about **which train this is on**, plus two design questions in the rewritten decision 4.

### Blocking

**1. ADR number 0120 is already taken, by an Accepted ADR.**

`origin/next` has `docs/adr/0120-durable-first-party-email-state.md` — *"First-party email delivery state is local durable single-writer state"*, **Status: Accepted**, dated 2026-08-25, published alongside its implementation for #1515. A Draft can't take a number an Accepted-and-implemented decision already holds.

Numbering has genuinely diverged between the two trains, so this needs checking against both:

- `main` has up to 0115, then 0123 (no 0116–0122, no 0124)
- `next` has 0115–0122 and 0124 (no 0123), and its 0115 is a *different decision* from main's 0115
- open PRs hold 0112 (#1846) and 0125 (#1934)

First genuinely free number across all of that is **0126**.

To be fair to the author: the process in `docs/adr/README.md` was followed. It says read the number off the index and "check open PRs for a number already reserved off-main" — and 0116–0122/0124 were claimed by PRs that had already **merged to `next`**, which is neither the index on main nor an open PR. `scripts/check-docs.sh` couldn't catch it either; the uniqueness gate globs the local working tree (`tools/doclint/src/curie_doclint/generate.py:112`), so it can't see the other train by construction. That's a real gap in the rule, not just this PR's mistake — worth a separate issue (see note at the end).

**2. The ADR-0119 prerequisite doesn't exist on this PR's base.**

The second commit made ADR-0119 a hard blocking dependency — decision 5 now rehydrates from "ADR-0119's full, prefix-preserving structured message replay," and the header says durable ACP suspend/replay "cannot be implemented until #1902 lands." But ADR-0119 lives only on `next`. On `main`, the base here, there is no ADR-0119 for a reader to follow.

It's also worth noting what that dependency chain looks like: ADR-0119 is itself a **Draft**, and it declares a dependency on ADR-0116 (Accepted, also `next`-only). So a Draft on `main` is now blocked on a Draft on `next` that is blocked on an Accepted ADR on `next`.

Related: ADR-0119 is referenced eight times in the body and **never as a markdown link**, unlike every other ADR the file cites (0040, 0085, 0102 are all linked). That's why the citation check passed — the plain-text form isn't a citation as far as the lint is concerned. So the "every citation resolves" line in the PR description is true but not load-bearing here.

**3. Base branch should probably be `next`.**

Per the release-train table in `AGENTS.md`, a feature for the next feature release bases on `next`; `main` is for bug fixes, security fixes, and changes shared by both lines. An inbound ACP client harness is the former.

The description justifies `main` as "where ADR-0040 lives," but that doesn't distinguish the two — `next` is downstream of `main`, so ADR-0040 is on `next` as well. Rebasing onto `next` resolves 1, 2 and 4 at once: a free number, ADR-0119 present and linkable, and no index conflict.

**4. The PR is conflicting and 34 commits behind main.**

`mergeable: CONFLICTING`. The conflict is `docs/adr/README.md`, as expected — main has picked up 0115 and 0123 rows since this branched. Needs a rebase and a regenerated index either way.

### Design questions on the rewritten decision 4

These are the parts I'd most want settled before this goes to Accepted. Both come from decision 4 assuming ADR-0035's one-shot grant carries inbound unchanged; I don't think it does yet.

**5. Inbound, the grant's tool name has no trusted author.**

Decision 4 says approval leads to "a fresh turn with the ADR-0035 one-shot grant," and decision 6 asks conformance to prove "exactly-once execution."

The thing that makes that grant safe today is ADR-0046's replacement of the old summary-prefix sniff with runner-authored provenance. `ApprovalGate.block()` sets `pending_granted_tool` to the exact tool `can_use_tool` denied (`runner/src/curie_runner/approval.py:551-557`), described in the code as "the trusted, runner-held value the resume-turn grant binds to, never parsed from a string and never model-supplied."

Inbound there is no `can_use_tool` denial. The request arrives as a `session/request_permission` JSON-RPC message from a separate vendor process, so the tool name that scopes the grant is **peer-supplied** unless this ADR says otherwise. That's the property ADR-0046 deliberately removed, reappearing on the new edge. Worth stating explicitly: who authors the grant target for an inbound request, and what validates it against the manifest/bundle rather than trusting the peer's string. Until that's decided, decision 6's exactly-once claim isn't provable.

**6. The grant is boot-turn-only, and a rehydrated ACP agent may not reissue in time.**

`reset()` keeps a freshly injected grant through the first reset and clears it on every later one (`approval.py:521-534`) — deliberately, so it can't leak across turns.

Decision 4 requires the rehydrated ACP agent to reissue the operation for the grant to authorize it. But a fresh ACP process does `initialize`, opens a session, takes a prompt, and may well do some reading before retrying the write. If the reissue doesn't land inside the boot turn, the grant is already gone. The consequences section does acknowledge "no grant on rejection or expiry," but doesn't reconcile that with the boot-turn window.

Concretely: does one ACP `session/prompt` cycle map to exactly one Curie turn for grant purposes? If not, either the window needs widening or decision 4 needs to say how the reissue is forced to land on the boot turn.

**7. Small contradiction between decisions 4 and 5.**

Decision 5 says "the adapter is stateless with respect to durable state"; decision 4 says the adapter "persists a durable `Approval`." I assume the intent is that the adapter holds no durable state *of its own* and writes through the existing plane — worth wording it that way so the two don't read as opposites.

### Minor

**8. The PR description is stale relative to the second commit.** It never mentions ADR-0119 or #1902, which are now the ADR's headline blocking dependency, and the alternatives list grew from 6 entries to 8. Since "Assumptions recorded" is doing real reviewer-facing work here, it's worth refreshing — particularly the numbering assumption, which is the one that turned out wrong.

**9. Suggest a follow-up issue for cross-train ADR numbering.** Independent of this PR: the numbering rule and its CI gate are both single-branch, `main` and `next` have already diverged (two different 0115s, and 0123/0124 exist on opposite trains), and that will surface as a conflict at convergence rather than at PR time. Having `check-docs.sh` compare against the other train, and updating the "Claiming a number" section to say so, would stop this recurring.

### What's good

Worth saying plainly: the non-goals in decision 7 and the "honest tradeoff" consequence are the strongest parts of this. Naming that an inbound agent's tool call is gated by a plane it doesn't know exists, and that ACP normalizes the turn wire but not the installer, credentials, or bundle format — with the explicit warning against reading this as "a second harness is now free" — is exactly the kind of thing that's usually discovered a quarter into implementation instead. The amend-by-new-ADR approach also matches ADR-0045 and the ADR-0102 precedent correctly.

Happy to re-review once it's rebased and renumbered.

jw3329:
Heads-up on the ADR number: **0120 is already taken on `next`**, so this one needs renumbering before it can land.

- `origin/next` has `docs/adr/0120-durable-first-party-email-state.md`, merged in #1597 (`task/1515-first-party-email-channel`).
- This PR adds `docs/adr/0120-an-inbound-acp-client-admits-any-acp-harness.md`.

Because this PR is based on `main` and `main` hasn't taken the `next` ADRs yet, nothing conflicts today — the collision only shows up when the v0.7.0 train merges down, and by then both records are Accepted and immutable. Cheaper to fix here.

Current allocation across both trains:

| Number | Holder |
|---|---|
| 0118–0122, 0124 | merged on `next` |
| 0123 | merged on `main` (#1924) |
| 0125 | claimed by open #1934 |

So **0126** is the next free number.

For reference, #1911 and #1919 hit the same thing earlier this month (both branched as `adr-0122`); #1919 renumbered to 0124 pre-merge, which is the pattern here.

TheConnMan:
Rebased onto `next`, renumbered to ADR-0127, regenerated the index, and refreshed the PR metadata.

The approval questions are now decided explicitly: stable ACP v1 display fields are not authority; a generic peer stays read-only unless a Curie-controlled contribution supplies the canonical tool and input; the approval card, durable digest, resumed grant, and execution check bind to that same operation; cancellation returns ACP's required `Cancelled` outcome; and the grant has both prompt-cycle and absolute expiry. The argument-bound frozen-contract prerequisite is tracked in #1956, while structured replay remains blocked on #1902.

The adapter is now described as durably stateless while writing through the existing worker/API planes. `bash scripts/check-docs.sh` passes at `994c1c2a`.

The cross-train numbering process gap is tracked separately in #1957 without restoring strict required-branch freshness.

TheConnMan:
Reopening immediately to retrigger the required pull_request-scoped Secret Scan and CodeQL workflows after the base retarget; the manual Secret Scan dispatch scanned unrelated full history by design.

Issue #1872, opened 2026-08-24
The kernel writes a ledger record per side-effecting call


jw3329:
Second half of #1863, split out so the `kernel.py` diff arrives on its own. The tables land in #1871; nothing writes them yet.

`apps/worker/src/curie_worker/kernel.py` already has the branch that sees a `SideEffectFlag` — it sets `saw_side_effect` and persists a marker so a crash before done still blocks auto-retry. That branch is where a record gets written.

- **One call, one row.** The opening frame (`status='pending'`, arguments, no result) creates the row; the closing frame, matched on `call_id`, completes it — `status` from `failed`, `result`, `completed_at`. It must not mint a second row.
- **`dedupe_key` is `event_id` + `call_id`.** The worker redelivers at least once (ADR-0013), so the second delivery of one turn adopts the existing row. The DB constraint is the guarantee; the writer-side check is a race.
- **`prior_state` and `target` come out of the connector's structured reply**, not out of the arguments. What a restore replays is what the tool read before it wrote, and no function of the forward arguments can produce it.
- **`kernel.py` is sacred under ADR-0013.** `saw_side_effect` must still latch on the first frame and the no-retry rule must read exactly the signal it read before. The write is additive to that branch, not a rework of it.
- Snapshots can carry environment variables and secret references. They pass through the existing redaction path — note that `runner/src/curie_runner/redact.py` explicitly scopes itself to stdout and span attributes and leaves ACI frames untouched, so what "the existing redaction path" covers needs checking before this lands, not assuming.

After this, the ledger fills up on its own and #1864 can rule on an undo.

jw3329:
Landed in #1877 (merged to `next` as ec932503), validated on a live local stack: a real turn produced `POST /actions 201` from the worker and a row carrying the call's arguments; the completion set `prior_state` and flipped `undoable`; a replayed completion was refused. Details on the PR.

Issue #1864, opened 2026-08-24
The actions API: record, list, and rule on an undo


jw3329:
Slice 3 of #1861 (ADR-0117, decisions 2, 3, 4 and 5).

The API surface over the ledger from #1863: record an action, list a conversation's actions, and rule on an undo. The ruling is deterministic and the model is not in the path (decision 2).

Two refusal rules, and the feature lives or dies on the second:

- **Authorization symmetry** (decision 3). An undo requires the authorization the forward action required, and no more. If the tool was gated by an approval policy, undoing it needs an authorizer of that same route, resolved in the API against membership the way ADR-0034 resolves an approver. If the tool was not gated, the undo is not gated either. The state being restored is one the cluster was already in, reached without anyone approving it.
- **Conflict** (decision 4). Before replaying, compare the resource's live state to what the action left. On a mismatch, refuse and name both states. A blind restore silently reverts a human's manual fix, which turns an undo button into a way for the platform to fight the operator. A refused undo must change nothing at all.

A record does not expire (decision 5): no TTL, because a change one minute later makes a restore unsafe and a quiet week does not make it safer. The conflict check is the bound.

The ruling and the execution stay separate in the code. The API rules, records the ruling, and returns; whichever executor lands (see the last slice in #1861) cannot bypass the check by holding a connector's address.

jw3329:
Split, and resequenced ahead of #1872.

The worker has no database of its own — it persists an approval by POSTing to the API — so the kernel write path has nothing to write to until a recording endpoint exists. The prototype hid this: its demo called `POST /actions` itself.

- **#1874** — record, complete, list. The surface the worker writes through. Stacked on #1871.
- **this issue** — the undo ruling and its two refusal rules (authorization symmetry, conflict check), once #1872 fills the ledger.

Recording is authenticated by the shared API key like every other router; who may *undo* is the authorizer question and stays here.

Issue #1861, opened 2026-08-24
Implement ADR-0117: the action ledger and undo


jw3329:
Tracks the implementing work for **ADR-0117** (Accepted): a tool that changes the world reports what it changed, and the platform can put it back.

The decision is recorded in [`docs/adr/0117-a-tool-that-changes-the-world-reports-what-it-changed.md`](https://github.com/curie-eng/curie/blob/next/docs/adr/0117-a-tool-that-changes-the-world-reports-what-it-changed.md). A prototype argued it and was stripped before the ADR merged (PR #1805); it survives on the `spike/adr-0117-action-ledger` branch as a reference, not as code to land.

**The ledger invariant** every slice is reviewed against: every side-effecting call produces exactly one durable record; a record claims to be undoable only when it holds a prior state, a target, and a successful outcome; and an undo either restores the recorded state or changes nothing at all.

## Slices

Each lands as its own PR against `next`. The order is a dependency order, not a schedule.

- [x] #1862 — the ACI carries a call's arguments and result, one frame per call
- [x] #1863 — the ledger tables, and the kernel writes a record per call
  - [x] #1871 — the tables (no writer)
  - [x] #1872 — the kernel write path (needs #1874 first)
- [x] #1874 — the actions API: record, complete, list (the surface the worker writes through)
- [x] #1864 — the undo ruling and its two refusal rules
- [x] #1865 — a write connector reports the state it read before writing
- [x] #1866 — the receipt: a per-action card with an undo control or a stated reason
- [ ] #1867 — where the undo executor lives — **ADR-0121 merged as a Draft (#1893), which authorizes no implementation.** Blocked on acceptance, and on #1873 as its stated prerequisite.

#1867 is the decision ADR-0117 deliberately does not make: nothing in the platform can reach a connector today, so the API can rule on an undo but cannot perform one. Until that ADR lands, an undo is authorized and recorded but not executed.

TheConnMan:
Post-merge review evidence against the landed ledger slices found three gaps in the stated invariant that every side-effecting call has exactly one durable record and undo either restores recorded state or changes nothing.

1. #1874 completion is a read-modify-write without a conditional update or lock. Two concurrent conflicting completions can both see `pending` and the later write overwrites the first `prior_state` or target.

2. #1881 uses the same pattern for undo authorization. Two simultaneous requests can each observe `undone_at=NULL`, each write an `authorized` audit row, and both return a restore payload. The current regression is sequential only.

[...]

PR #1874, opened 2026-08-24
Add the surface the ledger is written through

jw3329:
**Stacked on #1871** (base is `task/1863-agent-actions-tables`; it retargets to `next` when that merges). Part of #1864, slice 3a of #1861.

## Why this comes before the kernel

The worker has no database of its own — it persists an approval by POSTing to this API. So [#1872](https://github.com/curie-eng/curie/issues/1872), the kernel write path, has nothing to write to until this exists. The prototype hid the dependency: its demo called `POST /actions` itself, standing in for the kernel.

## The two frames, as two requests

| | |
| --- | --- |
| `POST /actions` | the call was made — `pending`, carrying its arguments |
| `POST /actions/{id}/complete` | what came back — outcome, result, and the `prior_state` + `target` a restore replays |
| `GET /actions?conversation_id=` | oldest first, the order a receipt lists them |
| `GET /actions/{id}` | one record |

## Both idempotent, and the second one matters more

The worker redelivers at least once (ADR-0013). A replayed create adopting its own row is bookkeeping. **A replayed completion returning unchanged is a safety property**: `prior_state` is the thing an undo replays, so a completion allowed to rewrite it moves the target of a restore that may already have been offered to a human. The first account of a call is the one that was true when it happened.

## Deny-by-default falls out, again

A completion carrying no prior state and no target produces a record that is not undoable, and says why in `detail`. Nothing declares that — it's `undoable` being derived (#1871). A connector answering in prose lands there on its own, and so does a third-party MCP tool nobody wrote a connector for.

## What is deliberately not here

**Ruling on an undo.** It has its own two refusal rules — authorization symmetry and the conflict check — and they deserve a diff of their own rather than arriving under CRUD. Recording is authenticated by the shared API key like every other router; *who may undo* is the authorizer question, and it lands with the ruling.

## Verification

```
apps/api           975 passed, 10 skipped
  test_actions.py    9 passed   (real Postgres)
ruff check .       All checks passed
mypy               no issues in 201 source files
check-contracts    current
check-docs         drift-free, every citation resolves
openapi.json       regenerated
```

`test_control_integration.py::test_cost_composes_metrics_for_the_agent` fails locally and on an untouched checkout alike — needs Langfuse, which `--minimal` skips.

### DIFF
--- a/apps/api/openapi.json
+++ b/apps/api/openapi.json
@@ -1,6 +1,294 @@
 {
   "components": {
     "schemas": {
+      "ActionComplete": {
+        "description": "The closing frame: what came back, and what it takes to put it back.\n\n``prior_state`` and ``target`` are what a restore replays. Both are optional\nbecause a connector that answers in prose reports neither, and a record that\nholds neither is not undoable -- which is the honest answer rather than a\nmissing one.",
+        "properties": {
+          "detail": {
+            "anyOf": [
+              {
+                "type": "string"
+              },
+              {
+                "type": "null"
+              }
+            ],
+            "title": "Detail"
+          },
+          "failed": {
+            "default": false,
+            "title": "Failed",
+            "type": "boolean"
+          },
+          "prior_state": {
+            "anyOf": [
+              {
+                "additionalProperties": true,
+                "type": "object"
+              },
+              {
+                "type": "null"
+              }
+            ],
+            "title": "Prior State"
+          },
+          "result": {
+            "anyOf": [
+              {
+                "additionalProperties": true,
+                "type": "object"
+              },
+              {
+                "type": "null"
+              }
+            ],
+            "title": "Result"
+          },
+          "target": {
+            "anyOf": [
+              {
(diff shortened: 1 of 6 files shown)


jw3329:
**No checks will appear on this PR while it is stacked.** Every workflow in `.github/workflows/` gates on `branches: [main, next]` for both `push` and `pull_request`, and this PR's base is `task/1863-agent-actions-tables`. So the empty check list here means "never ran", not "passed" — please don't read it as green.

CI fires as soon as #1871 merges and this retargets to `next` automatically. Until then the verification is local only:

```
apps/api           975 passed, 10 skipped   (real Postgres)
  test_actions.py    9 passed
ruff check .       All checks passed
mypy               no issues in 201 source files
check-contracts    current
check-docs         drift-free, every citation resolves
```

Say the word if you would rather I retarget to `next` now to get CI immediately — the cost is that the diff then carries #1871's commit too.

Issue #612, opened 2026-07-17
A fresh scaffold is red under its own documented fake-model loop, and three scaffold comments still promise it is green


TheConnMan:
## Problem

A freshly scaffolded bundle fails its own eval under the loop its own generated docs prescribe.

`runner/src/agentos_runner/fake.py:55-57` returns a fixed literal with no bundle name:

```python
        _assistant(ToolUseBlock(id="t1", name="Bash", input={"command": "echo hi"})),
        _assistant(TextBlock(text="all done"), usage={"input_tokens": 20, "output_tokens": 8}),
        _result(text="all done", usage={"input_tokens": 20, "output_tokens": 8}),
```

`_default_script` only branches on the approval marker; otherwise `return default_turn()`. Nothing echoes the query, bundle, or agent name. The graded answer is the final text (`cli/src/evals.rs` `graded_answer`), i.e. `"all done"`.

#553 changed the scaffolded grader to require the name (`cli/src/scaffold.rs:72`):

```rust
                "grader": {
                    "kind": "contains",
                    "expected": name,
                    "case_sensitive": false,
                },
```

`"all done"` does not contain e.g. `deal-desk` -> RED.

## Failure scenario

`agentos init deal-desk` -> `agentos skill up --fake-model` -> `agentos skill eval` fails on an untouched scaffold. That is verbatim the generated `AGENTS.md` loop ("1. `agentos skill up --fake-model`" ... "3. `agentos skill eval` -- must be green before any deploy") and `cli/src/guide.rs`'s ordered walkthrough (:87, :99 "This is the promotion gate; it must be green"). A new user's first run of the documented path is red before they change anything.

## Three stale doc comments now assert the opposite

- `cli/src/scaffold.rs:53` — "The `contains \"\"` grader passes on any completed (`Done`) turn, so the scaffold is green out of the box under both `--fake-model` and a real credential"
- `cli/src/scaffold.rs:8` — "a single smoke case that passes on any [completed turn]"
- `cli/src/scaffold.rs:364` (test comment) — "contains \"\" (status-gated pass)"

All three describe the pre-#553 grader that no longer exists.

## Why CI missed it

- `cli/scripts/e2e.sh:45-46` writes its **own** cases.json: "The single grader is `contains \"all done\"` so it passes under the offline fake model".
[...]

PR #1877, opened 2026-08-25
Write a ledger record for every side-effecting call

### LINKED ISSUE #1863: The action ledger tables, and the kernel writes a record per side-effecting call
Slice 2 of #1861 (ADR-0117).

Persist what slice 1 now carries. `agent_actions` and `action_audit_entries`, shaped after `approvals` and `approval_audit_entries` in `apps/api/src/curie_api/models.py`, because the routing, lifecycle and audit needs are the same ones the approval tables already answer.

- `undoable` is a **derived** property, not a stored column, so a row cannot claim reversibility it holds no snapshot for. A record is undoable only when it holds a prior state, a target, and a successful outcome.
- The kernel branch that reads a side-effect frame writes exactly one record per frame. One call, one row.
- Snapshots can hold secrets: a Kubernetes object can carry environment variables and secret references. They pass through the existing redaction path, and a resource that cannot be snapshotted safely reports itself irreversible rather than storing credentials in the control plane.
- Write the migration by hand. Alembic autogenerate compares against the wrong search path and proposes dropping Langfuse's `monitors`, `table_view_presets` and `default_llm_models`, which share the database. Prove it round-trips upgrade, downgrade, upgrade against a real Postgres.
- `docs/interfaces/relational-db/INTERFACE.md` carries a counted JSONB column total that the docs gate checks; update it with the new columns.

Retention and pruning of the ledger are explicitly out of scope in ADR-0117.

### LINKED ISSUE #1872: The kernel writes a ledger record per side-effecting call

jw3329:
Slice 2b of #1861. Closes #1872, and closes #1863 together with #1871.

The ledger now fills on its own. This is the last piece before an undo can be ruled on.

## The change to `kernel.py`

**71 added lines, 4 removed**, in the branch that already existed — a `side_effect_flag` sets `saw_side_effect` and persists the no-retry marker. It now also records what the call did.

`kernel.py` is sacred under ADR-0013, so what did *not* change matters as much: the marker still latches on the first frame, so the no-retry rule and the false-completion check read exactly the signal they read before. Two signature changes (`_consume` and `_apply_frame` take the turn and the agent id rather than the event id alone) exist because a record belongs to a conversation and an agent, and threading three scalars would have been worse.

## One call is one row

The opening frame creates the record; the closing frame, matched on `call_id` through the accumulator, completes **that** record rather than minting a second.

`dedupe_key` is the event id **and** the call id together — the event id alone collapses a turn that called the same tool twice, and the call id alone doesn't survive a redelivery of that turn.

`prior_state` and `target` come out of the connector's structured reply, not the arguments, and `curie_worker/actions.py` is the one place that convention is read. No function of `scale_deployment(replicas=10)`'s arguments can produce the replica count from *before* the call — which is the whole reason the reply is the declaration (ADR-0117 decision 1).

## A refused write ends the turn

`ledger-error`, deliberately absent from `RETRYABLE_CLASSIFICATIONS`. The change to the world has already happened, so a retry would re-execute a side effect — the rule ADR-0013 already holds — and completing the turn would hand back a receipt the platform can't stand behind. The user sees "The run hit an error (ledger-error) after starting an action; not retrying automatically. Flagging for a human."

This is the judgement call in the PR, so it's worth stating plainly: **not best-effort.** A change the platform has no account of is not a success, and this same branch already fails the turn when the no-retry marker can't be persisted. Losing the record of *what* changed isn't the lesser failure.

## Two deliberate tolerances

| | |
| --- | --- |
| a frame with no `call_id` | a producer predating ADR-0117. Nothing recorded — two such frames can't be told apart — but the no-retry rule reads presence, which the frame still carries. ADR-0036's reader policy cuts both ways. |
| no recorder wired | records nothing rather than failing every turn that touches the world, exactly as an unwired approval backend degrades. Every existing kernel test builds one this way. |

## Verification

Baselined rather than asserted — the worker suite has pre-existing failures on this machine:

```
apps/worker (this branch)   27 failed, 981 passed, 7 skipped
apps/worker (clean stash)   27 failed, 977 passed, 7 skipped
regressions                 none — identical failure lists, diffed
```

The 27 are eval-stream and dead-letter-alerting tests that need services this machine isn't running. New coverage: 6 kernel tests, 6 client tests.

```
ruff check .      All checks passed
mypy              no issues in 202 source files
check-contracts   current
check-docs        drift-free, every citation resolves
```

### DIFF
--- a/apps/api/tests/test_actions_worker_contract.py
+++ b/apps/api/tests/test_actions_worker_contract.py
@@ -1,0 +1,89 @@
+"""The worker's ledger client against the real API, in process (ADR-0117).
+
+`apps/worker/tests/test_action_client.py` drives the client against a mock
+transport; `apps/api/tests/test_actions.py` drives the API against its own
+schema. Both pass while disagreeing about the body: rename a field on one side
+and each suite still goes green, because neither has ever seen the other.
+
+That seam is not on the ACI -- the ACI carries what the RUNNER emits, and this is
+the worker-to-platform hop -- so nothing else pins it. The e2e ladder does not
+close the gap either: it asserts plumbing, never reply content (ADR-0055), and a
+refused ledger write surfaces as an escalation that still finalizes with a reply.
+
+So this drives the real ActionClient over ASGI into the real router and the real
+database, and asserts what the row ends up holding.
+"""
+
+from __future__ import annotations
+
+from typing import Any
+
+import httpx
+import pytest
+from aci_protocol import SideEffectFlag
+from curie_worker.actions import ActionClient
+
+pytestmark = pytest.mark.usefixtures("clean_db")
+
+
+async def _round_trip(app: Any) -> dict[str, Any]:
+    """One side-effecting call, both frames, through the real stack."""
+
+    async with httpx.AsyncClient(
+        transport=httpx.ASGITransport(app=app), base_url="http://api"
+    ) as http:
+        from curie_api.config import get_settings
+
+        recorder = ActionClient(
+            api_base_url="http://api", api_key=get_settings().api_key, client=http
+        )
+        recorded = await recorder.record(
+            SideEffectFlag(
+                tool="scale_deployment",
+                call_id="toolu_01",
+                arguments={"name": "api", "replicas": 10},
+                detail="non-idempotent tool executed",
+            ),
+            event_id="event-1",
+            conversation_id="C-contract",
+            agent_id=None,
+        )
+        await recorder.complete(
+            recorded.id,
+            SideEffectFlag(
+                tool="scale_deployment",
+                call_id="toolu_01",
+                failed=False,
+                result={
+                    "ok": True,
+                    "prior": {"spec": {"replicas": 3}},
+                    "target": {"kind": "Deployment", "name": "api"},
+                },
+                detail="non-idempotent tool completed",
+            ),
+        )
+        fetched = await http.get(
+            f"/actions/{recorded.id}", headers={"X-API-Key": get_settings().api_key}
+        )
+    body: dict[str, Any] = fetched.json()
+    return body
+
+
+def test_a_recorded_call_survives_the_worker_to_api_hop(client: Any, anyio_backend: Any) -> None:
+    """Every field the worker sends is a field the API stores, under that name."""
+
+    import anyio
+
+    row = anyio.run(_round_trip, client.app)
+
+    assert row["tool"] == "scale_deployment"
+    assert row["call_id"] == "toolu_01"
+    assert row["arguments"] == {"name": "api", "replicas": 10}
+    assert row["dedupe_key"] == "event-1:toolu_01"
+    # The half a rename would silently break: the connector reported `prior` and
+    # `target` inside its reply, and the row has to hold them as the columns a
+    # restore replays.
+    assert row["prior_state"] == {"spec": {"replicas": 3}}
+    assert row["target"] == {"kind": "Deployment", "name": "api"}
+    assert row["status"] == "succeeded"
+    assert row["undoable"] is True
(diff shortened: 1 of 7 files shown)


jw3329:
Added a contract test for the seam this PR creates, and a correction to what I claimed above.

**The e2e rungs do not prove the ledger works.** All three passed (cluster rung on kind, 7m3s), but the ladder asserts plumbing and never reply content (ADR-0055, #612) — and a refused ledger write surfaces as an *escalation that still finalizes with a reply*. A broken worker-to-API body would read as a passing rung.

Nor do the two unit suites: the worker's client is driven against a mock transport, the API against its own schema. Rename a field on one side and both stay green. The seam is not on the ACI either — the ACI carries what the runner emits; this is the hop after it.

`apps/api/tests/test_actions_worker_contract.py` drives the real `ActionClient` over ASGI into the real router and the real database, and asserts what the row holds.

I mutation-checked it rather than trusting a first-run pass — renaming `prior_state` to `priorState` in the worker's body makes it fail. That experiment also showed how this seam breaks when it breaks: **the API accepts the unknown field silently and stores NULL**, so the action is quietly not undoable rather than loudly rejected. That silence is why the test earns its place.

jw3329:
Validated on a live local stack running this branch's worker image (rebuilt from `task/1872`, overlay rebuilt, `CURIE_BASE_TAG=dev`), against real Postgres at alembic 0029.

A real turn — real runner container, real ACI, real worker kernel, real API over HTTP:

```
worker log     POST http://localhost:28000/actions "HTTP/1.1 201 Created"
row created    call_id=t1  tool=Bash  arguments={"command":"echo hi"}  status=pending
completion     status=succeeded  prior_state={"spec":{"replicas":3}}  undoable=True
replay         prior_state stayed 3 — a second completion claiming 99 was refused
dedupe         1 row, 1 distinct dedupe_key
```

The replay line is the safety property on the deployed image: a redelivered completion cannot move the prior state an undo would replay.

**One hop is test-proven, not live-proven, and it is the fake model's limit rather than this code's.** `runner/src/curie_runner/fake.py::default_turn` emits assistant messages only — it never yields a `UserMessage`/`ToolResultBlock`, because the fake executes no tools. So an offline turn produces only the OPENING frame and the row correctly sits at `pending` (the same state a turn that dies mid-call leaves). I completed the row by calling the API directly, which exercises worker→API→DB but not runner→worker.

That hop is covered by `runner/tests/test_translate.py` on one side and `apps/worker/tests/kernel/test_action_ledger.py` on the other. Filed separately: the offline e2e ladder cannot exercise a tool result at all, which became true when the ACI started carrying them.

PR #1886, opened 2026-08-25
Fail the build when an example's write connector has no gate

jw3329:
The sre-bot example ships its write connector commented out, because a gate naming an undeclared connector fails bundle validation for everyone who never enables it. So enabling the connector and declaring its gate are **two edits in two files**, and nothing made them happen together.

**On a real install they did not.** The connector was enabled and the gate was not, and a tool able to roll production Deployments ran for four and a half days with nothing in front of it. Nothing reported it, because there is nothing to report — the connector is healthy, the tool runs, no approval card is ever posted. The absence of a gate has no symptom.

### The check

Source-derived rather than runtime-derived, so it runs on a checkout with no cluster and no image: a connector whose own `server.py` declares a tool that is **not** `readOnlyHint` must be named by an `approvalPolicy` gate. That keeps it conservative — it only knows about connectors this repository carries the source for, which is exactly the set an example bundle can enable.

It also compares the **two write allowlists**, both ceilings on what may be written, living in different files:

```
connectors.yaml            K8S_WRITE_ALLOWLIST     enforced in the connector
manifests/write-role.yaml  resourceNames            enforced by the API server
```

Those drifted apart on the same install when its target cluster changed, for four days. That direction was harmless because the intersection was empty. **The reverse is not:** an allowlist entry with no matching `resourceName` is a 403 *after* a human approved the call, which spends the approval and delivers nothing.

### Verified by reproducing the incident

Enabling `k8s-write` with no gate:

```
FAIL: sre-bot declares connector 'k8s-write', whose restart_deployment is not
read-only, but no approvalPolicy gate names mcp__k8s-write__restart_deployment.
Enabling the connector and declaring the gate are two edits; this is the one
that gets forgotten.
```

The shipped state passes, as do the four bundles with no write path at all — so this stays cheap for every example that never enables one.

The enablement steps in `connectors.yaml` now say that steps 3 and 4 are one change, and why.

Ref sre-bot#90, sre-bot#116


TheConnMan (changes requested):
Requesting changes. This catches a real incident, but the guard currently freezes the SRE-bot's bespoke connector-as-policy workaround instead of enforcing the platform-level authorization invariant, and it does not yet carry a durable negative regression.

The target model should be builder-owned policy over the connector's published tool surface: every tool is denied, approval-required, or allowed, with unclassified tools denied by default. A connector publishes capabilities; a skill owns deterministic workflow logic; neither may silently widen the builder's authority. Until that tri-state contract exists, please frame any incident-specific check here explicitly as a temporary guard rather than as proof that "every write path is gated."

There are also concrete correctness gaps in this implementation:

- `write_tools` treats every decorated function in a file as a write tool whenever the file contains the exact literal `readOnlyHint=False`. A mixed read/write server false-positives its read tools, while equivalent annotation expressions or formatting can false-negative. It does not enumerate the actual published tool metadata.
- The workflow runs this only against the passing repository. A new guard needs a committed red-path regression that builds an isolated violating bundle, executes the real checker, and asserts non-zero plus the actionable diagnostic. Please cover at least a missing gate and both allowlist-drift directions.
- The placeholder bypass skips all allowlist comparison when any `<` or `>` occurs, so a mixed real target plus placeholder passes vacuously. Match the exact sanctioned placeholder shape instead.
- The RBAC collector unions every `resourceNames` entry from every Role rule without checking API group, resource, or verb, so an unrelated named resource can make a Deployment allowlist appear aligned.

The incident deserves a guard; these changes are necessary so that the guard proves the property it claims and does not become the architecture for connector authorization.

COVERAGE: 2 files touched; both changes are additive, so no modified prior lines had recoverable intent and the standalone prior-intent pass had nothing to check.

jw3329:
Addressed all four correctness gaps and the framing. Pushed in e5a958f1.

**Framing.** The docstring now leads with what this is *not*: it reads only source this repository carries, `readOnlyHint` is a hint rather than a boundary, and passing means two specific mistakes are absent from an example bundle — not that every write path is gated. It names the model it stands in for (builder-owned policy over the published tool surface; deny / approval-required / allow with unclassified denied by default; enforced at call time) and says this script should be deleted rather than extended once that exists. The success line no longer asserts the general property.

**`write_tools` → `tool_surface`.** Now AST-parsed and resolved per tool: module-level `ToolAnnotations` constants are collected, each tool decorator's `annotations=` argument is read, and `name=` overrides are honoured. Against the shipped tree, `k8s-write` and `k8s-scale` classify one write tool each and `tempo`'s four all classify read-only — the old file-level match would have claimed all of them given one write annotation or a prose mention of the literal. A tool whose `readOnlyHint` cannot be read is now reported as unclassified rather than passing.

**RBAC scoping.** Only `resourceNames` on rules granting a write verb over `apps/deployments` (including the scale subresource) count. Two regression cases cover it: a ConfigMap grant naming the same resource, and a read-only Deployment grant.

**Placeholder bypass.** Matches the exact `<name>/<name>` shape, all entries or none. Mixed placeholder plus real entry is now an error rather than a silent skip.

**One more found while fixing that.** Allowlist discovery was `K8S_WRITE_ALLOWLIST` only, but the bundle already ships a second write connector whose ceiling is `K8S_SCALE_ALLOWLIST` — it was not being compared at all. Discovery is now `K8S_*_ALLOWLIST`, with a test for it.

**Red path.** `examples/tests/test_write_path_gated.py` builds isolated violating bundles and runs the real script as a subprocess through a new `--examples` flag, asserting non-zero exit plus the actionable diagnostic. Covers a missing gate, both allowlist drift directions, unclassified tool, mixed placeholder, both RBAC-scoping cases, the second connector's allowlist, the mixed read/write false positive, and an empty examples directory failing rather than passing vacuously. To check the tests are load-bearing rather than decorative I reverted each fix in turn and confirmed only the matching test failed. 14 tests.

Local: mypy strict clean, ruff check and format clean, 14 passed, guard green on the real tree.

Two things I did not do, so they are visible rather than assumed. The guard still reasons about `apps/deployments` specifically, so a bundle whose write path is some other resource is not covered — stated in the docstring's limits rather than silently. And the tri-state contract itself is out of scope here; if you want an issue opened to track it against this scaffolding, say so and I will file it with this script named as the thing it replaces.

jw3329:
Added the piece your second point asked for. Pushed in 14139659.

**"It does not enumerate the actual published tool metadata."** Correct, and the source-derived check cannot — it reads `server.py` on a checkout with no cluster and no image, which is what makes it cheap and also what bounds what it can know. So this adds the half that does enumerate: start the connectors, perform the MCP handshake, call `tools/list`, and compare the declared gates against what each connector actually publishes, including its real `readOnlyHint` annotations.

`scripts/assert-gates-are-live-tools.py`, ported from the downstream repository. That repo is archived and read-only, so this is the only remaining path for it. Adapted to take `--bundle` instead of assuming one manifest location, since curie carries several example bundles.

**The two checks catch different failures and neither subsumes the other.** Source-derived catches a write-annotated tool with no gate, before anything is built. Runtime-derived catches the reverse — a gate naming a tool no connector publishes — which is invisible to source analysis and fails in the worst way available: the bundle validates, the deploy succeeds, the connector is healthy, the tool runs, and no approval card is ever posted. It is also easy to reach by accident, because curie's own deploy-time error message recommends the plugin-prefixed form for connector tools and following that advice produces exactly this. There is a test for that case specifically.

**Seven tests, and the stub is a real client's peer rather than a permissive mock.** It issues an `Mcp-Session-Id` on initialize and rejects any later request that omits it, and it will not answer `tools/list` before `notifications/initialized`. That is deliberate: an earlier version of this script skipped both steps, passed against a lenient stub, and then returned HTTP 400 the first time it met an actual FastMCP connector. A guard nobody can keep green gets deleted.

Load-bearing, checked by reverting: removing the session header from the script fails six of the seven tests; forcing every tool to read-only fails two.

Covered — gate naming a published write tool passes; gate naming nothing fails and names the closest live tool; published write tool with no gate fails; read-only needs none; a tool publishing no annotation counts as write rather than as safe; an unreachable connector fails rather than skipping, since an unproven assertion is what this script exists to stop anyone assuming.

**One thing I did not do:** wire it into CI. It needs live connectors, so it belongs in a lane that has them, not in the source-only job the other check rides. The tests prove the script works; they do not pretend it runs against real connectors anywhere yet. Tell me which lane you want it in and I will wire it.

Local: mypy clean, ruff clean, 42 tests in `examples/tests` pass.

jw3329:
Ran the runtime checker against the real connectors for the first time. It found a defect in itself, and it found a gap in what this PR claims to cover. Pushed in 811b6323.

## The run

Four live connectors in a real install, `tools/list` after a real MCP handshake:

```
  kubernetes         13 tools (0 write)
  grafana            48 tools (0 write)
  tempo               4 tools (0 write)
  k8s-write           1 tools (1 write)
all 1 gate(s) match a live tool; all 1 write tool(s) are gated
exit 0
```

66 published tools, exactly one not `readOnlyHint`, and it is the one the single gate names. That is the property you asked for enumerated from the published surface rather than inferred from source, and it is now measured rather than asserted.

## What the run broke, and the fix

The first attempt failed closed on `grafana`: HTTP 403 to every request. The script reports that as an unreachable connector, so a healthy connector read as a dead one and I went looking for a dead pod.

The cause was the `--host` flag I added for exactly this case. A connector behind `-allowed-hosts` is given entries that carry a port, and a Host header without one does not match. Measured on that connector:

```
Host: <svc>.<ns>.svc.cluster.local:8000   -> 200
Host: <svc>.<ns>.svc.cluster.local        -> 403   forbidden: host not allowed
```

Nobody writing `--host name=<svc>` means "and drop the port". So the URL's port is appended when the value has none, and a 403 now names the Host that was sent and points at `--host` — that is the one failure whose cause the exception hides. Two tests against a stub that enforces the check the way the real connector does; reverting the append fails the first and leaves the other eight green. 44 tests in `examples/tests` pass, ruff and mypy clean.

## The gap this PR does not cover, before you find it

**Both checks in this PR are about the write path. The read path has no gate at all, and its blast radius is decided by a field neither check reads.**

`manifests/read-access.yaml` mints a **ClusterRole plus ClusterRoleBinding, under names with no release prefix**, and the only thing bounding what the bot can read is which namespace the binding's subject lives in. The file's own comment invites changing it ("Change all four if you install the release somewhere else"). Following that advice into a tenant's release namespace hands that tenant cluster-wide `pods/log` and `configmaps` — measured on a cluster where it had happened:

```
list pods --all-namespaces        yes    <- what the file intends
get secrets --all-namespaces      no     <- the three no's the file asks for
delete pods --all-namespaces      no
create pods/exec --all-namespaces no
get pods/log -n <other-tenant>    yes    <- not intended by anyone
get configmaps -n kube-system     yes
```

The RBAC is exactly as designed; the *ownership* is wrong, and nothing represents ownership. Two releases on one cluster share the single unprefixed ClusterRoleBinding, whose subject list holds one entry, so the second apply silently rebinds the first — adjacent to #1535. Filed separately rather than widened into this PR.

So to your framing point: this PR is a temporary guard over one bundle's *write* path, and it should not be read as evidence that authorization is handled. The tri-state contract you described (deny / approval-required / allow, unclassified denied, enforced at call time, owned by the builder over the published tool surface) also has to cover read grants, which neither of these scripts can see.

## Your open question

You asked which CI lane the runtime check belongs in. It needs live connectors and it needs to reach them, and reaching them is the part that decides: the connectors' own NetworkPolicy admits ingress from runner-sandbox pods only, so the platform's own API pod is refused — verified, the deny is enforced. A lane running this must therefore either carry a pod with the sandbox labels or address the connectors from the node. The kind rung of the E2E parity ladder is the only existing lane that has both a cluster and those labels available. Tell me if that is where you want it and I will wire it there; I have not, because guessing at the lane is how it lands somewhere it cannot stay green.

Also still open from your review: whether you want the tri-state contract issue filed with these scripts named as the thing it replaces. Say the word and I will file it.

jw3329:
Filed the decision you offered to track: **Draft ADR-0126** in #1942.

It answers the framing point directly rather than around it. Your review said this check freezes a bespoke connector-as-policy workaround instead of enforcing a platform-level invariant, and named the target: builder-owned policy over the published tool surface, deny / approval-required / allow, unclassified denied. The ADR's decision is what gives that tri-state a subject it can actually attach to — **a tool name cannot carry three classifications, a field path can** — so the unit of authorization becomes `(resource kind, field path, value constraint)` and one tool covers restart, scale and whatever comes next.

Two things in it are relevant to this PR specifically:

- **Both scripts here are named as things the decision deletes**, not extends. They reason about tool names, which is the vocabulary the ADR replaces. That is the framing you asked for, made structural instead of a docstring promise.
- **It records the passthrough alternative properly rather than dismissing it.** A maintainer asked the reasonable version of it — why not one thin write tool the agent controls, instead of a connector per verb — and it deserved measured costs on both sides. Per-verb: 580 lines across two near-duplicate files for two operations, and `k8s-scale` merged and still undeployable for want of a published image. Passthrough: the ceiling moves to the persuadable component, and this bot's read grant is cluster-wide `pods/log` and `configmaps`, so its ordinary input is attacker-writable text.

This PR still stands on its own as the temporary guard, framed as one. The ADR is Draft, so it authorizes nothing yet.

TheConnMan (approved):

PR #1889, opened 2026-08-25
Desktop App UI/UX

alexrao2000:
## What

A native operator console for Curie, at `apps/desktop`. Electron shell plus a
React renderer, driving the `curie` CLI and the platform API from one window.

`apps/ui` is the web console and can only ever do what a browser tab can do,
which leaves out most of the product: the `skill` and `local` tiers are
containers on your machine, `curie init` and `curie build` touch your
filesystem, `curie secrets` writes to your private storage, and `docker stats`
is not something a web page can read. This is the surface that can drive the
whole parity ladder from one place.

## Why it is not a worse experience than the CLI

The command surface is generated, not written. Every command in
`cli/command-manifest.json` gets a real form built from its own declared
arguments, help text, allowed values and defaults. A command added to the CLI
appears after `pnpm gen:manifest` with no screen to build, and a removed flag
cannot linger because there is no hand written copy to go stale.

That claim is enforced three ways rather than asserted:

1. A coverage test asserts the app exposes exactly the commands the manifest
   declares, omitting only the ones clap itself marks hidden.
2. The string the UI shows and the argv the shell builds are separate
   implementations, compared against each other across every command with every
   flag filled. A preview that lies is worse than no preview.
3. Startup asks the installed binary for its own schema and reports both
   directions of drift in Settings. A command the app offers that the binary
   lacks is a broken button; a command the binary has that the app lacks means
   the app has quietly become the lesser surface.

This got a live proof during review. The branch was moved from `main` to `next`
mid change, `next` carries five commands `main` does not, and all five appeared
with correct tiers and no code change. `local delete` was classified destructive
by the name heuristic that exists for exactly that case.

## Surfaces

| View | Answers |
| --- | --- |
| Overview | What is happening, ordered by urgency. Anything blocked on a human first. |
| Resources | What each agent consumes. Usage over the daemon's capacity, compose projects as collapsible parent rows, ports and image as columns. |
| Canvas | How it is wired, derived from live state every render. Infrastructure placed by its position in the message path. |
| Commands | All 85, grouped by the parity ladder. |
| Activity | Every invocation with its full transcript, exit code and duration. |

## Verification

Run from `apps/desktop`:

```
pnpm install && pnpm lint && pnpm typecheck && pnpm test && pnpm build
```

131 unit tests, zero lint warnings, clean typecheck, both bundles build. A new
`desktop` CI job mirrors the `ui` job and fails on manifest drift.

Two integration suites skip themselves when their dependency is absent and were
run here against the real thing:

- `cli.integration.test.ts` runs `curie <command> --help` for every command both
  sides have, proving the argv the app builds is argv the CLI accepts.
- `resources.integration.test.ts` runs the resource feed against the live Docker
  daemon and asserts the contract holds on whatever containers happen to be up.

`scripts/check-docs.sh` and `scripts/check-commit-messages.sh` both pass.

## Bugs this found in itself

Recorded because each one is now a test, and the invariant is written down next
to the code in `apps/desktop/CLAUDE.md`.

- Rows keyed by a truncated container id collided, rendering one container twice
  and dropping a sibling section's header. Keyed by name now, which Docker
  guarantees unique.
- Container roles were the raw compose service name, so `curie-api` never
  matched a consumer looking for `api`. The canvas silently dropped most of the
  platform and failed every edge lookup: three orphan boxes and no edges. Roles
  are canonical now, with the real service names from a running stack as the
  regression test.
- A plain click on a canvas node was saved as a custom layout, which then pinned
  every node and permanently disabled relayout. Only a real drag persists now,
  and saved coordinates carry the layout version that produced them.
- Pinning the CPU chart's axis to the machine ceiling drew a genuine 95 percent
  load as a flat line at 8 percent height. The axis follows the data; the
  caption carries the denominator.
- Command forms seeded `--yes` and `--dry-run` as ordinary fields, so a dry run
  carried a confirmation flag the form never showed.

## Conventions, and two deviations to flag

Per `AGENTS.md`, "a UI surface ... is a feature, not an architectural decision",
so this carries no ADR. It should carry a tracking issue and I have not opened
one; happy to, or link an existing one and I will add the reference.

Base is `next`, per the release train table: this is a feature for the next
feature release. The branch is named `claude/...` rather than
`task/<short-description>` because the change was asked to stay on the branch it
was developed on.

## Shell decision: Electron, deliberately

Tauri was considered and rejected, and the reasoning is now recorded in
`apps/desktop/README.md` rather than left implicit.

Nothing in this renderer touches Chromium's differentiated capability. There is
no WebGL, WebRTC, video, wasm, worker or untrusted web content; it is CSS grid,
hand drawn SVG and IPC, all of which runs on WebKit. So Chromium is not buying
capability. What it does buy is one engine to verify against instead of three,
and the DevTools protocol, which is not theoretical here: the canvas layout bugs
below were found by driving the live native window over CDP because the
environment had no way to send it a keystroke.

The Tauri upside was a smaller download and no bundled Chromium. On Windows its
webview is WebView2, which is Chromium, so even that would have held on two
platforms of three. Not worth the port.

This also closes the WebKit question, which was an open unknown rather than a
verified fact. It is now moot. The renderer stays shell agnostic anyway, because
that costs nothing and keeps the decision reversible: everything privileged
crosses one contract file, and drag regions carry both attributes.

## Canvas

The graph is now legible rather than merely correct: lane headers behind the
columns, live load per container drawn against one core, hover and selection
lighting a node and its neighbours so a path can be traced, edge labels on chips
so they survive the dot grid, and inline chips naming each absent source when
there is no bundle open or no API reachable.

Its layout arithmetic is tested in `src/graph/model.test.ts`, which is where the
four bugs below are pinned.

## Known follow ups, not in this PR

- **No Playwright suite.** `apps/ui` has one; this app has 136 unit tests and two
  integration suites but no end to end coverage yet.
- **The `example` command group lands under Platform tooling** in the Commands
  view rather than its own tier. Cosmetic, and only visible on `next`.


alexrao2000:
CI status, separating this change from what `next` already carries.

**Green, and attributable to this PR:** `Desktop (lint + vitest + build)`, plus
`UI`, `Rust`, `Contracts`, `Commit messages (no AI attribution)`, `CodeQL`,
`gitleaks`, all four audit jobs and every image build. 25 checks pass.

**Pre-existing on the base commit, not caused by this change:**
`Python (ruff + mypy + pytest)` fails at the `Alembic revision gate` step, and
the three `E2E parity ladder` jobs plus `E2E required` fail. The run for
`next` at `8f9c630b`, which is this branch's exact base, fails the same jobs at
the same steps. This PR touches no Python, no migrations and no e2e harness.

One fix went in after the first push: the root `.gitignore` rule `*secret*`
silently excluded `apps/desktop/electron/ipc/secrets.ts`, so `git add` skipped it
without a word and the Desktop job failed at typecheck on a missing import. The
negation now sits with the five existing ones and carries the same
justification: that file handles secret names only, never material.

alexrao2000:
Two updates.

**Shell decision settled: staying on Electron.** Tauri is off the table and the
reasoning is recorded in the app README rather than left implicit. The short
version is that nothing in this renderer uses Chromium's differentiated
capability, so the dependency is buying one engine to verify against instead of
three plus the DevTools protocol, and the Tauri upside was only a smaller
download on two platforms of three. A useful side effect: WebKit compatibility
was previously listed here as unverified, and it is now moot rather than unknown.

**Canvas is legible.** Lane headers, live per container load, hover and selection
path tracing, edge labels on chips, and inline chips naming each absent source
when nothing is connected. Layout arithmetic is under test in
`src/graph/model.test.ts`, which is where the four canvas bugs listed in the
description are pinned. 136 unit tests now.

alexrao2000:
## Demo: build → deploy → talk to it → manage

75 seconds against a local stack. Captions are burned in, so there is nothing to enable.

**What it covers**

1. **Build** — the new-agent wizard: pick what it starts from, name it and choose where it lives, and see exactly what will be written before anything is.
2. **Configure** — what the agent should do, and its description, as fields rather than files to go and find.
3. **Deploy** — *Put it to work* opens the generated form, which shows the exact line it will run; the output lands in the console and the agent reports live.
4. **Reach it** — deployed is not the same as reachable. The *Answer in Slack* panel hands you the app manifest to paste at api.slack.com, the two tokens to collect, and a button that hands them to Curie.
5. **Use it** — send the agent a message and watch it answer.
6. **Manage** — the Canvas, derived from what is actually running: the agent you just built and the channel it answers in. Then Resources, and its own sandbox, named.

**Fixes visible in it**

- Resources attributes a running sandbox to its agent (`runner · shift-notes`). It used to read *role: other, agent: not attributed* — the one container that was an agent doing work was the one the tab could not name, under a heading that says "what each agent is using up".
- Agent rows show their real bindings. The app read `channel.channel_id`, a shape the API stopped sending when bindings went plural (ADR-0118), so every view said "no channel bound" about agents that had been answering in Slack for days.
- The generated form's sheet is titled for what you pressed ("Put it on this computer"), not for the command it runs. The exact invocation is still in the form, copyable.
- The content column is centred, and the bottom fade only runs when there is a console to fade into.

alexrao2000:
### The recording

https://github.com/curie-eng/curie/raw/claude/agentos-curie-desktop-ui-ea328e/docs/demo/desktop-console.mp4

Committed at [`docs/demo/desktop-console.mp4`](https://github.com/curie-eng/curie/blob/claude/agentos-curie-desktop-ui-ea328e/docs/demo/desktop-console.mp4) so it outlives this thread — same place the README's demo loop lives. 75 seconds, captions burned in, 1.1 MB.

alexrao2000:
### The walkthrough

Curie Desktop: build an agent through the wizard, deploy it, paste the Slack app manifest, message it and watch it answer, then find it on the Canvas and its own sandbox in Resources

75 seconds, captions burned in. Full-resolution mp4 is not in the tree — this GIF is, at [`docs/demo/desktop-console.gif`](https://github.com/curie-eng/curie/blob/claude/agentos-curie-desktop-ui-ea328e/docs/demo/desktop-console.gif), beside the README's own demo loop.

alexrao2000:
Closing in favour of #2419.

Same work, rebased onto `next`. The rebase was substantial enough that force-pushing this branch would have detached the review comments anyway, so a fresh thread is easier to follow than a rewritten one.

What changed beyond the rebase: console sign-out (the API could sign a console in and never out), an Access panel that says what is authorizing the window, sign-in bound to a subject the way ADR-0106 now requires, and the scratch agent bundles that got committed at the repo root during a demo recording are untracked and ignored.

PR #1891, opened 2026-08-25
Write down what the sre-bot write path may do, and where each ceiling is

jw3329:
The first acceptance criterion for the staging write-path work is that the write actions are *enumerated in a permission map*, and that "full access" is not an unbounded credential. This is that map, and it is opened for agreement rather than merged as settled — the second entry contains a design decision I do not think I should make alone.

### Entry 1 — `restart_deployment` (shipped, off by default)

The point of this entry is that the grant is wider than the verb:

`kubectl rollout restart` is not a distinct permission. It is a PATCH of the pod template, so `patch` on `deployments` is the *same grant* as `set image`, `set env`, and replacing the container command. RBAC cannot express "patch only this field".

So the ceilings are ranked strongest-first, because the weakest is the easiest to mistake for the whole answer:

1. the connector — two parameters, patch body is a constant in source
2. `resourceNames`
3. the allowlist (same bound, stated again; #1886 enforces they agree)
4. the approval gate — **ranked last on purpose**: it is a control on the agent, not on the credential, so the Role is scoped as though it did not exist

### Entry 2 — `upgrade_release` (proposed, and the reason this file exists)

Entry 1 is narrow because the *credential* is narrow. That is not available for an upgrade. Measured on a live staging release, one upgrade touches:

```
Deployments 6   StatefulSets 4   Services 7   PVCs 4
RBAC objects 9  Jobs 3   ConfigMaps 2   Secrets 1
+ 4 cluster-scoped CRDs   + schema migrations   + Helm release state
```

**There is no RBAC expression for "may upgrade Curie."** The nearest honest one is namespace-admin plus CRD write — an unbounded credential in every sense the criteria exclude.

The proposal moves the ceiling entirely into the connector (one tool, one parameter, validated against a version allowlist) and states plainly that the credential is broad, rather than implying RBAC is doing work it cannot do.

### Four questions left open

Two of them decide whether that shape is right at all, so answering them in the file would have been guessing. The sharpest:

> The release being upgraded contains the worker running the turn that requested the upgrade, and the approval resumes *through* that worker restarting. Whether that survives is answerable by testing on a staging install, not by reasoning about it.

The others: how the connector actually performs a Helm operation, what rollback means when a migration is not reversible (0.7.0 moved `agents.slack_channel` into its own table), and who maintains the version allowlist.

Ref sre-bot#90

### DIFF
--- a/examples/sre-bot/docs/PERMISSION-MAP.md
+++ b/examples/sre-bot/docs/PERMISSION-MAP.md
@@ -1,0 +1,140 @@
+# Write path permission map
+
+Every write this bot can perform, what each one actually grants, and where its
+ceiling really is. Written because "give it write access" is not a decision —
+it is the absence of one, and the interesting question is always *which* write,
+bounded by *what*.
+
+Two entries. The first ships and is off by default. The second is proposed and
+is a different kind of problem, which is most of why this file exists.
+
+---
+
+## 1. `restart_deployment` — rolling a named Deployment
+
+| | |
+|---|---|
+| Tool | `mcp__k8s-write__restart_deployment(namespace, name)` |
+| Kubernetes verbs | `get`, `patch` on `apps/deployments` |
+| Scoped by | `resourceNames`, one namespace |
+| Gate | `approvalPolicy`, human approves each call |
+| Status | Implemented, ships commented out |
+
+### What that grant actually permits
+
+`kubectl rollout restart` is not a distinct permission. It is a PATCH of the pod
+template that sets an annotation, so `patch` on `deployments` is the same grant
+as:
+
+```
+kubectl set image deploy/<name> app=something-else
+kubectl set env   deploy/<name> ANTHROPIC_API_KEY=...
+kubectl patch     deploy/<name> -p '{"spec":{"template":{"spec":{"containers":[{"command":[...]}]}}}}'
+```
+
+**A grant that permits a restart permits replacing what runs.** Kubernetes RBAC
+cannot express "patch only this one field", so nothing at the RBAC layer
+separates them.
+
+### Where the ceiling really is
+
+Ranked by strength, because the weakest one is the easiest to mistake for the
+whole answer:
+
+1. **The connector.** The tool takes `namespace` and `name` and nothing else, and
+   the patch body is a constant in the source. There is no parameter through
+   which a caller reaches an image, a command, or an env var. The agent is
+   prompt-injectable; this process is not.
+2. **`resourceNames`.** Bounds which workloads, so even a compromised connector
+   reaches only what an operator named. The only narrowing RBAC offers here.
+3. **The allowlist.** The same bound stated again in the connector's own config,
+   because a ceiling written in one place moves when someone edits the other.
+   These two must agree; `scripts/check-write-path-gated.py` enforces that.
+4. **The approval gate.** Stops a call before it executes — but it is a control
+   on the *agent*, not on the *credential*. If the credential leaks the gate is
+   irrelevant, so the Role is scoped as though the gate did not exist.
+
+### Deliberately not granted
+
+- `pods delete` — a restart with worse failure modes: no surge, no rollback, no
+  record on the Deployment. `rollout restart` covers the same intent safely.
+- `deployments scale` — a separate blast radius; scale-to-zero is an outage.
+  Add it with its own gate when someone asks, not pre-emptively.
+- Anything cluster-scoped, anything in `kube-system`, any `create` or `delete`.
+
+---
+
+## 2. `upgrade_release` — moving Curie from one version to the next
+
+**PROPOSED. Not implemented. The design question below is the reason this file
+exists rather than a Role.**
+
+### Why this one does not fit the pattern above
+
+The first entry is narrow because the *credential* is narrow. That is not
+available here. Upgrading a Curie release touches, in one operation:
+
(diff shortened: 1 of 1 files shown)


TheConnMan (changes requested):
Requesting changes on the ownership model, not on the incident evidence or the RBAC analysis.

The proposed `upgrade_release` entry moves the boundary "entirely into the connector" and makes the connector own the constructed operation and version allowlist. That is the wrong abstraction for agent builders. A connector/MCP server should publish tools that interact with an external system. The skill should own deterministic workflow logic and sequencing. Builder-owned config should classify each published tool as denied, approval-required, or allowed; a skill must not be able to expand that authority.

Please rewrite the proposed path around those boundaries:

- Kubernetes/Helm connectors publish the available external capabilities.
- The skill owns upgrade orchestration, templating, verification, and recovery instructions.
- Agent config decides which tools are denied, require one-shot approval, or are allowed.
- RBAC, scoped credentials, connector validation, and narrow parameter surfaces remain defense in depth, but do not substitute for the builder's authorization policy.

Curie currently has approval-required and an implicit allow-by-omission posture, but not a complete builder-owned deny/approval/allow contract. If that contract is a prerequisite for a safe upgrade workflow, this document should say so rather than solve the gap with a bespoke policy-bearing connector.

Entry 1's factual warning about Kubernetes `patch` is valuable and should stay. Please revise the claim that the connector is the agent's "real ceiling" so it distinguishes the connector's capability surface from the builder's authorization decision.

COVERAGE: 1 new file; no modified prior lines had recoverable intent, so the standalone prior-intent pass had nothing to check.

jw3329:
Rewritten around those boundaries. Pushed in 974aae0e.

**The two axes are now separated explicitly.** The document opens by splitting what it had collapsed into "ceiling": a connector's **capability surface** (what is reachable against an external system) versus the builder's **authorization decision** (whether this agent may call a published tool). Stated as non-substitutes, with the failure mode in each direction — a narrow surface with no authorization is the ungated write this bundle already shipped once; an authorization decision over a surface wide enough to do something else is a gate on the wrong thing.

**Entry 1.** The `patch` analysis is unchanged. What changed is the claim built on it: the two-parameter signature and constant patch body are now described as defense in depth against a prompt-injectable agent, which is what they are, rather than as the agent's real ceiling. The `approvalPolicy` gate is named as the authorization decision. I also wrote down the allow-by-omission posture there, since it is what makes a forgotten gate a silent grant rather than a refusal.

**Entry 2.** `upgrade_release(target_version)` is withdrawn — you are right that it makes the connector the policy holder, and it also makes every different upgrade policy a different connector image. Replaced with: Helm/Kubernetes connectors publish ordinary external capabilities (`helm_upgrade`, `helm_history`, `helm_rollback`, `helm_status`, reads); the skill owns preflight, ordering, verification and recovery; agent config classifies each published tool denied / approval-required / allowed, with the version allowlist as builder config because "may this agent move to that version" is an authorization question, not a capability one. There is a per-tool classification table. Defense in depth is listed as such and explicitly not as the policy.

**The prerequisite is stated rather than solved locally.** Curie has approval-required plus allow-by-omission and no builder-owned deny/approval/allow contract with unclassified denied by default. The document says why entry 1 survives that gap and this workflow does not — a repository lint can compare a small set of source-carried connectors against declared gates, but not a general-purpose Helm connector where one tool added in a future version silently widens the agent — and concludes the contract is a prerequisite for shipping the workflow.

**Three of the four open questions are now measured**, from a local `0.7.0 → 0.7.1 → 0.7.0 → 0.7.1` run rather than reasoned about. The headline: an approval-suspended turn is resumed by the API's sweeper, not by the worker being replaced, so the bot can upgrade the release it runs in *provided approval precedes execution* — demonstrated in miniature by having the bot restart the API that hosts that sweeper, with its own turn still completing. An in-flight turn not suspended on an approval is held by the dead consumer for minutes (808s and 428s observed), so that ordering is load-bearing and the skill's sequencing must not invert it. Also: rollback is a Helm operation only where nothing migrated, since migrations run as an api initContainer and `helm rollback` does not downgrade; and the Helm history is not a usable change record on its own, because every `cluster up` on a runsc-less cluster records a failed revision first, so a bare `helm rollback` lands on one (curie#1899).

**Two inherited claims were wrong.** While checking the numbers against the chart rather than carrying them over: the `agents.x-k8s.io` CRDs are not in this chart, so an upgrade does not carry them — a CRD bump is a separate cluster-scoped step and a separate, stronger authorization question than the one this tool was arguing about. And the four PVCs come from StatefulSet `volumeClaimTemplates`, which are immutable on update, so storage is the part an upgrade cannot change; that asymmetry is why recovery is not the inverse of upgrade. Both corrections are in the document with the argument they change. The fixed object inventory is replaced by the command that re-derives it, since it moves with which components are enabled.

One scope note: I did not open the tri-state contract issue, since it is the platform's to shape and you may already have a form in mind. Say the word and I will file it with this document's prerequisite section and `scripts/check-write-path-gated.py` named as the scaffolding it retires.

TheConnMan (approved):
The revised permission map incorporates the maintainer boundary: connectors publish capabilities, skills own deterministic workflow, and builder-owned agent config classifies each tool as denied, approval-required, or allowed. The document correctly records the fail-closed tri-state contract as a platform prerequisite rather than hiding policy in a bespoke connector. Approved.

Issue #1873, opened 2026-08-24
ADR-0117 claims a redaction path that does not cover ACI frames


jw3329:
ADR-0117's consequences say a snapshot's secrets "pass through the existing redaction path". That path does not cover the frames the snapshot travels on.

`runner/src/curie_runner/redact.py` scopes itself explicitly, in its own module docstring: the runner's stdout logs and the gen_ai span attributes, per #518. It then names what it deliberately leaves out — `server.py` "streams verbatim model output as NDJSON ACI frames, which is a larger surface left deliberately untouched here: the frames are the product contract, and scrubbing them is a separate decision."

That was defensible while a `side_effect_flag` carried a tool name and a constant string. As of #1869 it carries the call's `arguments` and the tool's `result`, and once the ledger fills (#1872) both land in Postgres and stay there. A snapshot of a Kubernetes object can carry environment variables and secret references; so can a tool's arguments.

What is already true and limits the blast radius:

- `summarize_tool_call` in `runner/src/curie_runner/approval.py` already renders a full tool input into an approval summary that is persisted, so tool arguments reaching the platform's database is not a new class of exposure — it is a wider one.
- ADR-0117 states the honest fallback: a resource that cannot be snapshotted safely "must report itself irreversible rather than storing credentials in the control plane". That puts the decision on the connector author, with nothing in the platform enforcing it.

Not blocking the ledger work. What needs deciding is whether scrubbing reaches the ACI frames at all — `redact.py` calls that "a separate decision" in as many words, so it may warrant its own ADR rather than being settled inside an implementation PR.

jw3329:
Draft ADR-0122 is up for review as #1919. Not merging it — ADRs go through review.

Writing it turned up something that changes the shape of the answer, so it is worth stating here too: **the obvious fix is wrong, and provably.**

Running the existing redaction rules over a snapshot works exactly as designed:

```
before  {"env": [{"name": "API_TOKEN", "value": "Bearer sk-live-abcdef123456"}]}
after   {"env": [{"name": "API_TOKEN", "value": "[REDACTED:bearer_token]"}]}
```

And that is the problem. A snapshot is not a log line — it is the **input to a write**. ADR-0121 replays it through the connector's restore verb, so a scrubbed snapshot means an undo sets the resource's token to the literal string `[REDACTED:bearer_token]` and reports success.

[...]

Issue #1867, opened 2026-08-24
Decide where the undo executor lives (Draft ADR needed)


jw3329:
Last slice of #1861, and the one decision ADR-0117 deliberately does not make. **This needs its own Draft ADR before any implementation** (ADR-0085 as amended by ADR-0102).

Nothing in the platform can reach a connector today: neither `apps/api` nor `apps/worker` has an MCP client, and every out-of-loop HTTP call either makes is to the platform's own API. So the API rules on an undo and returns, and something else has to perform the call the ruling authorizes. Until this lands, an undo is authorized and recorded but not executed.

Three candidates, from ADR-0117:

- **An MCP client in the worker.** The dependency is already present transitively, so this is the smallest change. It also makes the control plane a direct client of tenant connectors, which is a reachability question ADR-0008 has opinions about.
- **A new ACI verb, executed by the runner in a sandbox.** Architecturally the most conservative: the sandbox is where a connector's credential, allowlist and network policy already apply, so an undo could reach nothing the forward call could not. Also the most work — the runner would need to invoke an MCP tool outside the model loop, which a spike found it cannot do today, because MCP servers are handed to the SDK and the SDK owns those client sessions.
- **The connector exposes a plain HTTP replay endpoint.** Avoids MCP entirely and changes the connector contract for every author.

Whatever lands, the ruling stays separate from the execution: a refusal is recorded and returned before anything could act on it.

jw3329:
Draft ADR-0121 is up for review as #1893.

Correcting my own process error: I merged it as #1887 and reverted that in #1892. The procedure permits a Draft to merge for discussion, but a Draft merged without anyone reading it gets discussed by nobody, and then reads as settled to whoever finds it next. The decision it proposes — where an undo executes — is exactly the kind of call that exists to be argued with before it is on the record.

Content is unchanged. Nothing on this issue is implemented, and nothing will be until the ADR is Accepted.

jw3329:
ADR-0121이 `next`에 들어갔습니다 (#1893, `73d65d21`) — 리뷰 세 지적 반영본입니다.

**머지가 구현을 허가하지는 않습니다.** 파일의 상태 줄은 여전히 `Status: Draft`이고, ADR-0085(ADR-0102로 개정)에 따라 구현을 허가하는 것은 `Accepted`뿐입니다. Draft 머지는 논의를 위한 게시입니다.

[...]

PR #1893, opened 2026-08-25
Draft ADR-0121: a restore is the connector's own verb, run where the forward call ran

jw3329:
Refs #1867. **Draft — authorizes no implementation** (ADR-0085 as amended by ADR-0102). **Please review; I will not merge this.**

> Re-opened after I merged it myself in #1887 and reverted that in #1892. A Draft may merge for discussion, but one merged without anyone reading it gets discussed by nobody, and then reads as settled to whoever finds it next. Content unchanged from #1887.

ADR-0117 shipped its ruling half and named three candidates for where an undo is executed, deliberately choosing none. Building the ruling showed the three **conflate two questions with different answers**, and that the harder one was never asked.

> All three candidates describe **who calls the connector.** None answers **what call to make.**

A restore replays `{"spec": {"replicas": 3}}` onto a Deployment. Turning that into `scale_deployment(replicas=3)` means knowing that this snapshot's `spec.replicas` is that tool's `replicas` argument — per tool, per resource. That is the mapping DSL ADR-0117 rejects, arrived at from the other direction. The connector already holds that knowledge, because it produced the snapshot.

## What it proposes

1. **The connector restores its own resource** — a `restore` verb taking the recorded `target` and `prior_state` exactly as the ledger holds them. Same argument as ADR-0117 decision 1 applied to the return leg: the reply was the declaration because a manifest could disagree with what the tool returns; the restore is the connector's verb because a platform-derived call could disagree with what the tool does.
2. **The runner invokes it, in a sandbox, under the connector's own binding** — where its credential, allowlist and network policy already apply, so an undo can reach nothing the forward call could not.
3. **The executor confirms completion to the ledger**, closing the compromise the ruling shipped with: `undone_at` is set at authorization today because nothing reports back, so a restore that never runs leaves a record saying it did.
4. **A connector that reports a prior state and exposes no restore is a bundle-validation error** — not an undo-time surprise in front of someone who just pressed a button on a receipt that promised the action could be put back.

## The rejection worth arguing with

**An MCP client in the worker** is the smallest change and the ADR rejects it on decision 2: it would run the undo under the *control plane's* network reachability rather than the sandbox's. An undo that can reach something the forward call could not is a privilege escalation wearing a safety feature's clothes.

## The cost worth arguing with

The runner must invoke an MCP tool **outside the model loop, which it cannot do today** — the same thing ADR-0117's spike found. The ADR argues it's a smaller mechanism than it looked (one verb, fixed argument shape, no model in the path), but that's the claim to attack first.

Also flagged: the snapshot travels further than it does today, which sharpens the redaction gap in #1873 — `redact.py` scopes itself to stdout and span attributes and explicitly leaves ACI frames alone.


TheConnMan:
The connector-owned restore boundary is the right direction, but I see three architecture gaps to resolve before this ADR is accepted:

1. **Pin the replay contract.** The original sandbox may no longer exist, and the current connector may not understand a snapshot produced by an earlier version/schema. The ledger needs to pin connector identity/version (and snapshot schema), or the ADR needs an explicit compatibility and migration contract. Unless literal sandbox preservation is intended, describe restore as executing in a sandbox carrying the same pinned connector version and policy envelope—not necessarily the original sandbox.

2. **Make validation enforceable.** Whether a forward tool call returned prior state is a runtime fact. Bundle validation cannot enforce “prior state implies a restore verb” without a statically inspectable declaration/contract, or a deliberately broader rule that is statically checkable. Please choose and specify that contract.

3. **Resolve snapshot exposure before replay.** This depends on the redaction decision in #1873: ACI frames are outside the current redaction path, and returning a snapshot to a sandbox widens its exposure. The ADR needs to state the exposure/redaction guarantee, or explicitly make #1873 a prerequisite.

TheConnMan (changes requested):
Requesting changes for the three architecture gaps summarized in the top-level comment: pinned connector/snapshot compatibility, a statically enforceable restore contract, and the ACI snapshot-redaction prerequisite.

jw3329:
All three are right, and the first one caught the ADR asserting something that cannot be true. Pushed.

**1 — Pin the replay contract.** You are right that the original sandbox is gone: sandboxes do not survive their turn (which is why `WorkflowStateEntry` exists) and an undo can be pressed days later. "Invoked inside the sandbox the forward call ran in" was impossible as written — the title and filename said it too, and both now say what is actually required: **a sandbox carrying the same pinned connector image and the same policy envelope.** The envelope is named so it can be checked rather than assumed — the connector's own credential, its allowlist, and the sandbox network policy that reaches it.

For identity I used what already exists rather than inventing a field: **the resolved digest from `connectors.lock.yaml`.** ADR-0113 already makes that "the only identity rendered into a Deployment or used to start the local connector", so pinning it pins the snapshot's shape transitively. Deliberately **no separate snapshot-schema version** — it would be a second thing to keep in agreement with the first, and a digest mismatch already names the interesting failure (a connector upgraded under a stored snapshot).

And the compatibility contract is **refusal, not migration**. A snapshot reinterpreted by code that did not write it is the guess this whole design exists to avoid, so an unavailable digest refuses the undo with that as its stated reason. I added the cost to Consequences, because it surprised me: **undoability now has a shelf life set by deploys rather than by time**, which is not what ADR-0117's "a record does not expire" leads a reader to expect.

**2 — Make validation enforceable.** Correct, and the old decision 4 asked bundle validation to enforce a runtime fact it cannot see. Split in two:

- **static** — whether the connector *has* a restore verb is its advertised tool list. The checkable rule is deliberately broader than the runtime question: a connector advertising any non-read-only tool must advertise `restore`, or it advertises none and is treated as restoring nothing. `readOnlyHint` is already how this repo's own write-path gate finds write tools.
- **runtime** — whether a given call returned a prior state stays the reply's job.

This is not the manifest ADR-0117 rejected. That rejection was of a declaration that an action *is reversible*, which could disagree with what the tool returned; a tool list says which verbs exist and cannot disagree with itself. Both halves fail closed together, so **a connector cannot make its actions undoable by reporting a snapshot it has no verb to replay.**

**3 — Resolve snapshot exposure first.** Made #1873 an explicit prerequisite with its own section. Writing it out narrowed my own claim: the return leg adds **no new party** — same connector, same runner, same binding, all of which saw the value on the way out — but it adds **duration and repetition**, storing a snapshot indefinitely and re-transmitting it to a process that did not exist when it was captured. And the guarantee that would make that acceptable cannot be stated while `redact.py` scopes itself to stdout and span attributes and calls the ACI frames "a larger surface left deliberately untouched". ADR-0117's answer — the author reports the resource irreversible — asks every connector author to reason about a control plane they cannot see, which is not sufficient here.

One more cost I added while addressing 2: the broader capability rule gives an honestly-mixed connector no way to say "this verb is reversible and that one is not" short of splitting servers — which is what `k8s-write` and `k8s-scale` already do, so the pattern is established but the cost is real.

TheConnMan (approved):

PR #1910, opened 2026-08-25
Give the kubeconfig leak check a needle that cannot occur by chance

jw3329:
`reader_kubeconfig_is_owned_secret_stdin_and_connectors_reconcile_after_deploy` failed on a **docs-only PR** (#1893). It is a security check — the generated reader token must not reach stdout or argv — and it failed with no credential anywhere near the output.

## What actually happened

The fixture token was the three-character string `"abc"`, and `observable` is stdout plus the recorded kubectl and helm argv, which carry temp paths built from random UUIDs. **UUIDs are hex, so `a`, `b` and `c` are all in their alphabet.** The failing run had rendered its values file to

```
/tmp/curie-helm-values-0e7b51b5-ab30-4abc-a064-7cee3121be61.yaml
                                    ^^^^
```

and the leak check fired on a filename.

## The flakiness is the smaller half

A three-character needle cannot distinguish *"the credential leaked"* from *"a UUID contained abc"*. So even on the runs where it passed, it was not asserting much. And a leak check that fires spuriously gets re-run until green — which is how it stops checking anything at all.

## The fix

The token becomes `curie-test-reader-token-must-not-leak` — in the Secret fixture, the equality assertion, and both needles (raw and base64). Long enough not to collide, and **containing characters no UUID can produce**, so the property is structural rather than probabilistic.

## Verified both directions

Not just green. A needle nobody checks could be made unfindable by the same edit that makes it stable, so the more important half is the first one:

```
token spliced into observable   →  1 failed   (the assertion still has teeth)
unmodified                      →  26 passed
cargo fmt / clippy              →  clean
```

Test-only. No product code.

### DIFF
--- a/cli/tests/example_sre_bot_install.rs
+++ b/cli/tests/example_sre_bot_install.rs
@@ -171,7 +171,7 @@ case " $* " in
     *" get secret sre-bot-reader-token "*)
         case "$CURIE_TEST_READER_TOKEN_MODE" in
             success)
-                printf '%s\n' '{"apiVersion":"v1","kind":"Secret","data":{"ca.crt":"Zml4dHVyZS1jYQ==","token":"YWJj"}}'
+                printf '%s\n' '{"apiVersion":"v1","kind":"Secret","data":{"ca.crt":"Zml4dHVyZS1jYQ==","token":"enp6enp6enp6enp6"}}'
                 exit 0
                 ;;
             read-failure)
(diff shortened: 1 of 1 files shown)


jw3329:
Corrected: my first fix traded one problem for the other, and gitleaks caught it.

The long distinctive token fixed the UUID collision and became a `generic-api-key` finding — a 52-character base64 blob with entropy 4.57, sitting immediately after the word `"token"` in a Kubernetes Secret fixture. **That is the secret scanner working correctly**, and `AGENTS.md` explicitly says not to silence a finding with `gitleaks:allow` because it hides the next real one.

Both constraints, measured rather than guessed:

```
"abc"                    base64 entropy 2.00   a UUID can produce it     ← the original bug
"curie-test-reader-…"    base64 entropy 4.57   a UUID cannot             ← my first fix
"zzzzzzzzzzzz"           base64 entropy 2.00   a UUID cannot             ← now
```

`z` is outside hex, so no UUID or digest can produce the needle, and the repeated block keeps the base64 form's entropy exactly where `"abc"` had it. No suppression anywhere, and the value reads as the placeholder it is rather than as a plausible credential.

Re-verified both directions after the change: spliced into `observable` it fails, unmodified the suite passes 26/26.

PR #1916, opened 2026-08-25
Let a stored binding serialize without re-passing the write rule

### LINKED ISSUE #1914: One legacy agent_channels row makes GET /agents fail for every agent
`GET /agents` returns 500 for **every** agent when a single `agent_channels` row holds an address the current write rule would reject. One legacy row takes the whole endpoint down.

Observed on a local stack running `next`:

```
$ curl -H "X-API-Key: $KEY" http://localhost:28000/agents
Internal Server Error

pydantic_core._pydantic_core.ValidationError: 1 validation error for AgentOut
channels.0
  Value error, slack channel 'C-6a9274f5' is not a Slack channel ID: ...
```

19 agents existed; 18 had addresses of the form `C-<hex>`. Deleting those rows made the endpoint 200 again — the one valid agent had been unreachable through the API purely because of its neighbours.

## Why a write rule runs on reads

`ChannelBinding` (`apps/api/src/curie_api/schemas.py:645`) carries a `model_validator(mode="after")` that calls `_validate_channel_binding(kind, address)`, and the same class is the element type of `AgentOut.channels` (`:889`). So the shape rule that guards a *bind* also runs when an existing row is *serialized*.

The file is already alert to this class of mistake — `ChannelBindingWrite` was split out precisely so write-only fields would not "422 valid reads and leak a write rule into a read contract". The address *shape* rule stayed on the shared model and is the one that leaked.

## There is a production path to such a row, and it is an upgrade

Migration `0021_agent_channels.py:137` backfills the new table from the old column with no validation at all:

```sql
INSERT INTO curie.agent_channels (id, agent_id, kind, address)
SELECT gen_random_uuid(), id, 'slack', slack_channel FROM curie.agents
```

`agents.slack_channel` is exactly where #143's bad values lived — a literal `#name` or bare name stored before the validator existed, which "reported success and never routed". Any install carrying one of those upgrades cleanly, and then every read of `/agents` fails.

So the failure mode is: an install that was merely *mis-routed* before the upgrade becomes an install whose agent listing is *entirely unavailable* after it, with an error naming Pydantic rather than the data.

## Suggested shape of a fix (not decided here)

Serialization of a stored row should not re-litigate whether the row should have been stored. Options, in rough order of how much they change:

1. Read-side model without the shape validator (mirror of the existing write/read split), so a bad row serializes and stays visible — arguably *more* useful, since an operator can see the wrong value and fix it.
2. Per-row degradation: surface the offending binding with a flag rather than failing the whole response.
3. A migration that quarantines or normalizes addresses the current rule rejects, so the data matches the rule.

Option 1 plus 3 seems right, but the read contract is ADR-0096/ADR-0118 territory and the choice belongs with whoever owns it. What should not survive is a single row deciding whether the endpoint answers.

Found while verifying the ADR-0117 undo path end to end (#1861).

jw3329:
Closes #1914.

One `agent_channels` row holding an address the current rule rejects made `GET /agents` fail for **every** agent. Nineteen agents, one bad neighbour, and the whole listing answered 500 with a Pydantic error rather than a value.

## Why a write rule ran on reads

`ChannelBinding` carries the address-shape rule from #143, and **three write paths inherit it** — `ChannelBindingWrite`, `ChannelTokenRequest`, `TurnIn`. It was *also* the element type of `AgentOut.channels`, so the rule that guards a bind ran again every time an existing row was serialized.

The file was already alert to this class of mistake: `ChannelBindingWrite` was split out precisely so write-only fields would not *"422 valid reads and leak a write rule into a read contract"*. The address **shape** rule stayed on the shared model, and it's the one that leaked.

## There is a production path, and it is an upgrade

`0021_agent_channels.py` backfills with no validation:

```sql
INSERT INTO curie.agent_channels (id, agent_id, kind, address)
SELECT gen_random_uuid(), id, 'slack', slack_channel FROM curie.agents
```

`agents.slack_channel` is exactly where a literal `#name` from before the validator lived — the value #143 describes as *reporting success and never routing*. So an install that was merely **mis-routed** before the upgrade became an install whose agent list is **entirely unavailable** after it.

## The fix

Reads get their own model. `ChannelBindingOut` is **deliberately not a subclass** — inheriting from `ChannelBinding` is what caused this.

Writes keep the rule untouched, and that half is tested too: a fix that made reads tolerant by making writes tolerant would re-open #143.

| test | asserts |
| --- | --- |
| `test_a_legacy_binding_serializes_instead_of_failing` | the bad value is **shown**, not hidden behind a 500 |
| `test_one_bad_row_does_not_take_the_listing_down` | a neighbour's row no longer hides other agents |
| `test_binding_a_bad_address_is_still_refused` | create still 422s (#143 intact) |
| `test_rebinding_to_a_bad_address_is_still_refused` | `PATCH /agents/{id}/channels` still 422s |

Showing the bad address is also the better outcome — an operator cannot fix a value the API refuses to tell them.

## Not a wire change

The shape stays `{kind, address}`, exactly what `ChannelBinding` serialized, so no generated client sees a different response. Only the component *name* moved — which the **CLI field-parity gate caught**, working as designed. `api-mirrors.json` now names `ChannelBindingOut`, and the Rust mirror documents why it deserializes the read shape.

## Verification

```
apps/api                        1056 passed, 10 skipped
  the four above                   4 passed  (real Postgres, legacy row written past the API)
cli field-parity / emit-parity   clean
cli api_deploy                  16 passed
ruff / mypy                     clean (229 source files)
check-contracts / check-docs    clean; openapi regenerated
```

Two pre-existing failures, both confirmed identical on a clean `next` checkout: `test_cost_composes_metrics_for_the_agent` (needs Langfuse) and `chart_check_passes_on_the_unmodified_chart`.

### DIFF
--- a/apps/api/openapi.json
+++ b/apps/api/openapi.json
@@ -647,7 +647,7 @@
           },
           "channels": {
             "items": {
-              "$ref": "#/components/schemas/ChannelBinding"
+              "$ref": "#/components/schemas/ChannelBindingOut"
             },
             "title": "Channels",
             "type": "array"
@@ -1530,9 +1530,9 @@
         "title": "BundleOut",
         "type": "object"
       },
-      "ChannelBinding": {
+      "ChannelBindingOut": {
         "additionalProperties": false,
-        "description": "Where one agent listens: a channel KIND and an ADDRESS (ADR-0096, #1459).\n\nAn agent holds ONE OR MORE of these (ADR-0118, #1525, amending ADR-0089's\nsingular clause). `AgentCreate` still carries one `channel` OBJECT -- the\nfirst binding, required, since an agent with none cannot receive a turn --\n`AgentOut` carries the `channels` LIST, and every write after the create\ngoes through the `/agents/{id}/channels` subresource. `AgentUpdate` carries\nno binding key at all.\n\n`kind` names the adapter that owns the binding, selects the address-shape\ncheck, AND routes: since ADR-0096 phase 2 the worker resolves on the\n`(kind, address)` PAIR, so the two together are the routing key the worker\nmatches on equality (a Slack channel id for `kind=\"slack\"`). One address can\ntherefore be bound twice under two different kinds.",
+        "description": "The READ side of a binding: the stored pair, serialized as it is stored.\n\nDeliberately NOT a subclass of `ChannelBinding`, and that is the whole point.\n`ChannelBinding` carries the address-shape rule three write paths inherit\n(`ChannelBindingWrite`, `ChannelTokenRequest`, `TurnIn`), and it used to be\nthe element type of `AgentOut.channels` as well -- so the rule that guards a\nBIND also ran when an existing row was READ, and one row it rejected failed\nthe whole response for every agent in it (#1914).\n\nAn install reaches that state by upgrading: migration 0021 backfills\n`agent_channels.address` from `agents.slack_channel` verbatim, and that column\nis exactly where a literal `#name` from before the validator lived. So an\ninstall that was merely mis-routed became one whose agent list was\nunavailable, reporting a Pydantic error instead of the bad value.\n\nSerializing a stored row must not re-litigate whether it should have been\nstored. Showing the bad address is also the more useful outcome: an operator\ncannot fix a value the API refuses to tell them.\n\nThe shape stays `{kind, address}`, identical to what `ChannelBinding`\nserialized, so this is not a wire change -- `ChannelBindingWrite`'s docstring\nalready describes that as the read contract.",
         "properties": {
           "address": {
             "title": "Address",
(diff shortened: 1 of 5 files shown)


jw3329:
Fixed the gitleaks failure. It was this repo's own `slack-conversation-id` rule, and it was right to fire.

My fixtures used `C0VALID001`/`C0VALID002`/`C0VALID003` — id-shaped strings I invented. `.gitleaks.toml` rejects those because a real channel id committed as a test fixture is a disclosure that cannot be rotated away, and it allowlists placeholders instead. Existing tests already use `C0EXAMPLE1`/`C0EXAMPLE2`; mine now use `C0EXAMPLE7`–`C0EXAMPLE9`, with a comment saying why the form is not free choice.

Landed as **one commit**, for the same reason the values changed: `gitleaks (full history)` scans every commit, so a discarded attempt fails the branch from its own past. (Same lesson as #1910 — worth knowing twice.)

PR #1917, opened 2026-08-25
Run the local stack on the checkout it was started from

### LINKED ISSUE #1915: The dev loop cannot run the local stack from a source checkout: CLI, images, and tree drift apart
`curie local message` cannot run against a `next` checkout on a machine that follows the documented dev loop. The CLI, the compose images, and the source tree drift apart, and nothing brings them into agreement.

## What happens

Two independent failures, both from the same cause.

**The CLI and the API disagree on a field name.** `curie local up` starts `ghcr.io/curie-eng/curie-api:${CURIE_BASE_TAG:-latest}` — a *published* image. The on-PATH `curie` is built from the source checkout by `curie update`. So:

```
$ curie local message "..."
Error: listing agents via http://localhost:28000 (is `curie local up` running?)
· decoding agent list: error decoding response body: missing field `channel` at line 1 column 316
```

The CLI expects `channel` (singular); the API returns `channels` (plural, ADR-0118). Both are internally correct — they are just from different commits.

**The dispatcher image predates the CLI that drives it.**

```
$ curie local message "..."
Container curie-dispatcher-enqueue-... Created
/app/.venv/bin/python: No module named curie_dispatcher.enqueue_once
```

Same shape: the CLI invokes a module the published dispatcher image does not have yet.

## Why the documented loop does not prevent it

`CLAUDE.md` names `curie update` as "the per-change loop, so you never re-run the bootstrap script". Its own help says what it covers:

> Rebuild this CLI from the source checkout and reinstall it on PATH. Pass `--image` to also rebuild the local runner image.

**The CLI and the runner. Not the api, worker, or dispatcher images**, which are four `ghcr.io/curie-eng/*` references in `compose.dev.yaml`. So a contributor working on `next` gets a source-built CLI talking to whatever the registry last published, and the gap widens with every merge that touches an API response shape or a dispatcher entrypoint.

`CURIE_BASE_TAG` is the existing escape hatch, but using it requires knowing to build three images with the right names and tags first:

```
docker build -t ghcr.io/curie-eng/curie-api:dev    -f apps/api/Dockerfile .
docker build -t ghcr.io/curie-eng/curie-worker:dev -f apps/worker/Dockerfile .
docker build -t ghcr.io/curie-eng/curie-dispatcher:dev -f apps/dispatcher/Dockerfile .
CURIE_BASE_TAG=dev curie local up --minimal
```

That is the "copy-pasted `docker build`" that `CLAUDE.md`'s one-entry-point rule exists to remove.

## What I would suggest

A single verb that makes the local stack match the checkout — `curie update --stack`, or `curie local up --build`, whichever fits the existing surface better. It should build every image `compose.dev.yaml` names from source, tag them consistently, and run with that tag, so the answer to "am I testing my code?" is not "depends which process you ask".

Worth considering alongside it: `curie local up` could compare the CLI's own version against `GET /health` (or the API's version) and refuse with a clear message instead of surfacing a serde error about a missing field. The field-parity gate (`curie dev field-parity`) already exists to keep these two in agreement *within* a commit; nothing checks it *across* processes at runtime.

Found while trying to verify the ADR-0117 undo path end to end (#1861) on a stack built from `next`.

jw3329:
Closes #1915.

`curie local message` could not work against a `next` checkout. `curie update` refreshes the CLI and the runner image; the stack's api, worker, dispatcher and ui came from the **registry**, so a source-built CLI talked to whatever was last published. Neither failure looked like version skew — `missing field 'channel'` from a serde decode, and `No module named curie_dispatcher.enqueue_once` from inside a container.

`curie local up --build` builds what the run needs from this checkout and runs it.

## Five things had to be true, and testing found four of them

I would not have reasoned my way to any of the last four. Each was found by driving the loop and watching it still fail.

**1 — The override has to reach every image.** `CURIE_BASE_TAG` already drove api, migrate and worker; **ui and dispatcher were pinned to `latest`**, so one variable moved three services and silently left two on the registry. A test now derives the overridable set from `compose.dev.yaml` *and* the worker overlay's own `FROM`, and asserts the build set covers it — so a service added in either style cannot be forgotten.

**2 — The dispatcher is built even without `--slack`.** The profile governs the long-running service, but `local message` runs a **one-shot container from that same image on every invocation**, outside any profile. Keying the build set on profiles left the main dev-loop verb still failing.

**3 — The one-shot producer has to inherit the *running* stack's tag.** Compose substitutes `${CURIE_BASE_TAG:-latest}` from each invocation's own environment, so after `up --build` the stack ran `:dev` and `local message` pulled `:latest`. The tag is now **read off the running API container** rather than remembered — and off the API, not the worker, because the worker is a compose-built overlay whose `.Config.Image` carries no tag at all.

**4 — The runner counts, and it decides whether the *agent* runs your code.** The worker spawns it per turn from `CURIE_RUNNER_IMAGE`, also hardcoded to `latest`. With every platform service rebuilt, a turn still produced the **old fake model's frames** because the sandbox ran the registry's runner.

**5 — Rebuilding the base images is not enough.** `curie-worker` is a compose-built overlay over the published base, and without `--build` on the compose child, compose reuses the overlay it baked over the **previous** base. The stack reported itself source-built while running yesterday's worker on today's base, and the only visible symptom was a real turn recording no `post_state` — because the code that extracts it was in the layer that never got rebuilt.

## Verified by driving the loop

```
curie local deploy   ok      (the field mismatch is gone)
curie local message  ok      "scaled public/api from 3 to 10 — can be undone"
the ledger row       prior_state {"spec":{"replicas":3}}
                     post_state  {"spec":{"replicas":10}}
                     target      {"kind":"Deployment","namespace":"public","name":"api"}
worker POST_KEY      post        (was MISSING before fix 5)
CURIE_RUNNER_IMAGE   ...curie-runner:dev   (was :latest before fix 4)
```

## A plain `local up` is unchanged

No builds, no `--build` on the compose child, no tag override — a non-contributor keeps the fast restart and the published images. Asserted by `without_build_the_tag_is_left_to_compose` and `a_published_stack_leaves_the_tag_to_compose`.

```
cli --lib             49 local:: + 68 message:: passed
check-contracts       clean
check-docs            clean
cargo fmt / clippy    clean
```

`chart_check_passes_on_the_unmodified_chart` fails here and on a clean `next` checkout alike — pre-existing, unrelated.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -12,6 +12,7 @@ stay the *implementation*; it just isn't the interface.
 - Building the runner image → `curie build` (not a copy-pasted `docker build -f runner/Dockerfile ...`).
 - First-run dev bootstrap → `curie install` (or `./get-curie.sh` from a source checkout, which also puts `curie` on PATH the first time by building it).
 - Refresh the on-PATH CLI after a code change → `curie update` (rebuilds and `cargo install`s the CLI to `~/.cargo/bin`; `--image` also rebuilds the runner). The per-change loop, so you never re-run the bootstrap script.
+- Run the local stack on YOUR checkout → `curie local up --build`. `curie update` covers the CLI and the runner image; the stack's api, worker, dispatcher and ui otherwise come from the registry, so a source-built CLI ends up talking to whatever was last published. That skew does not announce itself — it surfaces as a serde error about a field name, or a missing Python module from inside a container (#1915).
 - Contributor/CI scripts (contract codegen, chart render-asserts, the e2e round-trip) → `curie dev <...>`. The `dev` namespace fences off commands that need a **source checkout + dev toolchains**; they error clearly when run from a released binary.
 - Operator/product commands stay top-level: `init`, `build`, `skill`, `local`, `cluster`.
 
--- a/apps/ui/src/generated/commandManifest.ts
+++ b/apps/ui/src/generated/commandManifest.ts
@@ -705,6 +705,18 @@ export const commandManifest = {
               "long": "env-file",
               "positional": false,
               "required": false
+            },
+            {
+              "global": false,
+              "help": "Build the stack's images from THIS checkout instead of pulling the published ones, and run them (#1915)",
+              "id": "build",
+              "long": "build",
+              "positional": false,
+              "possible_values": [
+                "true",
+                "false"
+              ],
+              "required": false
             }
           ],
           "hidden": false,
--- a/cli/command-manifest.json
+++ b/cli/command-manifest.json
@@ -702,6 +702,18 @@
               "long": "env-file",
               "positional": false,
               "required": false
+            },
+            {
+              "global": false,
+              "help": "Build the stack's images from THIS checkout instead of pulling the published ones, and run them (#1915)",
+              "id": "build",
+              "long": "build",
+              "positional": false,
+              "possible_values": [
+                "true",
+                "false"
+              ],
+              "required": false
             }
           ],
           "hidden": false,
--- a/cli/src/commands.rs
+++ b/cli/src/commands.rs
@@ -1266,7 +1266,11 @@ fn on_path(bin: &str) -> bool {
 
 /// Walk up from the current directory to the repo root: the nearest ancestor
 /// that contains `runner/Dockerfile`.
-fn find_repo_root() -> Option<PathBuf> {
+///
+/// `pub(crate)` for `local::build_source_images` (#1915), which builds the dev
+/// stack's images from the same root and wants the same "are we in a checkout"
+/// answer rather than a second walk with its own anchor file.
+pub(crate) fn find_repo_root() -> Option<PathBuf> {
     let mut dir = std::env::current_dir().ok()?;
     loop {
         if dir.join("runner/Dockerfile").is_file() {
--- a/cli/src/comms.rs
+++ b/cli/src/comms.rs
@@ -795,6 +795,7 @@ mod tests {
                 slack: false,
                 model_mode: mode,
                 env_file: None,
+                build: false,
             })
             .env;
             let connect_env = local_connect_commands(&local_comms_opts(false, mode))[0]
@@ -831,6 +832,7 @@ mod tests {
                 slack: false,
                 model_mode: mode,
                 env_file: None,
+                build: false,
             })
             .env;
             let disconnect_env = local_disconnect_commands(&local_comms_opts(true, mode))[1]
@@ -865,6 +867,7 @@ mod tests {
                 slack: false,
                 model_mode: ModelMode::DefaultFake,
                 env_file: None,
+                build: false,
             })
             .env;
             let connect_env = local_connect_commands(&local_comms_opts_with(
@@ -903,6 +906,7 @@ mod tests {
                 slack: false,
                 model_mode: ModelMode::DefaultFake,
                 env_file: None,
+                build: false,
             })
             .env;
             let disconnect_env = local_disconnect_commands(&local_comms_opts_with(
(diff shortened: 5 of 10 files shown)


jw3329:
CI caught a real regression, and reasoning would not have.

Making `CURIE_RUNNER_IMAGE` read `CURIE_BASE_TAG` broke the e2e ladder: **CI sets `CURIE_BASE_TAG=ci-local`** for the published platform images while building its runner as a plain `curie-runner`, so the worker went looking for a `curie-runner:ci-local` nothing had built.

The mistake was conceptual, not a typo. The runner **is not a compose service** — the worker spawns it per turn — and it is built and tagged on a different axis from api/worker/dispatcher/ui. One variable meaning *"which published platform images"* cannot also mean *"which runner the sandbox runs"*.

Compose now reads `${CURIE_RUNNER_IMAGE:-ghcr.io/curie-eng/curie-runner:latest}` — byte-identical behaviour to the old hardcoded line when nothing sets it — and `--build` names the image outright. A caller that overrides the platform tag keeps whatever runner it built, which is what the ladder relied on.

The invariant test grew a third form to scan (`image:` lines, the worker overlay's build ARG, and now `CURIE_RUNNER_IMAGE`). **That test has caught two of my own omissions now**, which is why it is written as a derivation from the compose files rather than a hand-maintained list.

Re-verified the loop after the change: worker reports `CURIE_RUNNER_IMAGE=...curie-runner:dev`, `local message` completes, and the turn prints "scaled public/api from 3 to 10 — can be undone".

jw3329:
CI caught **two more instances of one mistake**, and the mistake was mine about what `CURIE_BASE_TAG` means.

It does not mean "every image the stack runs". The ladder sets it to `ci-local` for the images it builds — the api and the worker's base — and builds its dispatcher **separately as `:latest`**. Widening the variable's reach made compose pull a `curie-ui:ci-local` and a `curie-dispatcher:ci-local` that nothing had built: `manifest unknown`, whole stack down.

So ui, dispatcher and runner each read **their own variable**, defaulting to the `latest` those lines used to hardcode. A caller that sets only `CURIE_BASE_TAG` now gets byte-identical behaviour to before this branch — which is what CI relied on — and `--build` sets each one outright.

The one-shot dispatcher needed the same correction **plus one more thing**. Deriving its image from the running stack's tag is right; but a tag that exists for the platform images need not exist for the dispatcher, which is exactly CI's shape. It now derives the candidate **and checks the image is present**, falling back to compose's default when it isn't. Both halves are load-bearing, and there's a test for each.

Also regenerated `cli/command-manifest.json` and the UI's mirror — a new clap flag is a grammar change, and `command_manifest_matches_committed_artifact` is the gate for it. `check-contracts` does not cover that, which is why CI had to tell me.

Worth noting: the build-set invariant test has now caught **three** of my own omissions. That is the argument for deriving the set from the compose files rather than restating it as a list.

Re-verified: `local up --build` then `local message` prints "scaled public/api from 3 to 10 — can be undone", and no `dispatcher:ci-local` exists locally, which is the condition that broke the ladder.

jw3329:
Fourth one, and this gate is worth naming.

The release compose generator pins every `ghcr.io/curie-eng/curie-*` tag, matching a plain `:latest` or `:${CURIE_BASE_TAG:-latest}`. My overrides are a **third shape** — the whole reference is a defaulted variable, `${CURIE_UI_IMAGE:-ghcr.io/curie-eng/curie-ui:latest}` — and T3's pattern matches the ref *inside* the braces.

So it pinned the tag and **left the wrapper**: the release asset would have shipped `image: ${CURIE_UI_IMAGE:-ghcr.io/curie-eng/curie-ui:0.7.0}`, and that asset has no shell to resolve it. Compose would try to pull an image literally named `${...}`.

Unwrapping first collapses the form to the plain ref T3 then pins — the same outcome the dev file's own unset default already produces. Verified on the generated asset: every curie image is a literal `:0.7.0`, and zero shell-variable image forms survive.

What I want to flag is **which** gate caught it: a Python test over the release generator (`compose/tests`). Nothing in the CLI, the contract gates, or check-docs would have. A compose edit made for a purely local-dev reason reaches the **release artifact**, and that test is the only thing standing between the two — I would not have thought to look there.

("E2E required" is the aggregate check that follows the ladder; not a separate cause.)

PR #1919, opened 2026-08-25
Draft ADR-0124: a snapshot is sealed to the connector that wrote it

jw3329:
Refs #1873. **Draft — authorizes no implementation** (ADR-0085 as amended by ADR-0102).

Renumbered to ADR-0124 so PR #1911 remains ADR-0122 and PR #1924 remains ADR-0123.

## Why this decision exists

ADR-0117 says snapshot secrets pass through the existing redaction path. They do not: that policy covers stdout and `gen_ai_span`, while `prior_state` rides a verbatim ACI frame and is later replayed as the input to a restore write. Scrubbing it would replace the real value with a redaction marker, silently corrupt the restored resource, and still allow the receipt to report success.

## What the measured spike established

The first draft proposed a cluster-held key. The `spike/adr-0121-restore-executor` probes showed that this does not compose with ADR-0121: the runner deliberately has no private sealing key, while the connector executing the restore is the only party that needs to open the snapshot. Moving the executor, opening in the worker, or giving the sandbox the cluster key each broke an existing boundary.

`probe_sealed_roundtrip.py` exercised the real API and Postgres: the ledger stored an opaque blob, the version-token conflict check refused `v2` versus `v3`, and an authorized undo returned the blob for the connector to open without the API reading it.

## Settled contract

- Every replayable or undoable `prior_state` is sealed by the connector that produced it. There is no cleartext exception based on author judgment.
- Under ADR-0086 and ADR-0009 ownership, the connector declares an operator-supplied durable sealing key through the existing named connector-secret or `SecretRef` seam. The value is delivered only to the hosted connector—never the API, worker, runner, ledger, or bundle.
- The ledger stores an opaque envelope containing a non-secret key identifier and ciphertext.
- Rotation retains old decryption keys while corresponding records remain undoable. A missing or retired key explicitly refuses restore with no fallback.
- A connector without a usable configured key is irreversible and cannot produce an undoable record.
- The version-token conflict check and receipt-only summary remain. Scrubbing `arguments` and `result` is a separate decision.

This revision changes only the ADR and generated ADR index; it does not change `packages/aci-protocol` or `packages/plugin-format`.


jw3329:
Checking my own claim before you spend time on it: **decision 1's "reuse the sealing that already exists" is not a straight reuse, and may not be a reuse at all.**

The sealed-credential seam runs **CLI seals → worker opens** (`cli/src/sealing.rs` → `apps/worker/src/curie_worker/sealing.py`), and its purpose is a bundle carrying a credential *encrypted to the cluster that will run it* (ADR-0094). The blob is authored outside and opened inside.

What this ADR needs is the opposite direction: the **platform** sealing a snapshot at runtime. There is no platform-side sealer — the sealer is the Rust CLI — and the private half lives with the opener. So sealing `prior_state` means either a new sealer in the API/worker, or handing the platform the public half of a key whose flow was designed to point the other way.

The wire format is reusable; the *direction* is not. That makes decision 1 heavier than I wrote it, and it deserves the same skepticism as ADR-0121's runner cost.

Also worth pairing: I re-confirmed ADR-0121's load-bearing unknown while looking. `runner/src/curie_runner/adapter.py` hands `mcp_servers` to the SDK, so the SDK owns those sessions — which is exactly what ADR-0117's spike found and why "the runner invokes it" is the expensive half of that proposal.

Both are spike questions rather than review questions, and I would rather answer them with something that runs than argue them on paper.

jw3329:
Spiking the executor before arguing it further, and the first probe **falsifies the cost this ADR inherited**.

ADR-0117's spike concluded the runner cannot invoke an MCP tool outside the model loop without "building a second MCP client inside the runner, which is a large new mechanism for one job". ADR-0121 carries that forward as "the most work".

A hosted connector's MCP entry is `{"type": "http", "url": "http://<service>:<port>/mcp"}` — **streamable HTTP, not a stdio subprocess the SDK spawned.** So the question was never "take a session from the SDK", it was "how big is an HTTP client for one call".

Measured against the real `k8s-scale` connector, started locally and called with no agent SDK in the process:

```
tools/list -> ['scale_deployment']
tools/call -> is_error=False
reply keys -> ['ok', 'post', 'prior', 'summary', 'target']
```

The client is **13 lines**:

```python
async with streamablehttp_client(url) as (read, write, _):
    async with ClientSession(read, write) as session:
        await session.initialize()
        result = await session.call_tool(tool, arguments)
```

This *strengthens* ADR-0121's proposal rather than weakening it — its chosen option was the expensive one and is not. It also sharpens the argument against the alternative: the worker could do this just as cheaply, so the case for the sandbox is now **purely about reachability and policy envelope**, not about cost. That is a cleaner argument than the one the ADR currently makes, and the Consequences should stop describing this as the heavy option.

Continuing the spike through the full round trip; will report what else it falsifies.

jw3329:
Spiked the whole loop rather than arguing it on paper. Branch: `spike/adr-0121-restore-executor` (probes only, nothing to land).

**The loop works, and the central claim holds.** Real API over HTTP, real Postgres, real connector:

```
world 3 → agent scales to 10 → ledger records prior={"spec":{"replicas":3}}
refused undo (world moved)   → 409, and the connector is never called: 10 → 10
authorized undo              → API returns {target, prior_state}
executor replays             → connector: "restored public/api from 10 to 3"
world 3
```

The executor hands the connector back **its own words**. No mapping table exists anywhere — which is this ADR's reason to exist, and it survives contact.

**But the two drafts do not compose, and that is the finding.**

Sealing is `crypto_box_seal`, so it needs only the public half — the worker can seal to itself. Opening needs the private key:

| | holds the sealing key |
| --- | --- |
| worker | yes (`CURIE_SEALING_PRIVATE_KEY`) — can open, and can seal |
| runner | **no**, by design: it is the sandbox |
| api | no |

ADR-0121 puts the executor in the **runner**. ADR-0122 seals `prior_state` to the **cluster key**. A sealed snapshot the runner cannot open is a restore the runner cannot perform.

Four ways out, none of them free:

1. the executor moves to the worker — which ADR-0121 rejects on reachability
2. the worker opens and hands plaintext to the runner — spends the seal on the hop it was protecting
3. the sandbox gets a key — which is what the sandbox exists not to have
4. `prior_state` is not sealed, and its exposure is answered another way

I do not think an implementation PR should pick. It is the kind of choice these two ADRs exist to make, and it is better made now that the constraint is measured rather than assumed.

**Also confirmed missing:** `POST /actions/{id}/confirm-undo` 404s. Until it exists, `undone_at` is stamped at ruling time and a restore that never runs leaves a record saying it did.

jw3329:
**Revised, and the revision came from spiking it rather than from arguing it.** Also retitled — the old title named the wrong mechanism.

## What I decided, and why it is none of the four ways out

The four all argue about **which platform component holds the key**. That is the wrong question. Exactly one party ever needs to read a snapshot: **the connector that produced it.** Not the API, not the worker, not the runner.

So the connector seals to its own key. Measured, against the real API and real Postgres:

```
prior_sealed : oXQXOYtoi8OkdJ8px94XTl8s9qWWbMqaKnCHU+bn...
platform can read the replica count from it: False
ledger row   : undoable=True, and prior_state is opaque
conflict     : left=v2 observed=v3  → HTTP 409, refused
authorized   : HTTP 200; the API hands back a blob it never read
connector    : "restored public/api from 10 to 3"
```

| | |
| --- | --- |
| **The tension disappears** rather than being traded | the runner needs no key because it never opens anything, so ADR-0121 keeps its executor *and* its reachability argument |
| **Stronger than what I first proposed** | cluster-sealed, the platform holds the key and an operator has to evaluate that; connector-sealed, there is nothing to evaluate |
| **It is decision 1 of ADR-0117 followed through** | the tool's reply is the declaration — now the tool also decides what the platform may read |

## The one thing it breaks

A sealed box is non-deterministic, so ciphertexts cannot be compared and ADR-0117's state-based conflict check cannot survive sealing.

The comparison moves onto a **version token** the connector reports — `resourceVersion`, ETag, generation. And this is a *better* question than state equality, not merely a compatible one: it catches a change that reverted to the same value, which a state comparison silently permits, and it does not false-positive on fields that churn on their own.

**The cost is real and is in the consequences, not glossed:** a refusal names `v2 vs v3` instead of `10 vs 7`. ADR-0117 explicitly valued an operator seeing their own fix named, and this spends some of that.

## What I am least sure about

**The keypair burden.** Every write connector that snapshots sensitive state now needs a durable key. I made it optional — a connector whose snapshot is `{"spec": {"replicas": 3}}` may report it in the clear and keep the richer refusal — but "optional" means an author decides, and deciding wrong is a credential in the control plane. That is the same criticism I levelled at ADR-0117's escape hatch, and I have not fully escaped it.

Probes: `spike/adr-0121-restore-executor`.

TheConnMan (approved):
Maintainer approval after direct revision: Draft ADR-0124 now makes connector sealing mandatory for every undoable prior_state, pins key custody and rotation/refusal behavior, preserves the measured spike evidence, and remains docs-only with docs lint passing.

PR #1921, opened 2026-08-25
Moonshot ADR-0115: implement agent-to-agent delegate calls

alexrao2000:
## Ships ahead of ADR-0115's Accepted status

**Read this first.** [ADR-0115](https://github.com/curie-eng/curie/blob/main/docs/adr/0115-agents-call-each-other-with-no-third-party.md) is still `Status: Draft` on `main` -- nobody has flipped it to Accepted, and per this repo's own rule ("a Draft may merge for discussion but cannot authorize implementation") that would normally block this PR outright. It's here anyway on the repo owner's explicit direction, given knowingly. Flagging it plainly rather than quietly, so reviewers see the process gap instead of assuming it was followed.

## What this is

The real implementation, built on top of the demo prototype this PR used to be (see prior history for the "demo, not for merge" version). Closes the gaps the prototype's own notes documented, using the mechanisms ADR-0115's Decision section and TheConnMan's maintainer-decision comments on #1793 actually specify:

- **Real depth/cycle bound (part 6).** Replaces the prototype's "refuse if the caller's own conversation id starts with `delegate:`" hack with an actual chain: each call's `chain`/`depth` is looked up from its parent `DelegationCall` (when the caller is itself a delegate target), a cycle or a chain past `Settings.delegate_max_depth` (1 for v1) is refused **and recorded as its own queryable call row** -- "one recorded refusal" -- not just a log line.
- **Real attribution (part 4).** `aci_protocol.DelegationMeta` (new optional `QueuedTurn.delegation` field, a patch bump) carries structured `immediate_caller`/`accountable_principal` fields, mirrored durably on `DelegationCall`. v1 identity is payload-level agent identity, not IdP-backed, per the maintainer's decision #2 on discussion #1049 -- `accountable_principal` equals `immediate_caller` today because depth is capped at 1.
- **Bundle declares, operator arms (part 5).** `PluginManifest.delegatesTo: list[str]` is the bundle-declared half (validated at deploy: non-empty, no self-reference, no duplicates); `POST /delegate/grants` stays the separate, explicit operator action, matching how this codebase already treats `Agent.approval_routes` (declared name, operator-bound value, not cross-validated between the two today -- a documented follow-up, not a silent gap).
- **Approval-gate correctness (part 4).** An `awaiting-approval` outcome from the target's turn is no longer treated as terminal. The target suspended on its own gate exactly as a user-started turn does, and will resume and fire another completion event -- resolving the call on `awaiting-approval` would have dropped that eventual real answer on the floor. Found and fixed while implementing, not present in the prototype's narrower demo path.

## What's still a documented gap, on purpose

- **The kernel steering-rule change ADR-0115 itself calls out is NOT done.** `apps/worker/CLAUDE.md`: *"One change owns \[kernel.py/consumer.py/threadlock.py/markers.py\] at a time... any change needs an adversarial review... if you are touching kernel.py as a side effect of another change, stop."* This PR does not touch those files. Instead, the delegate turn rides `TurnSource.WEBHOOK` deliberately (not a corner cut): `hooks.py`'s own docstring states `source=WEBHOOK` "is what stops it steering a live conversation," which is exactly the safety property needed here, already implemented via `is_job` and the kernel's existing bounded-retry reclaim (`ThreadBusyError`) -- no kernel change required. A dedicated message-lane `TurnSource` value (pending ADR-0112, also unimplemented) remains a legitimate future taxonomy cleanup, not a safety gap.
- **Depth is capped at 1** for exactly the reason above: tracing `accountable_principal` truthfully across a second hop needs the calling agent's own sandbox to see its own inbound turn's attribution, which needs a boot-env field only the kernel claim path can wire in.
- **`delegatesTo` is not cross-validated against `arm_grant`** -- arming a pair the bundle never declared currently succeeds, matching the existing `approval_routes` precedent rather than inventing a stricter check for this one feature.
- **One binding per agent (issue #1525, not yet built)**: arming a target replaces its previous channel binding. Real platform constraint, not something this PR introduces or fixes.

## Verification

- `apps/api`: 1100+ tests pass (`pytest apps/api/tests -q`), including new `test_delegate.py` covering the armed/unarmed gate, the depth cap, the cycle refusal (which caught a real bug in my own first pass -- comparing a bare UUID against `agent:`-prefixed chain entries, fixed), and the attribution the minted turn carries. `openapi.json` regenerated.
- `apps/worker`: 901 pass. `runner`: 594 pass. `packages/{aci-protocol,plugin-format}`: 502 pass.
- `cli`: 875+ pass (two hardcoded exact-wire-shape tests updated for the new `delegation` field; a handful of Valkey-backed tests are flaky under parallel execution against a shared real Valkey and pass cleanly with `--test-threads=1`, unrelated to this change).
- `ruff`, `mypy`, `cargo fmt --check`, `cargo clippy -- -D warnings`: clean.
- `scripts/check-contracts.sh`, `scripts/check-docs.sh`: clean (caught and fixed a stale JSONB-column-count claim in `docs/interfaces/relational-db/INTERFACE.md`).

## Review asks

Same five asks as ADR-0115's own PR (#1793), still open since the ADR is still Draft. Additionally, specific to this implementation:

1. Is reusing `TurnSource.WEBHOOK` (rather than blocking on a new message-lane value / ADR-0112) an acceptable interim, or should this wait for that value to exist?
2. Is depth-capped-at-1 an acceptable v1 scope cut, given raising it needs kernel-owner involvement this PR deliberately avoided?
3. Is the `delegatesTo`/`arm_grant` non-cross-validation acceptable at parity with `approval_routes`, or does this feature need the stricter check `approval_routes` itself lacks?


TheConnMan (approved):
@alexrao2000 approved, is this prod-quality or prototype code?

TheConnMan:
Post-approval review notes — I agree the capability is necessary. Please address these before merge:

- Split the `packages/aci-protocol` and `packages/plugin-format` changes into their own reviewed prerequisite contract PRs, per the frozen-contract rule. Coordinate the ACI version with #1935: both PRs currently publish different wire changes as `0.4.3`, so they cannot both land with that version/wire lock.
- Rebase onto current `main` and adapt this to ADR-0118 multi-surface bindings. The branch still uses scalar `Agent.channel`, and arming a delegate currently replaces the target agent’s existing binding.
- Preserve the delegation audit trail when an agent is deleted. `DelegationCall.caller_agent_id` and `target_agent_id` use `ON DELETE CASCADE`, which deletes the whole call/refusal row; this directly contradicts the adjacent claim that snapshotted provenance remains readable after deletion. Prefer nullable `SET NULL` relationships plus the existing snapshot fields, with a deletion regression test.
- Do not infer parent attribution from an unverified `body.caller_conversation_id`. A sandbox can name any `delegate:<uuid>` and the API inherits that row’s `chain` and `accountable_principal` without checking that the authenticated caller was that parent call’s target. Bind the parent to the authenticated caller/current turn, and add an adversarial spoof test.

After the rebase/contract prerequisites, rerun the required checks and E2E evidence on the final head.

PR #1933, opened 2026-08-26
Write down what staging actually runs, so it can be reproduced

jw3329:
Answers "where is the SRE bot implementation code, I can't run replication tests without it".

The short version: the implementation is in `examples/sre-bot` on `next`, but **it is not what staging runs**, and the difference was never written down. This adds `examples/sre-bot/docs/STAGING-DEPLOY.md` recording exactly what runs, why it differs, and the steps to reproduce it.

## Why nothing in the repo reproduced the live bot

Three of the five connectors declare `build:`. A local build records a local image id, which is refused at the cluster tier — a cluster cannot pull an image that exists only in one machine's Docker daemon. So a cluster deploy of this bundle needs published images or a registry, and the deployed bundle therefore differs from this directory in four ways:

| Delta | Why |
|---|---|
| `k8s-scale` and its gate removed | no published image; a gate naming an absent connector fails validation |
| `k8s-write`, `tempo` pinned to published images | the images built from the archived predecessor repository this bundle was extracted from, which is what the install has been running |
| allowlist filled with real targets | the bundle ships the `<namespace>/<deployment>` placeholder |
| `deploy.yaml` deleted | its placeholder channel ids can rebind a live agent, and the bot then goes silent with nothing in any log naming the cause |

The first one matters for how the bot reads to a reviewer: **the live bot has one write verb, not two.** Asked what it can do, it answers `restart_deployment` — correct for what is deployed, not a limitation of this directory.

## Three steps that exist because something fails silently

Each is documented where it bites:

- **The two allowlists.** `K8S_WRITE_ALLOWLIST` and the Role's `resourceNames` are two allowlists over one question in two files. They disagreed on this install for four days after the box was rebuilt — the Role was re-scoped by hand, the connector kept the previous cluster's target (`public/api`), the intersection was empty, so the write path was inert and nothing reported it.
- **`curie-grafana-connector`.** Created by a chart hook that does not exist in 0.7.0. Without it the connector pods fail `CreateContainerConfigError` while the previous ReplicaSet keeps serving — so `kubectl get deploy` reads `1/1` and the deploy looks successful. The doc says to read pod names and ages instead. I hit this and briefly reported a broken deploy as healthy.
- **`--env` defaults to `dev`.** Deploying a prod-environment agent without `--env prod` creates a version that is never applied: the command prints `deployed ... -> dev`, the reconciler reports `applied, 0 failed`, and the running connectors keep their old config. `--target` normally supplies this, which is why deleting `deploy.yaml` makes it explicit. I hit this too — the first promote was inert and only the connector env showed it.

## Provenance

Every value in the file was read off the working deployment rather than taken from intent. Two things are marked unverified rather than asserted: that the `build:` contexts produce the pinned images byte-identically, and whether a published `k8s-scale` image exists at all.

## What retires this file

PR #1923 (`--write-allowlist`) collapses steps 1–4 into one command and renders the Role's `resourceNames` and the connector's allowlist from one input, making their disagreement unrepresentable rather than merely detectable. Once it lands this file describes history. The one delta it does not cover is `k8s-scale`, which needs a published image before any cluster deploy can carry it.

### DIFF
--- a/examples/sre-bot/docs/STAGING-DEPLOY.md
+++ b/examples/sre-bot/docs/STAGING-DEPLOY.md
@@ -1,0 +1,262 @@
+# What actually runs in staging, and how to reproduce it
+
+This bundle is not deployable to a cluster as it stands, and the thing running in
+our staging cluster is this bundle plus four deltas applied by hand. That gap is
+why a reviewer could interrogate the live bot and then find nothing in the
+repository that reproduces it. This file closes that gap: it records the deltas,
+why each exists, and the exact steps.
+
+Written after the fact from a working deployment rather than from intent, so every
+value here was read off the cluster.
+
+**The host is a single-node k3s cluster**, not a managed one, which is why the
+chart-hook and registry-access notes below bite here and may not bite on yours.
+Do not confuse it with the cluster the bot *reports on*: that is a separate,
+larger cluster the read-only connector points at.
+
+**Two placeholders**, kept out of the repository because they name someone's
+private deployment rather than this project (`.gitleaks.toml`'s
+`downstream-repo-slug` rule):
+
+| placeholder | what it is |
+|---|---|
+| `<ARCHIVED_ORG>` | the org publishing the archived predecessor repository's images -- the same org as this repo; operators of this install already know it |
+| `<ns>` | the release namespace of the install you are reproducing into |
+
+## Why the bundle is not directly deployable
+
+```
+kubernetes   image   ghcr.io/containers/kubernetes-mcp-server@sha256:6d650f4b...
+k8s-write    build   connectors/k8s-write
+k8s-scale    build   connectors/k8s-scale
+grafana      image   docker.io/grafana/mcp-grafana@sha256:5efeafd0...
+tempo        build   connectors/tempo
+```
+
+Three connectors declare `build:`. `curie build --plugin-dir <dir>` without
+`--registry` records a local image id, which is usable at the skill and local
+tiers and **refused at the cluster tier** — a cluster cannot pull an image that
+exists only in one machine's Docker daemon. So a cluster deploy of this bundle
+needs either a registry to push to, or images already published.
+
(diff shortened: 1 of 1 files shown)


jw3329:
Force-pushed, and the branch is now one commit rather than two. Saying why, because a rewrite on an open PR deserves an explanation.

`gitleaks (full history)` failed with six findings, all `downstream-repo-slug`, all in this file. **They are not false positives.** The rule says "describe the repo's role, do not name it", and its allowlist covers this repository under both its names only — the archived predecessor's slug and the `ghcr.io/.../sre-bot-*-mcp` images built from it are exactly what the rule exists to catch. So `.gitleaksignore` was the wrong tool: that file is for reviewed false positives and says so at the top.

The gate scans the PR's whole base..head range, so a fix commit on top would have left the offending content in the two commits below it and the job would still be red. Hence the squash. No review comments were lost — there were none.

What changed in the content:

- The predecessor repository is described by its role. Its image refs carry an `<ARCHIVED_ORG>` placeholder, with a two-row placeholder table at the top, matching the convention that repo's own runbook already uses for install-specific values.
- **A correction the fix surfaced.** The doc pinned the archived tempo image as if a reproduction should use it. It should not: this repository publishes its own (`ghcr.io/curie-eng/curie-sre-bot-tempo`, from the `Build sre-bot-tempo image` job) and `curie example sre-bot install` resolves it to a digest. The pin is recorded because it is what the install *is* running, not what a new one should run. No in-repo equivalent exists for `k8s-write`, which is why that one still needs the placeholder — and that absence is the real blocker, now stated as one.
- The intro says the host is a **single-node k3s cluster**, and that it is not the cluster the bot reports on. That distinction is load-bearing for this file: the chart-hook and registry-access notes below it bite because of the host, and conflating the two clusters is the specific mistake this document exists to prevent.

Every recorded value was re-read off the running deployment while fixing this, and all four connector images still match what the file says.

PR #1934, opened 2026-08-26
Draft ADR 0133: the control agent renders screens, a human presses

alexrao2000:
Draft ADR for an agent that operates the platform from a chat channel. **Decision only. No implementation.** The realizing code path it names is in the stacked PR, held on this one.

## What it decides

The agent's authority ends at rendering.

* **Reads and proposals** go through a `control` scoped sandbox token (ADR 0033), minted for exactly one operator-named agent. Two settings on two services must agree, and both default to empty, so the plane ships closed.
* **Execution and button presses** require the platform key plus an operator resolved server side, where a sandbox cannot reach. The agent is refused on both.
* **A press names a screen and a button id, never an action.** The API re-renders the screen and reads the action off its own button, so a caller cannot compose one and the confirmed sentence and the executed parameters are the same fact.
* **Screens carry no channel rendering** (ADR 0020), so a Discord adapter needs a renderer and no new authorization decision.
* **CLI coverage is a gate**, not a claim: every leaf command maps to a screen or a written exemption, checked in both directions.

## Why the ADR is worth reading before the code

Three things in it are decisions rather than implementation detail, and the alternatives are recorded:

1. **Why the ADR 0010 approval gate is not sufficient here.** It works and stays the right mechanism for domain-scoped writes. It cannot protect the one thing under attack when the arguments are platform UUIDs: the human's reading of a model-authored sentence.
2. **Why irreversible actions are a separate table** rather than "the safe subset the model may execute". A conditional boundary invites the subset to grow.
3. **Why the exemption categories are the honest half.** They describe the real boundary between a CLI and a chat agent, and they are what keeps the coverage number from being achieved by relabelling.

## Numbering

This started life as 0114. After later ADRs landed and an open PR reserved 0132, it is renumbered to 0133 at rebase.

## Status

`Draft`, so per `docs/adr/AGENTS.md` it does not authorize implementation. A Draft may merge when publishing it for discussion is useful, which is the intent here: land the decision, discuss it, and accept it separately before the stacked PR merges.

Stacked implementation: #1935

### DIFF
--- a/docs/adr/0133-the-control-agent-renders-screens-a-human-presses.md
+++ b/docs/adr/0133-the-control-agent-renders-screens-a-human-presses.md
@@ -1,0 +1,313 @@
+# 133. The control agent renders screens; a human presses the buttons
+
+Date: 2026-08-22
+
+Status: Draft
+
+## Context
+
+Every agent Curie runs today is scoped to itself. Its sandbox token names its
+own agent and nothing else ([ADR-0033](0033-scoped-sandbox-state-token.md)), its
+state namespace is its own, and the platform key is deliberately kept out of the
+box. [ADR-0107](0107-an-agent-reads-its-own-runs-through-a-first-party-scoped-surface.md)
+(Draft) extends that self-scoping to observability: an agent may read *its own*
+runs. Nothing may read the fleet.
+
+The ask is an agent that operates the platform from a chat thread: "what is
+running", "why did the support bot fail that thread", "roll it back". Two
+distinct capabilities sit behind that, and they are not the same risk.
+
+**Reading across agents** is a widening of scope. Useful, and cheap to reason
+about: the worst outcome is a wrong or over-shared answer in a channel.
+
+**Changing the fleet** is different in kind, because the component asking for
+the change reads untrusted input for a living. A control agent lives in a
+channel where people paste tickets, forward alerts, and relay other agents'
+output. Give it a `kill_agent` tool and the tool is reachable by anything that
+can get text in front of it. That is not a hypothetical about this design; it is
+the ordinary operating condition of a chat agent.
+
+The obvious mitigation is a permission gate: declare the mutating tools in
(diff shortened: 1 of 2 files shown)


alexrao2000:
## What the decision looks like implemented

The control agent: screens, buttons, and the refusals

Recorded from the stacked implementation in #1935, not from this PR, which is the ADR alone. Every screen, label, button, refusal and audit id above is a live HTTP response against a real Postgres. The human's lines are scripted and no model runs, so it is deterministic and needs no model credential.

The chat frame is drawn in the terminal because that is what a terminal recording can show honestly. In an install with Slack connected the same payloads render as Block Kit: screens are channel neutral per ADR 0020, so this is one more renderer rather than a mock of another one.

Three moments in it bear on the decision under discussion here:

1. **A kill request the agent cannot fulfil.** It opens the screen and says the button is the operator's. The refusal is the platform's, not the model's good manners.
2. **An injected PagerDuty alert claiming pre-approval for a delete.** The API refuses it as an unknown action: `delete_agent` is not in the vocabulary the agent may propose, so there is no path even if the model were persuaded. This is the concrete case behind the two-table split in the Decision section.
3. **A non-operator presses a button** and gets `403 ... is not a Curie operator on this install`, resolved server side against the operator set rather than by the channel.

Reproduce with `bash examples/curie-control/demo/run.sh` on the #1935 branch.

TheConnMan:
Tiny clarification before merge: please state that the trusted `ScreenOut` → channel adapter → authenticated click → `/fleet/screens/actions` round trip is not implemented by the linked prototype; the demo proves the API behavior only. Name that surface path as required before acceptance, and replace the downstream repository identifier in the ADR.

TheConnMan (approved):
Approved to merge as a Draft. The ADR now explicitly makes the trusted surface round trip a prerequisite for later acceptance and implementation; this review does not accept the ADR or authorize the stacked implementation.

PR #1935, opened 2026-08-26
Moonshot ADR 0133: Implement the control agent: fleet plane, screens, and the bundle

alexrao2000:
Stacked on #1934. **Base is `task/curie-control-agent`, not `next`**, so the diff here is the implementation alone. Merge #1934 first, or retarget this to `next` after it lands.

Realizes Draft ADR 0133. That ADR does not authorize implementation, so this is held on its acceptance or a recorded ADR 0102 coordinated exception. Read the ADR first; this follows it.

The control agent: screens, buttons, and the refusals

Every screen, label, button, refusal and audit id in that recording is a live HTTP response against a real Postgres. The human's lines are scripted and no model runs, so it is deterministic and needs no credential. Reproduce with `bash examples/curie-control/demo/run.sh`.

## The trust boundary

| Layer | What it stops | Where |
| --- | --- | --- |
| The worker mints a `control` scoped token (ADR 0033) only when the resolved agent name matches `CURIE_CONTROL_AGENT` | Any other sandbox holding the credential at all | `apps/worker/src/curie_worker/binding.py` |
| The API verifies the token names ITS configured control agent, with scope `control` | A forged token, and a state token replayed as a control token | `apps/api/src/curie_api/routers/fleet.py` |
| Execute and button routes accept the platform key only | This agent running its own proposals or pressing its own buttons | same file |

Both settings default to empty, so the plane ships closed. The worker compares the name the platform assigned at deploy, never what `plugin.json` claims, so a bundle cannot elect itself.

## The design choice worth reviewing

**A press names a screen and a button id, never an action.** The API re-renders that screen and reads the action off its own button:

* there is no field in which to send `{"action": "delete_agent"}`;
* the sentence the operator confirmed and the parameters that run are the same fact;
* a stale tap is refused rather than replayed, so two operators racing on one posted message do not both act.

Pressing needs the platform key **and** an actor in `CONTROL_OPERATORS`. The second is what makes the first mean anything: a platform key caller is the dispatcher relaying a click, and without the operator check it would relay anyone's.

## CLI coverage: 26 with a screen, 59 exempt, of 85

`SCREEN_FOR_COMMAND` maps every leaf command in `cli/command-manifest.json` to a screen or a written exemption. `test_screen_coverage.py` fails in both directions.

The gate earned its keep during the rebase: `cluster surfaces` and `local surfaces` landed on `next` mid-flight and it failed until they were answered, which is a surfaces screen with per-binding unbind buttons rather than a quiet drop in coverage.

| Category | Why a button would have to lie | n |
| --- | --- | --- |
| `local-machine` | Your working directory, Docker daemon, files | 25 |
| `source-checkout` | Needs a repo clone and dev toolchains | 17 |
| `credential-entry` | A channel stores, indexes and displays what you type into it | 7 |
| `substrate-lifecycle` | Installs or destroys the cluster the agent runs inside | 5 |
| `cli-introspection` | The CLI describing itself to a coding agent | 3 |
| `is-the-channel` | `message` stands in for a channel when you have none | 2 |

## Bugs this surfaced

* **`control_proposals.target_agent_id` is SET NULL, not CASCADE.** CASCADE erased every kill, rollback and budget change ever made to a deleted agent, **including the deletion itself**. The one irreversible action left no trace. Caught by a test.
* mypy caught `version.label` on a model whose column is `version_label`. Every test passed at the time because none exercised a rollback; those tests were added after.

## Contract and platform changes

* Two declared boot keys, `CURIE_CONTROL_URL` and `CURIE_CONTROL_TOKEN`. ACI protocol 0.4.2 to 0.4.3, the compatible class under 0.x.
* Migration 0034, revising 0033.
* Ten `/fleet` routes declared in the closed telemetry operation domain (ADR 0076), manifest regenerated.

## Deliberately not in scope

* **Resolving approvals from a screen.** Approval cards already do this in channel with the authorizer, the self-approval block and the audit trail. A second resolve button would be a second implementation of the most safety-critical path in the system. The approvals screen reports and points at the card.
* **A `curie` verb for proposals.** ADR 0041 makes a new verb a three tier surface plus `api.rs` mirrors, field parity and manifest regeneration. Worth its own PR.
* **Adding a channel binding from a button.** Needs a channel id a button cannot carry. The surfaces screen says so rather than leaving it looking forgotten.

## Verification

```
ruff check apps packages examples          clean
mypy                                       clean, 233 source files
scripts/check-contracts.sh                 OK, artifacts current
scripts/check-docs.sh                      OK, catalog drift-free
scripts/check-commit-messages.sh           clean
plugin_format.validate_bundle              valid, no warnings
pytest apps packages                       10 failed
```

Those 10 failures are **byte-identical to the same command on unmodified `next`** (`2caced8e`), run from a clean worktree for comparison. Nine need a Langfuse the local stack was not running; one is a macOS specific `/etc` symlink assumption in `test_connector_lock.py`. This change adds none.

66 of the passing tests are new, across `test_fleet.py`, `test_screens.py`, `test_screen_coverage.py` and `test_control_token_mint.py`. They are written as attacks rather than feature coverage: replayed state tokens, wrong agent tokens, tampered signatures, expired tokens, self execute, self reject, caller supplied summaries, cross agent rollback, stale button replay, non operator presses, unbinding an agent's last surface, and an agent trying to propose a delete.

## Reading order

1. `apps/api/src/curie_api/routers/fleet.py`, the auth guards and the press path.
2. `apps/api/src/curie_api/screens.py`, the vocabulary and the CLI map.
3. `apps/api/tests/test_screens.py`, where the claims above are checked.


alexrao2000:
ADR numbering fix: this PR's stated ADR (`Draft ADR 0125`) was renumbered to **ADR-0133** when its prerequisite PR (#1934, "Draft ADR 0125: the control agent renders screens, a human presses") merged into `next` — it landed as `docs/adr/0133-the-control-agent-renders-screens-a-human-presses.md`, not 0125. ADR-0125 is already taken by a different, unrelated, Accepted ADR ("Managed repository workspaces and approval-gated publication are platform capabilities").

This PR's own diff still added `docs/adr/0125-the-control-agent-renders-screens-a-human-presses.md` and every in-code/test/doc citation still said "ADR-0125" — which would have collided with that unrelated ADR on merge. Pushed a fix: renamed the file to `0133-...`, regenerated `docs/adr/README.md`, and updated every citation (`apps/api`, `apps/worker`, `packages/aci-protocol`, `charts/`, `compose.dev.yaml`, `examples/curie-control/`, the demo `.tape`) from ADR-0125 to ADR-0133, plus the PR description. `check-docs.sh` and the affected test suites pass clean.

TheConnMan (approved):

TheConnMan:
Post-approval review notes — I agree the control-agent direction is necessary. Please address these before merge:

- Complete the operator-authentication boundary. `/fleet/screens/actions` accepts `actor` from `ScreenActionIn` and only checks that caller-supplied string against `CONTROL_OPERATORS`; any platform-key holder can therefore impersonate an allowlisted operator. The channel adapter must pass an authenticated principal through a mechanism the API can verify, with a spoofing regression test.
- The Draft ADR’s acceptance prerequisite is the real `ScreenOut -> channel adapter -> authenticated operator click -> /fleet/screens/actions` path. This diff contains no dispatcher/channel adapter, and the demo calls the API directly. Either add and exercise that path end to end or split/describe the API plane honestly without claiming the prerequisite is demonstrated.
- Move the frozen `aci-protocol` boot-key/version change into its own reviewed prerequisite PR. Coordinate it with #1921: both currently claim incompatible `0.4.3` wire changes.
- Rebase onto current `next`, resolve the conflict, and run the required checks/E2E on the resulting head. GitHub currently reports this PR conflicting and has no check runs for the present head.

PR #1936, opened 2026-08-26
Correct ADR-0121's measured MCP transport cost

jw3329:
**The decision is unchanged.** This corrects three claims in it that the spike
measured and found wrong — a reviewer reading ADR-0121 as it stands would be
weighing the wrong trade.

Refs #1867. Still Draft, authorizes no implementation. Probes on
`spike/adr-0121-restore-executor`.

## What was wrong

ADR-0121 wrote the runner-side executor up as **"the largest cost"**, inheriting
ADR-0117's spike conclusion:

> the SDK owns the client sessions, so a second client inside the runner is a
> large new mechanism for one job

A hosted connector's MCP entry is `{"type": "http", "url": ".../mcp"}` —
**streamable HTTP, not a stdio subprocess the SDK spawned.** No SDK-owned
session has to be reused for the call.

Against the real `k8s-scale` connector, with no agent SDK in the process:

```python
async with streamablehttp_client(url) as (read, write, _):
    async with ClientSession(read, write) as session:
        await session.initialize()
        result = await session.call_tool(tool, arguments)
```

The streamable-HTTP MCP client and call primitive is **13 lines**. That does not
measure the complete runner integration: sandbox lifecycle, pinned connector
selection, policy, failure handling, and completion confirmation remain unbuilt.

## Why this matters to the decision rather than just to the prose

It removes transport-client cost as the selector in decision 2. The same
primitive can run in the worker or runner, while total integration cost in
either location remains unmeasured. The case for the sandbox therefore rests on
reachability and the policy envelope, which is the architectural trade a
reviewer should judge.

## Also recorded: what the spike ran, and what remains missing

The ruling-to-restore leg ran through the real API, Postgres, and connector: an
agent scales 3 → 10, the ledger records the prior state, a refused undo leaves
the world at 10 and never reaches the connector, an authorized undo returns
`{target, prior_state}`, and the executor replays back to 3. No mapping table
exists anywhere, which is decision 1's claim surviving contact.

The complete loop has not run: `POST /actions/{id}/confirm-undo` returns 404, so
confirmation remains absent and decision 4's compromise stands until that
endpoint exists.

### DIFF
--- a/docs/adr/0121-a-restore-is-the-connectors-own-verb-run-under-the-same-pinned-connector.md
+++ b/docs/adr/0121-a-restore-is-the-connectors-own-verb-run-under-the-same-pinned-connector.md
@@ -165,9 +165,11 @@ executor should not ship before it.
 
 ## Alternatives considered
 
-- **An MCP client in the worker** (ADR-0117's first candidate). Not the smaller
-  change -- the spike above measured both at the same 13 lines -- so cost does not
-  choose between them and decision 2 does. The worker is the control
+- **An MCP client in the worker** (ADR-0117's first candidate). The spike
+  measured the streamable-HTTP MCP client and call primitive at 13 lines. That
+  primitive is the same in either process, so transport-client cost does not
+  choose between them; total integration cost in either location remains
+  unmeasured, and decision 2 chooses by policy boundary. The worker is the control
   plane; giving it a direct client of tenant connectors makes the platform
   reachable to places ADR-0008 keeps separate, and the undo would then run under
   the control plane's network reachability rather than the sandbox's. An undo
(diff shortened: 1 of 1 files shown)


TheConnMan:
Tiny evidence wording tweak before merge: scope “13 lines” to the streamable-HTTP MCP call primitive, not the complete runner integration, and describe the exercised path as ruling → restore rather than the whole loop because `confirm-undo` still returns 404.

TheConnMan (approved):
Approved. The correction now scopes the 13-line evidence to the MCP transport primitive, keeps total integration cost unmeasured, and accurately records that confirmation still returns 404.

PR #1942, opened 2026-08-27
Draft ADR-0126: a write connector declares permitted field paths, and the agent composes within them

jw3329:
Draft ADR. **Decision only, no implementation** — per `docs/adr/AGENTS.md` a merged Draft does not authorize any.

Answers the gap named in #1886's review: that a per-bundle check over one write connector is not a model for connector authorization, and that the target is builder-owned policy over the published tool surface with unclassified operations denied.

## The decision

One write tool per resource kind. What it may change is a **builder-declared set of field paths with value constraints, enforced inside the connector.** The agent composes a change within those paths and can neither author a path nor widen the set.

`restart` and `scale` become two policy entries against one tool instead of two connectors:

```
apps/Deployment  /spec/template/metadata/annotations/kubectl.kubernetes.io~1restartedAt
apps/Deployment  /spec/replicas
```

Each entry is classified `deny` / `approval-required` / `allow`, and an operation matching no entry is denied — the tri-state the review asked for, given a subject specific enough to carry it. **A tool name cannot hold three classifications; a field path can.** That is the whole idea.

## Why this shape and not the two obvious ones

The ADR records both alternatives with measured costs rather than arguments.

**Per-verb connectors (the status quo).** 580 lines across two near-duplicate files for two operations — kubeconfig loading, client construction, allowlist parsing, the TLS refusal, timeout handling and the ADR-0117 prior/post report all written twice; what differs is one tool function and one patch body. Plus an image, a gate and an allowlist each. And the second verb, `k8s-scale`, is **merged and still undeployable**, because no image was ever published for it and a gate naming an absent connector fails validation. That is what "linear in verbs" looks like when you go and look.

**A general passthrough the agent controls.** This is the attractive one and it deserves a real answer rather than a reflex. It collapses the 580 lines and every future verb to zero. It also moves the ceiling to the only persuadable component, and RBAC cannot catch it: `patch` on `deployments` is exactly the grant that permits `set image`, `set env`, and an arbitrary template patch, and there is no RBAC expression that separates a restart from replacing what runs.

What makes that concrete rather than theoretical for *this* bot is its read ceiling. `read-access.yaml` grants `pods/log` and `configmaps` cluster-wide — verified on a live install with `kubectl auth can-i --as=` — so the bot's ordinary working input is text anyone who can write a log line controls, and its output is a chat message. The file's own header already reasons this way about why `configmaps` is in the list and `secrets` is not.

And the approval gate does not rescue it: with one tool every call is gated, so what the human reads at 3am is a model-authored patch body. Draft ADR-0125 (#1934) names that exact failure.

## The seed was already in the tree

`connectors/k8s-scale`'s docstring:

> `replicas` is a caller parameter here because it is the verb's argument rather than a channel into an arbitrary patch

A caller-supplied **value at a fixed path** is safe; a caller-supplied **path** is not. This ADR generalizes that one sentence.

## What it deliberately does not decide

- **Whether an agent may change its own policy.** It may not — the decision is worthless if the constrained party can edit the constraint. *How* a policy change is made belongs to Draft ADR-0125 (#1934).
- **Bundle-owned vs platform-owned policy.** This says builder-owned, matching where `approvalPolicy` and `connectors.yaml` already live. Operator-owned policy over every bundle on a cluster is a stronger answer with a bigger blast radius, and a separate decision.

## Costs it does not hide

The policy evaluator is real work — strategic-merge vs JSON-patch semantics, escaped path segments, value validation — and it becomes load-bearing, so a bug in it is a ceiling failure. It needs the test attention the per-verb constant patch bodies get for free today. Paid once rather than per verb.

It also changes the gate vocabulary, which is a real cost against ADR-0050's "armed exactly or not at all" — that ADR is worth reading before this one is accepted.

`bash scripts/check-docs.sh` is green; the regenerated `docs/adr/README.md` is committed.


TheConnMan (changes requested):
We discussed using vanilla MCP with bot-level allow / approval-required / deny policy instead. This ADR has been superseded by that direction.

jw3329:
Closing per your review — the vanilla-MCP-with-bot-level-tri-state direction supersedes this, and I agree with it.

Recording why my objection was wrong, since it is the useful part: I argued the ceiling could not move to "bot level" because the agent is prompt-injectable. I was reading "bot level" as "the model decides". It isn't — the policy is bot *configuration*, which the model cannot edit, so the thing I was guarding against does not happen in your direction either. I was also only imagining "thin" as something to arrange inside a connector, and missed the simpler place to put it.

Two things from the original reasoning that are worth carrying into the new shape, both discovered the hard way and documented in `connectors.yaml`:

1. `configuration_view` returns the kubeconfig as YAML, bearer token in cleartext, and it is annotated `readOnlyHint=true` — because reading a credential is a read. The bundle's comment notes that "all 14 tools are readOnlyHint" was verified and still missed it. With `--read-only` dropped, that tool has to be denied explicitly; annotation-derived classification will not catch it.
2. The credential widens even as the policy narrows. Today's write kubeconfig can patch three named Deployments and nothing else. A vanilla server covering several verbs needs RBAC for all of them, and RBAC has no field granularity — so `patch deployments` still means "can replace what runs". The policy stops the agent; it does not stop a compromised connector. Worth deciding deliberately rather than inheriting.

The measurements are already recorded elsewhere and are not lost with this PR: #1944 (the forward-merge divergence), #1945 (the unpublished write connectors), and `examples/sre-bot/docs/STAGING-DEPLOY.md`.

PR #1961, opened 2026-08-27
Recover stranded cluster-message turns after a callback-trust rollout

### LINKED ISSUE #1532: Reply delivery lifecycle: dead reply targets poison the run stream and a worker restart strands pending entries for 15 minutes
Two faces of the same consumer and delivery lifecycle, both observed live.

## 1. Dead reply targets poison the run stream

When a reply target dies (a killed CLI stub), the worker retries delivery forever. The stream entry stays pending, and with a single consumer the head of line blocks every later turn for every agent on the install. Recovery required manually `XACK`ing the entries on the valkey pod and restarting the worker. Needs a delivery TTL or a dead letter path.

Reproduces trivially: `curie cluster message`, kill the CLI mid turn, send another message.

## 2. Worker restart strands pending entries for 15 minutes

On a worker pod restart the new consumer does not claim its dead predecessor's pending entries at startup. They sit until `reclaim_min_idle_ms` (15 minutes by default), so a turn submitted just before a rollout goes silent for 15 minutes. Claiming a dead consumer's pending entries at startup would close it.

## Environment

Single node k3s, chart and CLI v0.7.0-rc.4 built from main (e2c68941), cluster tier. Observed 2026-08-14.

TheConnMan:
## Summary

The first `curie cluster message` after a default install temporarily trusts the reply stub, which rolls the single worker. `kubectl rollout status` returns while the outgoing pod is still Terminating and still blocked in XREADGROUP, so that consumer can claim the turn, fail to write the reply, and leave the entry pending for the 15-minute reclaim window.

This change does two things. The CLI does not start enqueue plumbing until worker pods have no `deletionTimestamp` and at least one Ready replica. The runs-lane consumer also claims a dead peer's pending entries from `XINFO CONSUMERS` idle (default 15s) so a stranded PEL is recovered without waiting `reclaim_min_idle_ms`. Eval keeps the 15-minute entry-idle window because its read loop awaits the handler inline.

## Related issue

Closes #1532

## End-to-end verification

| skill | n/a | Does not reach plugin packaging, the runner turn loop, ACI events, or skill eval/check. | | |
| local | required | Dead-consumer reclaim runs in the worker stream consumer used by local and cluster. | fake | `uv run pytest -q apps/worker/tests/kernel/test_consumer.py::test_reclaims_a_dead_consumers_pending_entry_without_waiting_min_idle apps/worker/tests/kernel/test_consumer.py::test_reclaim_does_not_steal_from_a_fresh_live_peer_when_min_idle_is_high` on `b12d046c` against compose Valkey. Observed: `2 passed in 0.06s`; `reclaimed == 1` with `reclaim_min_idle_ms=900000`. |
| local-release | n/a | Does not change released binary identity, install path, version pins, or release compose. | | |
| cluster | required | First `cluster message` after a default install races its own callback-trust worker rollout. | fake | Rebuilt `curie` against `b12d046c`: `curie cluster message hello --namespace curie-1532-e2e --release curie1532 --channel C0EXAMPLE1 --listen-host 10.20.30.40 --timeout-secs 5 --debug` after installing a throwaway worker Deployment with `terminationGracePeriodSeconds=8`. Observed: `elapsed_s=12`; kubectl log is patch, `rollout status`, then 18 `get pods` polls, then `port-forward`. Namespace `curie-1532-e2e` deleted. |
| live provider | n/a | The race is stream-consumer identity during rollout, not model routing or provider credentials. | | |
| external integration | n/a | Does not reach Slack, git webhooks, connector OAuth, or a third-party API shape. | | |

Positive: first invocation waited out the terminating dummy worker (18 pod polls, 12s) before enqueue plumbing. Negative/control: `test_reclaim_does_not_steal_from_a_fresh_live_peer_when_min_idle_is_high` left a just-claimed live peer pending (`reclaimed == 0`); `worker_pods_refuse_enqueue_while_a_terminating_worker_can_still_claim` returns false while `deletionTimestamp` is set.

Durable red-on-revert: `cli/tests/cluster_stub_trust.rs` (`disconnected_cluster_message_waits_out_a_terminating_worker_before_enqueue`) and the two kernel tests above.

## Checklist

- [x] Tests pass for the area I touched (see CONTRIBUTING.md for the commands).
- [x] Docs updated if behavior changed.
- [x] An ADR is added under `docs/adr/` if this is an architectural decision.

No ADR. StreamBroker gained `xinfo_consumers` and `xclaim` the same way ADR-0039 added verbs; INTERFACE.md is updated. Frozen ACI/plugin-format contracts are unchanged.

Defaults recorded: base is `main` (v0.7.3 patch shared by both lines); runs-lane `dead_consumer_idle_ms` default 15000; eval uses `reclaim_min_idle_ms`; SIGTERM does not cancel an in-flight XREADGROUP.


TheConnMan:
Retriggering checks: GitHub Actions did not start on 43b9df50, and the Claude/Cursor suites stayed queued with no runs. Closing and reopening to fire pull_request events again.

PR #1988, opened 2026-08-27
Install the SRE bot's gated write connector by default, with an empty ceiling

jw3329:
> the SRE bot should have its write connector enabled by default

Doing that. Absence of `--write-allowlist` used to remove the connector; it now installs it with an **empty ceiling**, and `--no-write` is the explicit opt-out.

## Why default-on is the right call

A capability nobody can see is indistinguishable from one that was never built. That is not hypothetical here — it is exactly how this bundle's second write verb came to read as a design choice rather than a missing image (#1945). Installed with an empty ceiling, the tool is published and gated, and the bot can answer "I can restart a Deployment, and no targets are configured" instead of the verb simply not existing.

## The half worth reviewing

Two ceilings have to be closed for "installed, empty" to be a safe state.

**The connector.** It starts, publishes its tool, and refuses every call. This is not a new behaviour — its own source says so:

> the server starts, answers, and refuses every call. That is the correct default for a write path: an operator opts workloads in by name, and a missing config fails closed rather than granting everything.

**The Role — and this is the trap.** An RBAC rule whose `resourceNames` is empty or absent **does not grant nothing. It grants the verb on every resource of that type.** So the empty case must *omit* the Role rather than render one with no names. It emits the ServiceAccount and its token Secret and stops.

Getting that backwards would have made the new default hand the bot `patch` on every Deployment in the namespace — the exact opposite of what an empty allowlist reads like, and it would have looked correct in review. There is a test pinned on it.

The identity is minted either way, because bring-up refuses without `K8S_WRITE_KUBECONFIG`. An identity that can do nothing is what makes this state bootable.

## Tests

- no targets renders exactly `ServiceAccount` + `Secret`, no `Role`
- no targets keeps the connector with an **empty** allowlist, never the shipped `<namespace>/<deployment>` placeholder
- `--no-write` leaves the connector absent, exactly as absence used to

Load-bearing: rendering a Role with empty `resourceNames` for the empty case fails the first and leaves the other 915 green.

`--no-write` together with `--write-allowlist` is a usage error rather than a silent precedence rule.

## Your other question

> was there no MCP server for making k8 operations we could have used?

There is, and the bundle already uses it — `ghcr.io/containers/kubernetes-mcp-server`, for the whole read side. It is run `--read-only --toolsets core --disable-destructive --disable-multi-cluster`, and it publishes 13 tools, all read.

Two things from `connectors.yaml` on why the write side was purpose-built instead of just dropping `--read-only`, both learned the hard way and both worth carrying into the tri-state direction:

1. **`--read-only` is not sufficient by itself.** `configuration_view` returns the kubeconfig as YAML — bearer token in cleartext — and it is annotated `readOnlyHint=true`, because reading a credential is a read. The comment notes that "all 14 tools are readOnlyHint" was verified and still missed it; dropping to `--toolsets core` removes exactly that tool. Under a policy table that classifies from annotations, this one has to be denied explicitly.
2. **The reason that no longer holds.** The other argument was that dropping `--read-only` exposes create/update/delete across every kind, "gated only by what `approvalPolicy` remembers to name — and a tool nobody thought to gate has nothing in front of it." That was true when gating was allowlist-by-name. With unclassified-denied it is not: a forgotten tool is denied, not open. So your direction removes the premise the purpose-built connector was built on, and I think it is the better shape.

One thing that does not change with it, flagging because the policy table will not show it: the *credential* widens even as the policy narrows. Today's write kubeconfig can patch three named Deployments and nothing else. A vanilla server covering several verbs needs RBAC for all of them, and RBAC has no field granularity — so `patch deployments` still means "can replace what runs". The policy stops the agent; it does not stop a compromised connector. Worth deciding deliberately rather than inheriting.

### DIFF
--- a/apps/ui/src/generated/commandManifest.ts
+++ b/apps/ui/src/generated/commandManifest.ts
@@ -4740,10 +4740,22 @@ export const commandManifest = {
                 },
                 {
                   "global": false,
-                  "help": "Keep the approval-gated restart tool, scoped to these Deployments (`namespace/name`, comma separated). One list renders BOTH ceilings: the Role's resourceNames and the connector's K8S_WRITE_ALLOWLIST. Omit to install read only",
+                  "help": "Scope the approval-gated restart tool to these Deployments (`namespace/name`, comma separated). One list renders BOTH ceilings: the Role's resourceNames and the connector's K8S_WRITE_ALLOWLIST. Omit and the connector is still installed, gated, with an empty ceiling that refuses every call until targets are named",
                   "id": "write_allowlist",
                   "long": "write-allowlist",
                   "positional": false,
+                  "required": false
+                },
+                {
+                  "global": false,
+                  "help": "Leave the gated write connector out of the install entirely",
+                  "id": "no_write",
+                  "long": "no-write",
+                  "positional": false,
+                  "possible_values": [
+                    "true",
+                    "false"
+                  ],
                   "required": false
                 },
                 {
--- a/cli/command-manifest.json
+++ b/cli/command-manifest.json
@@ -4737,10 +4737,22 @@
                 },
                 {
                   "global": false,
-                  "help": "Keep the approval-gated restart tool, scoped to these Deployments (`namespace/name`, comma separated). One list renders BOTH ceilings: the Role's resourceNames and the connector's K8S_WRITE_ALLOWLIST. Omit to install read only",
+                  "help": "Scope the approval-gated restart tool to these Deployments (`namespace/name`, comma separated). One list renders BOTH ceilings: the Role's resourceNames and the connector's K8S_WRITE_ALLOWLIST. Omit and the connector is still installed, gated, with an empty ceiling that refuses every call until targets are named",
                   "id": "write_allowlist",
                   "long": "write-allowlist",
                   "positional": false,
+                  "required": false
+                },
+                {
+                  "global": false,
+                  "help": "Leave the gated write connector out of the install entirely",
+                  "id": "no_write",
+                  "long": "no-write",
+                  "positional": false,
+                  "possible_values": [
+                    "true",
+                    "false"
+                  ],
                   "required": false
                 },
                 {
--- a/cli/src/examples.rs
+++ b/cli/src/examples.rs
@@ -50,6 +50,13 @@ const WRITE_RULE_RESOURCES: [&str; 1] = ["deployments"];
 const WRITE_RULE_VERBS: [&str; 2] = ["get", "patch"];
 const READER_TOKEN_TIMEOUT: &str = "2m";
 const KUBECONFIG_SECRET_KEY: &str = "K8S_READONLY_KUBECONFIG";
+// The gated write connector's published image. A `sha-<commit>` tag rather than a
+// semver one because no tagged release has carried this connector yet: it is
+// published by `release.yaml` on every push to a release branch, and a commit tag
+// is immutable in the way `latest` is not. Move this to a semver tag once a
+// release publishes one.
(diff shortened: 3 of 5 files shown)


jw3329:
Converting this to draft rather than pushing a green fix, because the remaining CI failure is a real ordering dependency and not something the code can settle.

## What the failure is

Six integration tests in `cli/tests/example_sre_bot_install.rs` fail with:

```
Error: connectors.k8s-write
  (resolve_context: the build context does not exist in the embedded workspace)
```

Installing the connector by default means the deploy now tries to **build** it, because the bundle declares `build: connectors/k8s-write` and the embedded workspace carries only the bundle's files, not the connector source. Even with the source present, a cluster deploy would still be refused: `build:` records a local image id, and a cluster cannot pull an image that exists only in one machine's Docker daemon.

So default-on requires the connector to be declared with a resolved `image:`, the same treatment tempo already gets at install time.

## Why I am not just doing that

I wrote it — generalised the tempo digest resolver, resolved `ghcr.io/curie-eng/curie-sre-bot-k8s-write:<tag>`, swapped `build:` for `image:` — and then stopped, because **that image does not exist yet.** Nothing publishes it until #1945 merges and a release runs.

I could make CI green by teaching the test registry stub to serve a second repository. The tests would pass and every real install would fail on a 404 from GHCR. That is precisely the shape of failure I have been filing issues about this week, and I am not going to introduce one to clear a red check.

## The order

```
#1945 merges            (adds the publish jobs for both write connectors)
  → one release runs    (the images exist)
  → digest resolution   (small PR; I have it written)
  → this PR is landable
```

`--no-write` and the empty-ceiling semantics in here are unaffected and still correct; the RBAC point the PR body makes — an empty `resourceNames` grants every Deployment, so no targets must render no Role — stands on its own and has a test pinning it.

Unblocked by #1945 plus a release, and I will pick it straight back up.

jw3329:
Out of draft — the blocker turned out not to exist.

I said this needed a release before the write connector's image would exist. That was wrong: `release.yaml` runs on **every push to a release branch**, so #1945's matrix entries have been publishing since it merged. Verified by pulling all three connector images at a `next` commit from a cluster that has registry access:

```
sre-bot-k8s-write   PULLED
sre-bot-k8s-scale   PULLED
sre-bot-tempo       PULLED
```

So the digest step is done here rather than deferred, and this PR now stands on its own.

## What changed since the draft

The tempo-only digest resolver takes a repository and tag, and the kept write connector gets the same `build:` → pinned `image:` treatment. The swap runs **after** the known-connector check — doing it before reported an unrecognised connector as a missing image, which an existing ordering test caught.

Fixtures, none of it weakening an assertion:

- The mock registry answers for any published sre-bot connector rather than one hard-coded repository, so adding a connector cannot silently turn these into "registry unreachable" tests.
- The kubectl stub answers for the writer token the way it already does for the reader, because the write identity is part of the default install now.
- Three expectations move to the new default rather than being deleted: the connector is present, pinned by digest, its ceiling is empty, and its gate ships with it.
- The registry-call assertion goes from a count to a per-repository check. The count alone would have passed if one connector were resolved twice.

## Test evidence

`cargo test --no-fail-fast`: **1783 passed, 13 failed.** Clean `origin/next` scores **1780 passed and the same 13 failures, by name** (migrate-store, parity ladder, eval trajectory, apply import). Pre-existing, untouched by this branch, which adds the 3.

## Still a decision for a reviewer

The image tag is `sha-<commit>` rather than semver, because no tagged release has carried this connector yet. It is immutable in the way `latest` is not, but it should move to a semver tag once a release publishes one — flagging it rather than leaving it to be discovered.

Issue #1603, opened 2026-08-17
Re-enable the cluster rung weather grade once a grader can assert provenance


TheConnMan:
## Summary

The graded weather case is disabled on the nightly ladder's live cluster rung. This ticket
is the record of why, and of what has to be true before it goes back on.

It is disabled rather than fixed because the case cannot express what it is meant to
assert. That is a grader capability gap, not a bug in the egress posture, the sandbox, or
the example agent.

## What the case actually measures today

`reports-a-temperature` in `examples/weather/evals/cases.json` grades on the regex
`\d+\s*(°|deg)`, which tests only that a temperature figure is PRESENT. It cannot test
where the figure came from.

Under the cluster rung's provider only egress the agent cannot fetch a forecast page, and
it responds in one of two ways depending only on how it chooses to phrase itself:

* It refuses cleanly and gives no figure. The case reds.
* It says plainly that it could not fetch, then supplies search summary numbers anyway
  behind a caveat. The figure is present, so the case greens.

Measured across three runs on commit `bb0c24f9` and its descendants:

| Run | Rung state | What the agent said |
| --- | --- | --- |
| 32009352813 | red | refused, no figure |
| 32019827183 | green | "weather.gov, AccuWeather, and BBC Weather were all blocked ... The figures above come from search-result summaries" |
| 32019884344 | green | egress deliberately pointed at TEST-NET-1 so nothing was reachable. Still green. "These numbers came from search-result summaries, not pages I was able to open and verify myself" |

Run 32019884344 was built as a falsifiability probe and was expected to red. It did not.
**So neither the green nor the red depends on whether the fetch succeeded.** The rung was
grading model phrasing.

## Why it was not fixed instead

Two approaches were tried or evaluated and both are closed.

**Resolve the forecast host and pin the addresses.** Structurally broken. The GitHub runner
resolved `forecast.weather.gov` to two addresses in `23.220.144.0/24`; the same three
hostnames resolve to seven addresses in `23.1.33.0/24` from another resolver. The pod
resolves through CoreDNS, not the runner's resolver, so runner side `/32`s do not match
what the pod dials. Verified end to end: the flags were accepted and the NetworkPolicy
carried both routes ("web egress opened to 2 declared destination(s)") and the fetch still
failed.

**Serve a deterministic forecast source inside the cluster.** Blocked twice.
`charts/curie/templates/security-networkpolicy.yaml` renders `runner-default-deny-egress`
[...]

Issue #1534, opened 2026-08-14
Per thread sandbox retention exhausts the sandbox ResourceQuota and ClaimTimeoutError diagnostics blame CPU


TheConnMan:
## Summary

Each thread pins a sandbox with a 1 CPU limit for the retention window, against the namespace's hard 8 CPU sandbox ResourceQuota. Eight CLI messages (or eight eval cases; the suite has twelve) exhaust the namespace and every later claim times out. The eval suite cannot complete on a default install for this reason alone.

## The diagnostics compound it

`ClaimTimeoutError` names CPU saturation as the likely cause. The three real causes hit in one day were quota exhaustion, controller RBAC, and reply delivery poisoning, on a node idle around 14 percent CPU. The error should surface quota state and controller conditions instead of guessing.

## Relationship to #1492

#1492 covers warm pools and the cold boot cost; this is the retention economics face of the same conversation (what a thread holds, for how long, and what back pressure looks like) plus the diagnostics fix, which is independent and small.

## Environment

Single node k3s, chart and CLI v0.7.0-rc.4 built from main (e2c68941), cluster tier. Observed 2026-08-14.

jw3329:
Picking this up. Your quota diagnosis reproduces structurally, and there is a second failure next to it that explains why my install behaved differently from yours — and that I think belongs in this issue, because it makes the ceiling you hit disappear silently rather than bind.

## The quota can be scoped to a PriorityClass that does not exist

`tenant-resourcequota.yaml` scopes on `$rq.sandboxPriorityClassName`:

```yaml
scopeSelector:
  matchExpressions:
    - scopeName: PriorityClass
      operator: In
      values: [ {{ $rq.sandboxPriorityClassName | quote }} ]   # default "curie-sandbox"
```

That value is independent of `priorityClasses.sandbox.name`, which is what the pods actually carry. Renaming the sandbox PriorityClass — without also overriding `resourceQuota.sandboxPriorityClassName` — leaves the quota matching a class nothing carries.

Our install does exactly that (`priorityClasses.sandbox.name: sre-bot-sandbox`, quota left at its default). Verified on the cluster, with a sandbox actually running:

```
$ kubectl get priorityclass | grep sandbox
sre-bot-sandbox   100000        # curie-sandbox does not exist at all

$ kubectl get pods -l app.kubernetes.io/component=runner-sandbox \
    -o custom-columns=PC:.spec.priorityClassName,CPU:...limits.cpu
[...]

PR #2045, opened 2026-08-29
ci: build curie-ui in the nightly local-release ladder rung

### LINKED ISSUE #2005: Nightly local-release rung fails: the full compose profile needs curie-ui, which the job never builds
## Problem

The nightly graded ladder has failed on eight consecutive scheduled runs, 2026-08-20 through 2026-08-27. Every one fails the same single job, `Graded parity ladder (local-release, live model)`, with a byte-identical terminating error (logs sampled from runs on 08-20, 08-22, 08-25, 08-27):

```
error: image 'ghcr.io/curie-eng/curie-ui:latest' is required by
compose.release.yaml's full profile and is not present locally.
```

The other three jobs pass, including `skill + local` and `cluster on kind` against a live model. This is one cause, not a flaky streak. (The 08-23 run also failed `skill + local`; that appears unrelated and has not recurred.)

## Root cause

The regressing commit is `3fcd8c9c` ("Wire trajectory scoring across eval tiers"). It is **not** an ancestor of the last green run's SHA (`402dd242`) and **is** an ancestor of the first red run's SHA (`70213a42`), which isolates it exactly to the green-to-red boundary.

It added this branch to `cli/scripts/e2e-ladder.sh`:

```bash
local compose_profile="core"
if [[ -f "$WORKDIR/bundle-release/evals/trajectory.json" ]]; then
    compose_profile="full"
fi
```

The ladder's bundle is the weather example and `examples/weather/evals/trajectory.json` exists, so the profile flipped from `core` to `full`. The `full` profile pulls in Langfuse and the UI, and the script then hard-fails on any `ghcr.io/curie-eng/*` image not present locally.

The `ladder-local-release` job in `.github/workflows/nightly-graded-ladder.yaml` satisfies exactly two such refs by hand, and its own comment says why:

> The generated release compose pins `:latest` images and carries no build directive, so satisfy those refs here: retag the api build, and build the worker-local overlay with plain `docker build`.

That list was written when this rung used the `core` profile. It retags `curie-api:ci-local` to `:latest` and builds `curie-worker-local:latest`. `curie-ui` is the third ref the `full` profile needs, and nothing in that job builds it.

## Fix

Add a UI build step to the `ladder-local-release` job, tagged `ghcr.io/curie-eng/curie-ui:latest`. The pattern already exists in the same workflow: the cluster job builds `apps/ui/Dockerfile` with `load: true` and tags it `curie-ui:local`. Mirror that with the `:latest` ghcr tag the generated compose pins.

`curie-ui` is the only curie-owned image the `full` profile needs that this job lacks. Langfuse, ClickHouse, and the OTel collector images are public and pulled on demand, and the script only checks `ghcr.io/curie-eng/*` refs.

## Non-cause, do not chase

Earlier in the same job the buildx layer cache emits `digest-mismatch: error` and repeated `blob sha256:...: not found` from the `type=gha` scopes. The run proceeds past these and they do not cause the failure. Do not rotate cache scopes as part of this change.

## Done when

- `ladder-local-release` builds and loads an image tagged `ghcr.io/curie-eng/curie-ui:latest` before the ladder step runs.
- A `workflow_dispatch` run of `nightly-graded-ladder.yaml` completes with all four jobs green.
- The next scheduled nightly is green.

TheConnMan:
The graded nightly's `ladder-local-release` job hand-satisfies the generated `compose.release.yaml`'s `ghcr.io/curie-eng/*` refs. That list was written when the rung used compose's `core` profile. `3fcd8c9c` made `cli/scripts/e2e-ladder.sh` select the `full` profile whenever the bundle carries `evals/trajectory.json` — which `examples/weather` does — so the rung has needed `ghcr.io/curie-eng/curie-ui:latest` ever since. Eight consecutive scheduled nightlies (2026-08-20 → 08-27) failed on exactly that missing image.

`ci.yaml`'s `e2e-ladder-release` job already builds it. This adds the same step, verbatim, at the same position, restoring parity between the two jobs.

Verified locally that `curie-ui:latest` is the only curie-owned ref the `full` profile needs that this job lacked:

```
$ python3 compose/generate_release_compose.py | docker compose -f - --profile full config --images
ghcr.io/curie-eng/curie-api:latest
ghcr.io/curie-eng/curie-ui:latest
ghcr.io/curie-eng/curie-worker-local:latest
(+ langfuse, clickhouse, otel-collector, postgres, valkey, rustfs — all public)
```

Per the issue, the `type=gha` cache scopes are untouched; the digest-mismatch / blob-not-found lines are incidental.

Closes #2005

### DIFF
--- a/.github/workflows/nightly-graded-ladder.yaml
+++ b/.github/workflows/nightly-graded-ladder.yaml
@@ -205,10 +205,27 @@ jobs:
           load: true
           cache-from: type=gha,scope=ladder-worker
           cache-to: type=gha,mode=max,scope=ladder-worker
+      # Mirrors ci.yaml's e2e-ladder-release step of the same name. The ladder
+      # selects compose's `full` profile whenever the bundle carries
+      # evals/trajectory.json, which the weather bundle this rung runs does, and
+      # that profile pins the ui image too (#2005). The ui image is not a FROM
+      # consumer, so it builds on the named cache-only builder like api/worker.
+      - name: Build the ui image locally (satisfies the generated release compose, no registry auth)
+        uses: docker/build-push-action@53b7df96c91f9c12dcc8a07bcb9ccacbed38856a # v7.3.0
+        with:
+          context: .
+          file: apps/ui/Dockerfile
+          builder: ${{ steps.releasecache.outputs.name }}
+          tags: ghcr.io/curie-eng/curie-ui:latest
+          push: false
+          load: true
+          cache-from: type=gha,scope=ladder-ui
+          cache-to: type=gha,mode=max,scope=ladder-ui
       # The generated release compose pins `:latest` images and carries no build
-      # directive, so satisfy those refs here: retag the api build, and build the
-      # worker-local overlay with plain `docker build` (it is the FROM consumer,
-      # so it must stay on the default plain-docker builder, #803).
+      # directive, so satisfy those refs here: the ui build above, plus retag the
+      # api build, and build the worker-local overlay with plain `docker build`
+      # (it is the FROM consumer, so it must stay on the default plain-docker
+      # builder, #803).
       - name: Tag the api image for the generated release compose (ghcr.io/curie-eng/curie-api:latest, no registry auth)
         run: docker tag ghcr.io/curie-eng/curie-api:ci-local ghcr.io/curie-eng/curie-api:latest
       - name: Build the worker-local overlay for the generated release compose (ghcr.io/curie-eng/curie-worker-local:latest, no registry auth)


TheConnMan:
`workflow_dispatch` run on this branch: https://github.com/curie-eng/curie/actions/runs/33233807056

**The target job is green.** `Graded parity ladder (local-release, live model)` passed for the first time since 2026-08-19 — the `full`-profile preflight now finds `ghcr.io/curie-eng/curie-ui:latest` and the rung runs to completion. `build-cli` also passed.

Two other jobs failed, both for reasons unrelated to this diff (which touches only `ladder-local-release`):

1. **`skill + local` (live model)** — a *trajectory grade* failure, not plumbing. The turn finalized fine; `curie local eval` then failed the case:
   ```
   reports-a-temperature: mode=in_order expected=['WebSearch', 'WebFetch'] observed=['Skill', 'WebFetch']
   ```
   The live model reached for `Skill` instead of `WebSearch`. This is the strict `in_order` trajectory expectation from `3fcd8c9c` meeting live-model nondeterminism.

2. **`cluster on kind` (live model)** — the known #1534 shape:
   ```
   cluster: message after repeated eval timed out at 45s; eval-owned sandboxes likely still hold the quota (#1534).
   ```
   The eval itself passed (report-only per #1603); the follow-on message hit the sandbox-quota timeout.

Neither is caused by, or fixable in, this change. #2005's "all four green" done-when therefore needs those two tracked separately.

PR #2082, opened 2026-08-29
Wire tests/soak into `curie dev soak` and a report-only nightly rung

### LINKED ISSUE #2056: Wire tests/soak into curie dev and a scheduled ladder rung (it never runs anywhere today)
### What are you trying to do?

`tests/soak/` (887 lines) documents a real definition-of-done claim that nothing ever exercises. The offline half (`test_harness_unit.py`, 101 lines) runs in default CI. The 332-line cluster scenario has never run in any automated path: `tests/soak/README.md:17` "The scenario never runs in default CI. It is gated on `CURIE_SOAK=1` and skips cleanly with no cluster." No workflow, no `curie dev` subcommand (against the repo's one-entry-point rule in `CLAUDE.md`), no `scripts/` reference, and no nightly-ladder rung invokes it. Its last substantive commit predates the July rename (`126deddd`). It is drifting evidence, not a gate. #21 (chart resilience soak) is closed.

### Proposed change

Wire it rather than archive it (maintainer ruling 2026-08-29: queue the wiring):

- add a `curie dev soak` subcommand (clap surface in `cli/src/main.rs`, handler in `cli/src/commands.rs` or its successor module) that sets `CURIE_SOAK=1`, targets the current cluster, and runs the scenario with `--json` output through `CliOutput`;
- add it as a report-only rung of the nightly graded ladder (`.github/workflows/nightly-graded-ladder.yaml`) so it runs on the k8scratch/cluster tier where a sandbox exists;
- update `tests/soak/README.md` so the gating sentence names the verb and the rung.

No test is deleted or weakened.

### Area

CLI

### Alternatives you considered

- Archive it: rejected; the tests encode a real claim and the repo forbids recommending test deletion.
- Leave it manual-only: that is the current state and it violates the one-entry-point rule.

### Review provenance and autonomy

- Source: 2026-08-29 architecture over-engineering review. Pillar: evaluation-continuous-improvement. Base branch: `main`.
- Behavior-preserving: yes (additive verb, additive rung). ADR required: no. Bonus-drainable: yes.
- Done when: `curie dev soak --help` exists and the command manifest is regenerated; the ladder workflow has a report-only soak rung; `curie dev soak` run against a cluster (k8scratch or the nightly rung) produces the scenario's pass/fail JSON, and with no cluster it skips cleanly with a named reason.

TheConnMan:
Wires `tests/soak/` into an automated path for the first time. Closes #2056.

The 332-line cluster scenario had never run anywhere: no workflow, no `curie dev`
subcommand, no `scripts/` reference invoked it, and its last substantive commit
predated the July rename. It was drifting evidence, not a gate.

## What this adds

- **`curie dev soak`** (`cli/src/main.rs` clap surface, `cli/src/commands.rs`
  handler). Sets `CURIE_SOAK=1`, preflights the kube context, then runs
  `tests/soak/test_soak_resilience.py`. `--namespace`, `--pool` and `--runs` each
  default from the harness's own env vars, so an already-exported harness
  environment is honored with no flag.
- **A `SoakOutput` result family** with a committed `cli/schema/soak.schema.json`
  registered in `cli/schema/index.json`. One family covers pass, fail and skip.
- **A report-only soak rung** on the nightly graded ladder's `ladder-cluster` job,
  reusing the kind cluster that job already installs rather than paying for a
  second install.
- **`tests/soak/README.md`** now names the verb and the rung.

No test is deleted or weakened, and the scenario's own assertions are untouched.

## The skip contract is the load-bearing part

With no reachable context, no such namespace, or no reachable `SandboxWarmPool`,
the verb emits `status: "skipped"` with a **named reason** and exits 0. That is
what lets the nightly rung tell "there was nothing to soak against" from "the soak
found something", instead of a false green or a false red.

The reasons distinguish a genuinely absent object (kubectl `NotFound`) from a
cluster that could not be queried at all (expired credentials, RBAC `Forbidden`,
API outage) — the latter quotes kubectl's own error line. Anything not positively
readable as `NotFound` classifies as operational, never as absent: a wrong
"absent" is a verdict about the cluster, while "could not be checked" is a verdict
about the probe, which is all an ambiguous failure supports.

## Proof against a real cluster

All three skip paths, run against k8scratch, each with its own named reason and
exit 0. And a full run that reached pytest and produced the scenario's real
pass/fail JSON:

```json
{"context":"k8scratch","exit_code":1,"namespace":"curie","pool":"curie-runner-pool",
 "reason":"the soak scenario ran and reported failing assertions (pytest exit 1).",
 "runs":1,"status":"failed"}
```

That `failed` is an honest cluster fact, not a verb defect: this install's warm
pool sits at 0 replicas and its sandbox pods crash-loop on a pre-existing
`PluginBundleError: invalid plugin bundle at /unused`, so `pool_ready` times out.
The cluster was left exactly as found.

## Two things a reviewer should judge, not take on trust

Both were flagged by independent spec-vs-impl and scope-creep reviews of this
commit. Neither is papered over.

**1. The uv acquisition can still redden the nightly.** The soak rung's own work
cannot: it captures the verb's exit status and ends in an unconditional `exit 0`,
the #1603 precedent — not `continue-on-error`, which the retry gate reads as a
retry marker on a step that gates repository code.

An earlier revision wrapped `setup-uv` in a retry trio with an endpoint re-probe.
Both reviews independently rejected it: nothing mandated it, the probe did not
cover the surfaces `setup-uv` actually fails on (GitHub API version resolution,
the python-build-standalone download), and it forced a `RETRY_ALLOWLIST` entry in
an unrelated gate's policy table. It is gone. uv is now a plain pinned `uses:`
step.

That leaves the residual: if `setup-uv` fails, the job goes red. The measurement
that decides whether this matters — `ladder-cluster` **already runs nine
unprotected third-party action steps** (`actions/checkout`,
`actions/download-artifact`, `azure/setup-helm`, `docker/setup-buildx-action`,
four `docker/build-push-action`s, and the kind-cluster action), none with a retry,
each able to redden the nightly on a third-party outage. This adds a tenth at the
same risk level rather than a new class of failure. **If you want zero red-risk,
say so and I will move acquisition inside the `exit 0` body** (a pinned
`python3 -m pip install uv==...`), which closes it at the cost of an acquisition
path unlike every other job's.

**2. The cluster proof shows wiring, not phase execution.** The run above reached
pytest against a real cluster and returned the scenario's real result object — but
`exit 1` there is also what a `pool_ready` fixture timeout yields, and that is what
happened. Both k8scratch installs have a warm pool whose sandbox pods crash-loop on
a pre-existing, unrelated `PluginBundleError: invalid plugin bundle at /unused`, so
no pool on this box can reach ready replicas. I scaled each pool up to try, saw the
same crash-loop, and restored both to 0.

So this proves verb → preflight → pytest → cluster end to end. It does **not**
prove the soak's phases (concurrent threads, batch burst, mid-run kill, resume)
ever executed. That needs a cluster with a healthy warm pool — the nightly rung
itself, or a fixed install. Calling that out rather than letting a green-looking
`failed` JSON imply more than it does.

## Follow-ups deliberately left out (this branch is additive-only)

- `soak_capture` duplicates the private `capture` in `cli/src/doctor.rs:2646`.
  Worth hoisting into `ops.rs` beside `run_capture`/`plain` — and this branch's
  copy is the better one, since it preserves the spawn error `doctor.rs` discards.
- The namespace probe could build its argv from `ops::namespace_get_cmd` so the two
  cannot drift on flags.
- The retry gate has no notion of a rung that cannot fail its job, which is why the
  endpoint re-probe above had to be invented. A third eligibility category would let
  the next such rung use `continue-on-error` on both attempts.

## Verification

`cli/` suite green in debug (including 16 new `dev_soak` tests), `cargo fmt --check`,
`cargo clippy -- -D warnings`, ruff, mypy, retry gate + doclint, release gate, and
the UI lint/typecheck/vitest suite. Both command manifests regenerated via the
documented commands and verified idempotent.

Because this touches `.github/workflows/**`, the full Python suite was run rather
than a per-package subset — it caught four gates a focused run would have missed
(two retry-gate rules, two doclint count claims), all fixed here.

### DIFF
--- a/.github/workflows/nightly-graded-ladder.yaml
+++ b/.github/workflows/nightly-graded-ladder.yaml
@@ -453,7 +453,231 @@ jobs:
           CURIE_CREDENTIALS: ${{ secrets.OPENROUTER_API_KEY }}
           CURIE_MODEL: ${{ inputs.model || 'z-ai/glm-5.2' }}
         run: bash cli/scripts/e2e-ladder.sh
-      # Surface why on any failure above: pod states, recent events, and the
+      # --- Soak rung (REPORT ONLY, #2056) -----------------------------------
+      #
+      # `tests/soak/` had never run in any automated path: it needs a STANDING
+      # cluster with a claimable warm pool, and no CI job had one to lend it.
+      # This job does -- the kind cluster installed above is still up and its
+      # runner-prewarm daemonset has already rolled out -- so the soak runs
+      # HERE rather than paying for a second cluster install in a job of its
+      # own. That is what #2056 means by "the cluster tier where a sandbox
+      # exists".
+      #
+      # The rung is report only, on the #1603 precedent
+      # (`cli/scripts/e2e-ladder.sh`, the live cluster eval): it prints its
+      # result into the run log and the job summary so a reader can act on it,
+      # but nothing it observes can turn this nightly red. A soak is a
+      # load/chaos probe on a shared 2-CPU GitHub runner; graduating it to a
+      # gate needs a stable baseline first, and a flaky gate on the nightly
+      # would train the skim reflex on a workflow whose other rungs are real
+      # gates.
+      #
+      # HOW it is report only is load bearing. The whole rung is ONE step that
+      # captures every exit status inside the shell and ends in an
+      # unconditional `exit 0`, exactly the shape #1603 uses. It is
+      # deliberately NOT `continue-on-error: true`. The retry-eligibility gate
+      # (`tools/ci-retry-gate/tests/test_retry_eligibility.py`) draws a closed
+      # world around `continue-on-error`: it reads that marker as a retry, and
+      # a retry is permitted only on a step that acquires a third-party
+      # dependency before any repository code runs, never on one that builds,
+      # tests or gates this repository's code. A rung driving this repository's
+      # own soak harness is exactly what that world exists to exclude, so
+      # blanket-swallowing its failure would be the gate's central prohibition
+      # rather than an exemption from it. `exit 0` says the same thing without
+      # asking CI to swallow anything: the step really did succeed at its job,
+      # which is to report.
+      #
+      # Placement is deliberate: after the graded cluster rung (so the soak
+      # never perturbs the pods the graded turn is asserting on) and before the
+      # diagnostics dump (so a soak-visible break still gets pods, events and
+      # worker logs in the same run).
+      #
+      # The uv acquisition is the one part of the rung that stays outside that
+      # step: `uv sync` needs uv on PATH before the rung's own shell starts.
+      # It is a plain, unprotected `uses:` -- no `continue-on-error`, no
+      # `id:`, no endpoint probe, no backoff and no `(retry)` copy -- because
+      # that is exactly how this job already acquires every other third-party
+      # action it depends on: `actions/checkout`, `actions/download-artifact`,
+      # `azure/setup-helm`, `docker/setup-buildx-action`, four
+      # `docker/build-push-action` steps and the kind-cluster action, none of
+      # them retried. A retry here would buy a marginal reduction in a failure
+      # mode this job already accepts nine times over, and the price is an
+      # entry in an unrelated gate's policy table
+      # (`RETRY_ALLOWLIST` in `tools/ci-retry-gate/tests/test_retry_eligibility.py`)
+      # for a step that is not the soak rung itself. No `uv lock --check` here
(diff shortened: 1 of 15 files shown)


TheConnMan:
Closing unmerged. Maintainer decision: the soak and chaos suite is platform-owner testing rather than project test surface, so it should live outside this repo instead of gaining a `curie dev` entry point and a scheduled rung here.

Withdrawing the wiring (the `curie dev soak` verb, its result schema, and the nightly rung) rather than landing it. Relocating the suite itself is tracked separately.

PR #2089, opened 2026-08-29
docs: label channel-ingress CLEAN and tag every parity seam with its enforcement

### LINKED ISSUE #2053: Label enforcement in the seam catalog and the parity list (channel-ingress is filed SOFT/1 with a shipped Protocol)
### What are you trying to do?

Two contract catalogs mix enforced and remembered rows indistinguishably, so a coding agent pays attention to every row as if it were load-bearing.

1. `docs/interfaces.md:18` files channel-ingress as `| SOFT | 1 |` while its own file and code say otherwise: `apps/worker/src/curie_worker/reply_sink.py:114` declares `class ReplySink(Protocol):`, `HttpReplyAdapter` ships as a second adapter behind a kind-dispatching `ReplySinkRouter`, and `docs/interfaces/channel-ingress/INTERFACE.md:33` admits "the port is the wire + Protocol contract" while its frontmatter still says `impls: 1`. By the table's own kind legend (CLEAN = a real Protocol draws the line in code) this is a description error, not a request to build a port: the port already exists. `origin/next` still carries the same row.
2. `AGENTS.md:340-360` lists 13 parity-seam pairs with identical typographic authority, but they split into three tiers: four vector-frozen pairs with two-sided tests (`tests/vectors/model-credential-forwarding.json`, `approval-action-ids.json`, `thread-reset-set.json`, `eval-memory-isolation.json`), roughly four enforced by construction or indirectly, and roughly four prose-only, e.g. `AGENTS.md:352` "- `local up` vs `local down` compose profile sets." (a pair that already shipped a bug: down no-ops on profiled services) and `AGENTS.md:351` "- CLI-side vs API-side input validation (validate at the API/persistence boundary, mirror in the CLI)."

### Proposed change

- Correct the channel-ingress kind/impls row and the INTERFACE.md frontmatter to describe what is built (a Protocol at egress with two adapters, a second wire producer at ingress). Regenerate the seam table via `bash scripts/check-docs.sh`.
- Add one enforcement column to the AGENTS.md parity list: `vector: <file>` / `gate: <test or CI job>` / `by construction: <shared module>` / `convention`. The list becomes self-auditing; no row is removed.

Explicitly NOT in scope (dropped by the review's challenger): relabeling telemetry-otel (SOFT explicitly includes a deliberate wire payload by the legend's definition) and triggers (its INTERFACE.md counts 3 implementations and then explicitly catalogs the further wake paths as bypassing them; it is internally consistent).

### Area

Docs

### Alternatives you considered

- Leave it: the weekly seam-audit scores INTERFACE.md dirtiness but does not check the catalog's kind/impls fields, so the drift persists.
- Promote channel-ingress to a frozen contract: not what the evidence supports; this is a label fix.

### Review provenance and autonomy

- Source: 2026-08-29 architecture over-engineering review. Pillar: connectivity-channels. Base branch: `next` (docs/interfaces.md differs across trains; the forward merge carries it to main).
- Behavior-preserving: yes (docs only). ADR required: no. Bonus-drainable: yes.
- Done when: `bash scripts/check-docs.sh` is green; the regenerated seam table shows the corrected channel-ingress row; every one of the 13 parity rows in AGENTS.md carries an enforcement value; no row deleted.

TheConnMan:
Closes #2053

Two contract catalogs mixed enforced and remembered rows indistinguishably, so an agent reading them paid equal attention to every row. This makes both self-describing.

## 1. channel-ingress was filed `SOFT | 1` against a shipped Protocol

`apps/worker/src/curie_worker/reply_sink.py:124` declares `class ReplySink(Protocol)`, with `SlackReplyAdapter` and `HttpReplyAdapter` behind the kind-dispatching `ReplySinkRouter` (`reply_sink.py:395`), and the Rust CLI mints the `QueuedTurn` wire payload as a second ingress producer (`cli/src/queue.rs`). By the catalog's own kind legend (`docs/interfaces.md:41`, "CLEAN = a real `Protocol`/typed port class draws the line in code") that is CLEAN. This is a description fix, not a request to build a port — the port already exists.

Front-matter now reads `kind: CLEAN` / `impls: 2 reply adapters behind the ReplySink port (Slack, HTTP) + a second wire ingress producer (Rust CLI)`. The seam table and the per-doc header blockquote were regenerated via `bash scripts/check-docs.sh`, not hand-edited. `grade: B-` is unchanged — swap-readiness is a separate judgement and `docs/architecture-vision.md` still grades Communication B-.

## 2. Every parity-seam bullet in AGENTS.md now carries one enforcement tag

The 13 pairs had identical typographic authority while splitting three ways. Each bullet gains one trailing tag — `vector:` / `gate:` / `by construction:` / `convention` — so the prose-only pairs are visible as such. No bullet removed, none reworded (verified: 13 on `origin/next`, 13 here, identical opening text).

| tag | count |
|---|---|
| `vector:` | 4 |
| `gate:` | 3 |
| `by construction:` | 3 |
| `convention` | 3 |

## Reviewer findings, all fixed

A read-only cross-engine (Codex) adversarial pass over the diff raised three, each confirmed against the tree before fixing:

- **High** — the `real SDK vs fake model session` row was first tagged `gate: runner/tests/test_conformance.py`. That overclaims: `runner/src/curie_runner/conformance.py:24` builds `SessionRunner(session_factory=FakeModelSession)` and never instantiates the real `ClaudeAgentSession`, so it does not compare the two paths. Retagged `by construction: runner/src/curie_runner/adapter.py::ModelSession` — which is what actually holds them together, both being `session_factory` values for the same `SessionRunner`.
- **Medium** — the legend defined `convention` as "nothing enforces it", which was too strong: all three such rows have narrow shared-code or partial-gate slices (e.g. `cli/tests/observability_query.rs:223` gates local/cluster grammar for the observability leaves only). Softened to "no gate covers the pair as a whole". Judgement call worth flagging: I kept all three as `convention` rather than promoting one to a qualified `gate:`, because naming a partial gate on one row and not the other two would read as coverage the pair does not have.
- **Medium** — the Known-leakage bullet still said there is "no ... second adapter", contradicting the corrected count. Reworded to acknowledge `HttpReplyAdapter` ships while keeping the leak it reports (no adapter framework, no second adapter proving conformance, #27).

Codex separately confirmed CLEAN is correct, and that the four vector tags, the stream-consumer and release-compose `by construction` tags, both `cli/src/local.rs` profile gates, and the API field-parity gate all check out.

Explicitly out of scope, per the issue: telemetry-otel and triggers were not relabelled. Verified untouched.

## Verification

- `bash scripts/check-docs.sh` — green (generated regions drift-free, every citation resolves)
- `uv run pytest tools/doclint/tests -q` — 135 passed
- every artifact named in a tag exists, and each named test function resolves (`down_passes_every_up_profile`, `compose_file_declares_core_and_full_profiles`, `test_runner_passes_aci_conformance`)

Docs only; no code changed.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -326,37 +326,53 @@ rather than silently diverging.
 
 Sibling-path drift is the dominant historical bug class here: logic or hardening
 lands on one side of a structural seam while its twin keeps the old behavior.
-Known seam pairs:
+Known seam pairs, each tagged with what actually enforces it -- `vector:` (a
+frozen two-sided vector file), `gate:` (a test or CI job), `by construction:` (a
+shared module both sides route through), or `convention` (no gate covers the pair
+as a whole; remembered only):
 
 - worker vs CLI credential forwarding -- `_SDK_PASSTHROUGH_ENV`
   (`apps/worker/src/curie_worker/sandbox/docker.py`) and the CLI picker
   (`cli/src/commands.rs`) can't share code across Python/Rust, so they are frozen
   together in `tests/vectors/model-credential-forwarding.json`.
+  [vector: `tests/vectors/model-credential-forwarding.json`]
 - dispatcher vs CLI approval action ids -- the approval-card action-id constants
   (`apps/dispatcher/src/curie_dispatcher/approval_actions.py`) and the CLI's
   `APPROVE_ACTION_ID_PREFIX` (`cli/src/chat.rs`) can't share code across
   Python/Rust, so they are frozen together in
   `tests/vectors/approval-action-ids.json`.
+  [vector: `tests/vectors/approval-action-ids.json`]
 - API vs worker vs CLI thread-reset SET -- `THREAD_RESET_SET` /
   `THREAD_RESET_INFLIGHT_SET` (`apps/api/src/curie_api/threadreset.py`,
   `apps/worker/src/curie_worker/consumer.py`) and the CLI's `THREAD_RESET_SET`
   (`cli/src/queue.rs`) can't share code across Python/Rust. Operator
   `reset-thread` and `local`/`cluster` eval (#1534) both SADD the same set.
   Frozen in `tests/vectors/thread-reset-set.json`.
+  [vector: `tests/vectors/thread-reset-set.json`]
 - worker vs CLI eval-isolate prefix -- `EVAL_ISOLATE_THREAD_PREFIX`
   (`apps/worker/src/curie_worker/binding.py`) and the CLI copy
   (`cli/src/queue.rs`, stamped in `cli/src/message.rs::run_eval_turns`)
   cannot share code across Python/Rust, so they are frozen together in
   `tests/vectors/eval-memory-isolation.json`.
-- real SDK vs fake model session in the runner (`FakeModelSession`, `runner/src/curie_runner/fake.py`).
-- runs lane vs eval lane stream consumers (`apps/worker/src/curie_worker/consumer.py` vs `eval/stream.py`, both on the shared `stream_consumer.py`).
-- CLI-side vs API-side input validation (validate at the API/persistence boundary, mirror in the CLI).
+  [vector: `tests/vectors/eval-memory-isolation.json`]
+- real SDK vs fake model session in the runner (`FakeModelSession`,
+  `runner/src/curie_runner/fake.py`).
+  [by construction: `runner/src/curie_runner/adapter.py::ModelSession`]
+- runs lane vs eval lane stream consumers (`apps/worker/src/curie_worker/consumer.py`
+  vs `eval/stream.py`, both on the shared `stream_consumer.py`).
+  [by construction: `apps/worker/src/curie_worker/stream_consumer.py`]
+- CLI-side vs API-side input validation (validate at the API/persistence
+  boundary, mirror in the CLI). [convention]
 - `local up` vs `local down` compose profile sets.
+  [gate: `cli/src/local.rs::down_passes_every_up_profile`]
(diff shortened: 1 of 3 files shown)


TheConnMan:
Rebased onto current `next` (`826e417c`) and force-pushed.

The first push was cut from a stale `origin/next` ref — `git fetch origin next` here updates `FETCH_HEAD` but not `refs/remotes/origin/next`, so the worktree was based 104 commits behind. Re-verified everything against the real base rather than trusting the clean rebase:

- the precondition still holds: `docs/interfaces.md:18` on current `next` is still `| SOFT | 1 |`
- `reply_sink.py:124` `class ReplySink(Protocol)`, `:224` `HttpReplyAdapter`, `:395` `ReplySinkRouter`, `slack_sink.py:258` `SlackReplyAdapter` — all still present
- every artifact and symbol named in an enforcement tag still resolves (`down_passes_every_up_profile`, `compose_file_declares_core_and_full_profiles`, `adapter.py::ModelSession`)
- 13 parity bullets, 13 tags, diff is docs-only (3 files)
- `bash scripts/check-docs.sh` green; `uv run pytest tools/doclint/tests -q` 135 passed

For the record on the pre-rebase run: the only red check was `Python (ruff + mypy + pytest)`, failing `apps/worker/tests/kernel/test_consumer.py::test_timed_out_liveness_renewal_retries_before_lease_expiry` (`assert 0 >= 2` on a 2-second `time.monotonic()` deadline). That test does not exist on the branch's own base and this diff touches no Python — it came from the base and is a timing flake, not this change. The re-run on the current base will confirm.

PR #2152, opened 2026-08-31
Fix receipt classification for read-only tools

### LINKED ISSUE #2130: "What I changed" lists every connector read, because readOnlyHint is never read
Every connector call a turn makes is listed under **"What I changed"**, including pure reads. A person reading that receipt is told the bot changed things it only looked at.

## What it looks like

A turn that read the cluster and a release feed, and changed nothing:

```
What I changed:
• called ToolSearch — non-idempotent tool completed
• the newest published release of curie-eng/curie is v0.8.1. This is what is
  available, NOT what this install is running -- … — non-idempotent tool completed
• called mcp__kubernetes__resources_list — non-idempotent tool completed
```

An earlier read-only triage turn listed nine of them, `mcp__kubernetes__namespaces_list` and `mcp__grafana__list_datasources` among them, each labelled "non-idempotent tool completed".

## Why

`SideEffectClassifier` decides by NAME against an allowlist:

```python
def is_side_effecting(self, tool_name: str) -> bool:
    return tool_name not in self._idempotent
```

and that allowlist is the harness's own declared read-only tools plus the platform's:

```python
classifier=SideEffectClassifier(readonly_tools=harness.readonly_tools)
```

Connector tools arrive from MCP servers at runtime carrying `annotations.readOnlyHint`, which is exactly this question already answered by the connector that published the tool. Nothing reads it — `grep -rn readOnlyHint runner/src apps/worker/src` returns nothing.

So every connector tool is unknown to the allowlist, and therefore side-effecting.

## The failing-closed part is right; the reporting is not

`side_effects.py` explains the allowlist deliberately: a denylist "fails open (a new non-idempotent tool is treated as safe)". That reasoning is sound for the **retry** decision it was written for — a failed turn that may have executed a write must not be retried blindly, and over-classifying is the safe direction there.

It is the wrong default for a **receipt**. "What I changed" is a claim to a person about what happened to their systems, and it is currently making that claim about reads. Over-reporting there does not fail safe; it trains people to skim a section that exists to be read.

## Two things that could be separated

1. **Retry safety** — keep failing closed. A tool the platform cannot classify should stay side-effecting for the no-retry rule.
2. **The receipt** — should list what actually changed. A connector that declared `readOnlyHint: true` has answered the question, and honouring that annotation costs nothing: it is per-tool, published by the server, and already on the wire.

Those can use the same signal with different defaults: unknown → side-effecting for retry, unknown → still listed for the receipt, but `readOnlyHint: true` → excluded from both.

## A second, smaller thing this surfaced

The receipt renders a connector's `summary` in place of the tool name when the reply has one (`receipt._described`). That is intended — a connector's own sentence beats a bare name — but combined with the above, a read's summary appears as a line item under "What I changed", which is how the example at the top reads so strangely.

Worth deciding whether `_described` should apply only to calls that actually changed something.

TheConnMan:
Fixes #2130.\n\nHonors explicit MCP read-only annotations for receipt/retry classification while preserving fail-closed unknowns and operator-gated tool receipts.

### DIFF
--- a/docs/approvals.md
+++ b/docs/approvals.md
@@ -35,9 +35,13 @@ that may write — any tool not explicitly `readOnlyHint=true`, including an unk
 unreachable surface — or when an explicit actionable approval gate exists. A bundle with
 no MCP tools, or only explicitly read-only MCP tools and no actionable gate, does not
 carry it: explain that the bundle cannot perform the action instead of fabricating a
-remediation request. `readOnlyHint` controls this tool advertisement only; it is not
-authorization and does not change gates or tool execution. Publication's dedicated
-approval flow and the state tools remain independent.
+remediation request. `readOnlyHint` is not authorization and does not change gates or
+tool execution. When a live probe explicitly reports it as `true`, Curie also adds that
+MCP tool's SDK-visible name to the read-only classifier: it emits no side-effect flag,
+does not take the no-retry-after-side-effects path, and creates no receipt line. A
+missing or `false` hint, an unknown surface, or a failed probe remains potentially
+write-capable and is classified conservatively. Publication's dedicated approval flow
+and the state tools remain independent.
 
 **Permission gate.** Configuration marks a tool approval-required, and the runner denies
 the call through the SDK's `can_use_tool` callback before it runs. The denied call never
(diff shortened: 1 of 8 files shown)


TheConnMan:
@TheConnMan review requested. This PR is intentionally unmerged and has not been self-approved.

Issue #2190, opened 2026-09-01
Pin Langfuse images to prevent floating-tag runtime drift


TheConnMan:
## Problem

The chart and Compose stack use floating `langfuse/langfuse:3` and
`langfuse/langfuse-worker:3` tags. On 2026-09-01 those tags moved from
Langfuse 3.225.5 to 3.225.6 while Curie CI was running.

Fresh runners pulling 3.225.6 reproducibly fail the real consumers:

- the Compose parity rung returns 500 from the first-party observability query;
- the Kubernetes delayed-Postgres runtime proof leaves ClickHouse migration 39
  dirty and Langfuse web restarts instead of becoming Ready with zero restarts.

The same Curie candidate passes both paths with the cached 3.225.5 images.

## Acceptance criteria

- Compose and the Helm chart use the same explicit Langfuse web/worker version.
- No Curie default uses the floating major-only Langfuse tag.
- An executable assertion proves both rendered consumers carry the pin and
  rejects a floating-tag mutation.
- The full Python baseline, skill/local parity rung, local-release rung, and
  real cluster delayed-readiness proof pass from clean state.

TheConnMan:
Follow-up from a fresh install of chart 0.8.3 on an EKS cluster with two t3.medium nodes: the pin here holds, but 24.8 is the reason it cannot be advanced. Migration 39 is rejected by ClickHouse 24.8 (`Only literals can be skip index arguments`) and by 25.8 (`Expected literal`), leaving `Dirty database version 39` both times; on 25.12 all 46 migrations apply. Langfuse's own docker-compose at v3.225.7 pins clickhouse-server:25.12. Filed as #2210 so the coupling between `langfuse.image.tag` and `clickhouse.image.tag` is tracked rather than rediscovered at the next bump.

PR #2189, opened 2026-09-01
Forward merge stable v0.8.3 into next

TheConnMan:
## Summary

Forward merge stable `main` v0.8.3 into `next` so the feature train carries Alembic head `0038`, authenticated approval principals, the stable cluster-up guards, and the released chart/runtime fixes. The conflict resolution preserves both branches' newer behavior: `next` keeps structured replay, built-in publication, and WebSearch control, while `main` contributes MCP capability probing, truthful provider/tool phase telemetry, and the stable approval-principal path.

## Related issue

Ref #2169

## Fix pin verification

This is a release-train forward merge and does not close a bug-labeled issue. The stable fixes retain their existing pins; the merge-specific tests assert both sides of the resolved runner and split-CLI behavior.

## End-to-end verification

| Tier | Required / n/a | Reason | Mode (fake / live) | Command and observed outcome |
| --- | --- | --- | --- | --- |
| skill | required | Runner session, approval-server, replay, and telemetry conflicts reach the skill turn loop. | fake | `CURIE_E2E_TIERS=skill,local CURIE_BIN=/home/theconnman/.cargo/shared-targets/agentos/release/curie cargo run --quiet -- dev e2e-ladder` on `32caa57b7fa8f6f8fc95a8c479b751af168a6c59`: skill up/status/message/eval/down passed with bundle digest `639a38e7fa9e...`; source mutation stayed out of the running snapshot, re-up changed the digest, and the occupied-runner refusal named and exercised recovery. |
| local | required | The merge reaches API, worker, dispatcher, runner, CLI, and console wiring. | fake | Same command and commit: cold-start deploy/message/down passed against the full backing stack; a forced non-retryable runner failure produced a classified failure and ERROR trace/log, then a restored worker completed a healthy turn; unknown-trace and unavailable-API negatives returned distinct semantic exits and one `{error,fix}` object. |
| local-release | required | Stable release identity and the generated release compose changed across the train. | fake | `CURIE_E2E_TIERS=local-release CURIE_BIN=/home/theconnman/.cargo/shared-targets/agentos/release/curie cargo run --quiet -- dev e2e-ladder` on the same commit with candidate-built images: generated `compose.release.yaml` cold-started, deployed the bundle and connector, completed a worker turn, matched eval-suite identity, and tore down cleanly. |
[...]

PR #2195, opened 2026-09-01
Harden the Langfuse pin gate and the dev-stack failure dump

TheConnMan:
## Summary

The Langfuse `3.225.5` pin that ended the repo-wide CI outage has already landed on both trains (#2191 into `main`, and #2189's forward merge into `next`). This PR is the leftover hardening from diagnosing that outage: three gaps around the pin, none of which the pin itself closed. It is deliberately small and each hunk can be dropped independently.

Root cause recap, since it is not what a floating-tag guess alone gives you: upstream rebuilt `langfuse/langfuse:3` **and** the minor tag `3.225` **in place** at `2026-09-01T14:50:59Z` to ship v3.225.6. No `3.225.6` tag was ever published, so searching for a new tag finds nothing — the mutation is only visible by digest. `3.225.5` did not move, so it is the only safe pin; **`3.225` would not have been**.

## Related issue

Ref #2190

## What this changes

**1. The dev-stack failure dump printed nothing at all.** `.github/workflows/ci.yaml` ran `docker compose -f compose.dev.yaml logs` with no profile and no service names. Every dev-stack service sits behind the compose `full` profile, so that command matched no services and emitted zero lines. This is why the outage was diagnosed with no container logs whatsoever — job `99917342054`'s dump step is empty. Verified against a live stack: without the flag, no output; with `--profile full`, all 8 services (`postgres`, `valkey`, `clickhouse`, `rustfs`, `rustfs-init`, `langfuse-web`, `langfuse-worker`, `otel-collector`).

**2. helm-ci could skip the job that runs the pin gate.** `compose.dev.yaml` was not in the workflow's `paths` filter, yet the pin gate renders it as the second shipping surface. A pull request floating the Langfuse tag *only* in Compose would have matched no path, skipped the job, and never run the gate that exists to catch precisely that.

**3. The gate did not cover every Langfuse image reference.** It asserted the two application images were pinned and rejected a literal `:3` anywhere, but did not require *every* Langfuse reference to equal the reviewed version. Both Deployments also run a `wait-for-clickhouse` initContainer on a Langfuse image:

```
curie-langfuse-web     initContainers  wait-for-clickhouse   langfuse/langfuse:3.225.5
curie-langfuse-web     containers      langfuse-web          langfuse/langfuse:3.225.5
curie-langfuse-worker  initContainers  wait-for-clickhouse   langfuse/langfuse-worker:3.225.5
curie-langfuse-worker  containers      langfuse-worker       langfuse/langfuse-worker:3.225.5
```

So a wrong-but-not-floating version in an initContainer (say `langfuse/langfuse:3.226.0`) passed, while being pulled and executed *before* the checked application container. That is the exact class that caused this outage — a version change, not specifically a `:3`. The gate now parses the render and requires every container and initContainer whose image starts with `langfuse/` to equal its Deployment's reviewed image, name-independently, with anti-vacuity checks so a vanished container or an empty render fails instead of silently passing.

Plus a comment correction on `charts/curie/values.yaml`: the landed comment does not say that `3.225` moved too, which is the actionable part. I dropped the "left ClickHouse migration 39 dirty" clause because this investigation did not reproduce it.

## Verification

| Check | Result |
| --- | --- |
| `bash charts/curie/ci/langfuse-image-pin-assertions.sh` | `chart pins 4 Langfuse image references (containers + initContainers) and Compose pins web/worker to 3.225.5`; both negative controls fire; exit 0 |
| Gate can actually fail | `EXPECTED_VERSION=9.9.9` exits 1 and names all six mismatches (4 chart + 2 compose) |
| `cargo test --test chart_check` | **7 passed, 0 failed** — includes the full 37-script sweep and `helm_ci_and_the_chart_ci_directory_cannot_diverge` |
| `bash charts/curie/ci/render-assertions.sh` | PASS |
| `helm lint charts/curie` | 0 failed |
| Both workflows | `yaml.safe_load` OK; no duplicated script invocations |

Note on hunk 2: the divergence test scans this workflow for `charts/curie/ci/*.sh` references, so the new comment deliberately refers to the gate in prose rather than by path — spelling out the path made the test read it as a duplicated step.

## Not behavior-bearing

No tier applies. This changes CI wiring, a CI diagnostic command, an assertion script, and a comment. It ships no runtime behavior: the image pin itself is already on `next` and is untouched here.

- [x] Or: this change is not behavior-bearing, so no tier applies, and the summary says why.

## Checklist

- [x] Tests pass for the area I touched.
- [x] Docs updated if behavior changed (n/a; the chart README image table already reflects the pin).
- [x] No new ADR is required.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -170,9 +170,12 @@ jobs:
           --event "$GITHUB_EVENT_PATH"
           --curie cli/target/release/curie
           --ref HEAD
+      # Every dev-stack service sits behind the compose `full` profile. Without
+      # --profile full, `docker compose logs` matches no service and prints
+      # nothing, so a real outage gets diagnosed with no container logs at all.
       - name: Dump dev stack logs on failure
         if: failure()
-        run: docker compose -f compose.dev.yaml logs --no-color --tail=200
+        run: docker compose --profile full -f compose.dev.yaml logs --no-color --tail=200
 
   rust:
     name: Rust (fmt + clippy + test)
--- a/.github/workflows/helm-ci.yaml
+++ b/.github/workflows/helm-ci.yaml
@@ -28,6 +28,11 @@ on:
       - "apps/worker/src/curie_worker/bundle_store.py"
       - "uv.lock"
       - "pyproject.toml"
+      # The Langfuse image-pin assertion step below renders compose.dev.yaml
+      # as the second shipping surface for the pinned Langfuse version. A PR
+      # that floats the tag only there would match no path here, skip this
+      # job, and never run the gate that catches exactly that.
+      - "compose.dev.yaml"
 
 permissions:
   contents: read
--- a/charts/curie/ci/langfuse-image-pin-assertions.sh
+++ b/charts/curie/ci/langfuse-image-pin-assertions.sh
@@ -1,70 +1,241 @@
 #!/usr/bin/env bash
+#
+# Issue #2190: keep the shipped Langfuse runtime on one reviewed version.
+#
+# On 2026-09-01T14:50:59Z upstream shipped v3.225.6 by rebuilding the
+# floating `:3` tag and the minor tag `3.225` in place onto its new digest,
+# rather than publishing a `3.225.6` tag — so a minor-tag pin would have
+# moved too (the exact patch tag `3.225.5` was unaffected). In CI, the
+# langfuse-web container reported healthy but never served
+# /api/public/health, timing out the bring-up gate. This gate renders both
+# user-facing consumers, requires one exact version, and proves a
+# floating-tag mutation is rejected.
+#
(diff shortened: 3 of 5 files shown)


TheConnMan:
Heads-up on an unrelated red check: **pnpm audit (apps/ui)** fails here, and not because of this PR — this diff touches no JS, UI, or lockfile files.

A Browserslist advisory (`browserslist <=4.28.6`, 2 high: unbounded memory growth, and an uncaught crash / prototype write) was published between 17:15Z and 17:36Z today. PRs #2188 and #2147 passed this check at 17:15Z; this PR's run at 17:36Z failed. Any re-run of those PRs will now fail the same way, so this is a fresh repo-wide blocker rather than anything branch-specific.

Our lockfile pins `browserslist@4.28.6`. The fix is a bump to **4.28.8** (published 2026-08-08), which is outside this PR's scope and wants its own change.

Issue #20, opened 2026-07-07
Transcript persistence across unplanned runner restarts


TheConnMan:
Suspend/resume rehydrates from `AGENTOS_HISTORY_REF` (ADR-0003), but an unplanned runner-pod restart mid-thread loses the in-pod transcript: only an explicitly suspended session has a resume handle.

Persist enough turn history that a restarted sandbox rehydrates the live thread, not just an explicitly resumed one.

Issue #1479, opened 2026-08-13
A fix must ship a test that fails when the fix is reverted


TheConnMan:
Across seventeen consecutive review windows the dominant pattern in this repo is that the fix or feature for a prior gap is where the next finding hides. Recent instances: #1445 wired the worker object store and #1468 rewrote the same assertions eleven days later, deleting a guard #1451 had certified as working; the ADR-0101 gate was amended by ADR-0103, whose own claim then diverged from its implementation immediately.

The common mechanism is that fixes land without a test that would fail if the fix were reverted, so nothing holds the ground the fix took.

## What is wanted

A repeatable, mechanical check that a fix is actually pinned: revert the fix hunk, confirm a named test goes red, restore. Reading the test is not sufficient evidence and never has been.

## Acceptance criteria

1. A documented procedure, runnable as a `curie dev` subcommand per the one-entry-point rule in CLAUDE.md, that takes a commit or PR and a test selector, reverts the non-test hunks of that change in a scratch worktree, runs the selector, and reports whether it went red.
2. The command exits non-zero when the selector still passes with the fix reverted, since that is the failure being detected.
3. It refuses clearly, rather than passing, when the change has no test hunks at all or when the revert does not apply cleanly.
4. It works for the three shapes this repo actually ships: a Python change under `apps/` or `packages/`, a Rust change under `cli/`, and a chart assertion script under `charts/curie/ci/`.
5. A test of the command itself proves both directions on a real in-repo example: a properly pinned fix reports pinned, and a fix whose test asserts presence rather than behavior reports unpinned.
6. `AGENTS.md` states that a fix PR is expected to be verifiable this way, and names the command.

## Notes

Scope this to the local verification command only. Do not wire it into CI as a required check in this issue, and do not attempt to infer which hunks are the fix versus the test beyond the straightforward test-path heuristic. The naive correct version is the deliverable.

Stop after the naive correct solution. Do not add abstractions, extra files, or flexibility not requested.

TheConnMan:
Completed by merged PR #1777. Closing manually because the implementation landed on the non-default release-train branch or the closing reference did not auto-fire.

PR #2226, opened 2026-09-02
Scope the CI image layer cache per image

jw3329:
## Summary

All ten legs of the `images` matrix run concurrently and every one of them wrote `cache-to: type=gha` with no scope, putting them in a single shared cache namespace. They build different Dockerfiles, so there was nothing to share there: they only contended for the same reservation. This names `scope=${{ matrix.name }}` on both `cache-from` and `cache-to`, giving each leg its own namespace. The ladder jobs further down the same workflow already build this way (`scope=ladder-runner`); this matrix was the one place left on the default.

Unscoped, the cache export outgrew the build it was meant to save. On CI run [33690551116](https://github.com/curie-eng/curie/actions/runs/33690551116) the `sre-bot-tempo` leg built in 80.5s and then spent 1057.7s in `#20 exporting to GitHub Actions Cache` writing a single layer:

```
#15 [linux/arm64 5/6] RUN pip install --no-cache-dir -r requirements.txt
#15 DONE 80.5s
...
#20 exporting to GitHub Actions Cache
#20 writing layer sha256:a1a26f26... 1057.7s done
#20 DONE 1063.3s
```

That is a 19 minute job on a 2 minute image, and it sat on the critical path of a 21m54s run. The same run logged the contention directly: `Failed to save: Unable to reserve cache with key docker.io--tonistiigi--binfmt-latest-linux-x64, another job may be creating this cache`.

Scope on this. The stall is a shared-namespace reservation problem, which scoping addresses. If a future run still shows a multi-minute export on one of the four example connectors, the follow-up is to drop `cache-to` for them entirely rather than to widen the scope: their whole build is `FROM python:3.12-slim` plus a two line `pip install`, so an 80s cold build is cheaper than exporting a base image that Docker Hub already serves.

## End-to-end verification

| skill | n/a | Changes only the GitHub Actions layer cache key for the no-push image matrix. No product code, chart, image content, or runtime path is touched. | | |
| local | n/a | Same | | |
| local-release | n/a | Same | | |
| cluster | n/a | Same | | |
| live provider | n/a | Same | | |
| external integration | n/a | Same | | |

The affected surface is CI itself, so the real verification is CI, plus the contract suites that read `ci.yaml`, run at commit `9c0e8868`:

```
uv run pytest release/tests tools/ci-workflow-paths/tests tools/ci-retry-gate/tests tools/e2e-ci-selection/tests tools/fix-pin-ci/tests -q
329 passed in 23.49s
```

Falsifiable negative for the new test: reverting both cache lines to the unscoped `type=gha` and re-running it fails on the `cache-from` comparison (`release/tests/test_authorize.py:1008`), and restoring the scope passes it again. The test also asserts the matrix names are unique, because the scope is only per image if the key it reads is.

The positive proof that the stall is gone is this pull request's own `Build sre-bot-tempo image (no push)` check.

- [x] Or: this change is not behavior-bearing, so no tier applies, and the summary says why.

## Checklist

- [x] Tests pass for the area I touched (see CONTRIBUTING.md for the commands).
- [x] Docs updated if behavior changed. No product behavior changed; the reasoning lives in the workflow comment and the test docstring, next to the thing it explains.
- [x] An ADR is added under `docs/adr/` if this is an architectural decision. Not one: this is a cache key, not a decision about structure.

## What this deliberately does not do

Two other critical-path costs came out of the same profiling, and both are blocked by executable contracts that were written on purpose. They are described in the review thread rather than changed here.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -531,8 +531,21 @@ jobs:
           platforms: ${{ matrix.platforms }}
           push: false
           load: false
-          cache-from: type=gha
-          cache-to: type=gha,mode=max
+          # Scope the layer cache per image. Every leg of this matrix runs at
+          # once, and an unscoped `type=gha` puts all ten of them in ONE cache
+          # namespace, where they contend for the same reservation instead of
+          # sharing anything: the legs build different Dockerfiles, so there is
+          # no cross-image reuse to give up by separating them. Left unscoped,
+          # the export -- not the build -- became the job. On run 33690551116
+          # sre-bot-tempo built in 80.5s and then spent 1057.7s writing a single
+          # layer into the shared namespace, for a 19 minute job on a 2 minute
+          # image, while the same run logged `Unable to reserve cache with key
+          # docker.io--tonistiigi--binfmt-latest-linux-x64, another job may be
+          # creating this cache`. The ladder jobs below already scope their
+          # builds this way (`scope=ladder-runner`); this matrix was the one
+          # place left on the default.
+          cache-from: type=gha,scope=${{ matrix.name }}
+          cache-to: type=gha,mode=max,scope=${{ matrix.name }}
 
   # The worker-local overlay is FROM the published curie-worker image, so it
   # cannot be built purely from the checkout. Build the base worker image
--- a/release/tests/test_authorize.py
+++ b/release/tests/test_authorize.py
@@ -989,6 +989,30 @@ class TestReleaseWorkflowContract:
         )
         assert "Build sre-bot-tempo image (no push)" in authorize_module.REQUIRED_CHECK_NAMES
 
+    def test_image_matrix_scopes_its_layer_cache_per_image(self):
+        """Every leg of the images matrix must name its own gha cache scope.
+
+        All ten legs run concurrently and build different Dockerfiles, so an
+        unscoped ``type=gha`` gives them one shared namespace with nothing to
+        share: they only contend for the same reservation. Unscoped, the cache
+        export outgrew the build it was meant to save -- sre-bot-tempo built in
+        80.5s and spent 1057.7s writing one layer, turning a 2 minute image into
+        a 19 minute job on the critical path.
+        """
+        workflow = yaml.load(CI_YAML.read_text(), Loader=yaml.BaseLoader)
+        image_job = workflow["jobs"]["images"]
+        build_step = next(
+            step for step in image_job["steps"] if step.get("name") == "Build (no push)"
+        )
+
+        assert build_step["with"]["cache-from"] == "type=gha,scope=${{ matrix.name }}"
+        assert build_step["with"]["cache-to"] == "type=gha,mode=max,scope=${{ matrix.name }}"
+
+        # The scope is only per image if the key it reads is unique per leg, so
+        # assert the matrix names are distinct rather than trusting the template.
+        names = [row["name"] for row in image_job["strategy"]["matrix"]["include"]]
+        assert len(names) == len(set(names)), f"image matrix names must be unique: {names}"
+
     def test_observability_stack_assertions_run_in_helm_ci(self):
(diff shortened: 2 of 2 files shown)


jw3329:
### The other two critical-path costs, and why they are not in this PR

Profiling run [33690551116](https://github.com/curie-eng/curie/actions/runs/33690551116) (21m54s) turned up three costs. This PR fixes one. The other two are each blocked by an executable contract that was written deliberately, so they need a maintainer's call rather than a drive-by edit.

**1. The Python job rebuilds the curie binary (213s).**

`Build the current curie binary for fix pin verification` runs a cold `cargo build --release --locked` inside the 20 minute Python job, while `Build the curie release binary` already produces the same binary in ~1m and uploads it as `curie-x86_64-linux-<sha>`, which four other jobs download.

That duplication is on purpose. `tools/fix-pin-ci/tests/test_fix_pin_ci.py::test_ci_keeps_the_required_python_status_and_calls_fix_pin_after_pytest` asserts all three of the obvious fixes shut:

```python
assert "needs" not in job, "the required Python check must not be skippable"
assert not any(_string(step, "uses") == "actions/download-artifact@v8" for step in steps), \
    "the Python job must not download its curie binary"
assert not any(_string(step, "uses") == "Swatinem/rust-cache@v2" for step in steps), \
    "the Python job must build directly without a Cargo cache dependency"
```

Introduced by `881787ef` ("Require fix pin verification in CI", Ref #1479), whose first bullet is "Make enforcement required inside the existing Python status". The gate lives inside an already-required check so it cannot be turned into a separate job that branch protection would let a skip satisfy. Adding `needs:` would also be slower, not faster: it would serialize the binary job in front of the longest job on the graph.

There is a fix that costs nothing the contract is protecting. `check.py` exits at `SKIPPED: no Fix pin declaration` without ever invoking curie, so on the majority of PRs the 213s build produces a binary that is never executed. Gating the build on the body actually carrying a declaration keeps it a direct build inside the required status, with no `needs`, no artifact, and no cargo cache:

```yaml
if: github.event_name == 'pull_request' && contains(github.event.pull_request.body, 'Fix pin:')
```

The test's `PR_CONDITION` regex still matches that `if`, and the `pytest_index < index < gate_index` ordering is unaffected. It does need the contract's assertions extended to cover the new condition, and `check.py` checked for whether it stats the binary path before deciding to skip. Worth doing, but it is a change to a required gate and should be its own reviewed PR.

**2. `pull_request: types: [..., edited]` re-runs all 22 minutes on a body edit.**

I had this down as "guard the heavy jobs with `if: github.event.action != 'edited'`". That is wrong and I am glad the contract caught it. The same test asserts:

```python
assert len(actions) == 4 and set(actions) == {"opened", "synchronize", "reopened", "edited"}, \
    "CI must rerun required checks when code or the pull request body changes"
```

and `tools/ci-workflow-paths/tests/test_workflow_paths.py:610` records why: "`edited` closes the body-only bypass that ci.yaml cannot see."

Worse, the guard would be actively unsafe here. `concurrency.cancel-in-progress` is keyed on the ref, so an `edited` event cancels the in-flight `synchronize` run. Editing the body while CI is running would kill the real run and replace it with one whose heavy jobs are skipped, turning a would-be-failing required check green. No change here without restructuring which checks are required.

**Also worth a look, separately:** in the released-upgrade job, `Nil unsafe helper negative control` ([ci.yaml:1528](https://github.com/curie-eng/curie/blob/main/.github/workflows/ci.yaml#L1528)) takes 489s because it does a full `helm install` of the platform into kind, then upgrade, uninstall and namespace delete, to prove two mutants are rejected. Its first assertion, that the nil-unsafe placement mutant fails to upgrade, is decided at template render time, so `helm template` or `--dry-run` proves the same thing in seconds. The second assertion, that a blank attester persists, does need real release state and should stay. That is a security gate, so it wants a deliberate review rather than an optimization pass.

Issue #2354, opened 2026-09-04
Close the store-existingSecret defect class structurally: one named template plus a generic render assertion


TheConnMan:
## Why

The chart has now had the same defect twice in the same helper. #2052 fixed `CLICKHOUSE_PASSWORD` and `REDIS_AUTH` in `curie.langfuse.env` and enumerated three more refs as the untracked remainder; #2327 fixed those three (`POSTGRES_PASSWORD`, `SALT`, `ENCRYPTION_KEY`).

Both fixes were the same shape: change the specific `name:` lines, then hand-enumerate assertions for the specific keys. Neither made the next instance impossible. Two structural changes would.

## 1. A named template for the store-secret idiom

`.Values.<store>.existingSecret | default (include "curie.secretName" .)` is inlined at roughly nine sites in `charts/curie/templates/_helpers.tpl` (including `curie.env.postgres`, `curie.env.valkey`, and the `curie.langfuse.env` refs), plus three more inline in `charts/curie/templates/langfuse.yaml`.

That is exactly the shape that produces this bug: a new consumer is added by copy-pasting a neighbouring line, and nothing forces it through the idiom — copying the *wrong* neighbour is silent and renders a valid manifest.

A `curie.storeSecretName` template taking the store block and returning that expression would make the correct form the only form on offer. Mechanical, but it touches every existing site, so it wants its own PR and its own render diff.

Note `curie.otlpAuthHeaderSecretName` (`_helpers.tpl`) is deliberately *not* this template — it branches on `otelCollector.otlpAuthHeader` first. Any generic helper has to leave that one alone.

## 2. A generic render assertion

`charts/curie/ci/valkey-existing-secret-assertions.sh` now covers valkey, postgres and the five Langfuse app keys, but it does so by hand-enumerating each key per store. It cannot catch a tenth ref added later that nobody thought to list — which is precisely how #2327's three refs survived #2052.

A generic check would subsume all of it: render with every store's `existingSecret` set to a distinct name, walk every `secretKeyRef` in the rendered manifests, and assert that no ref whose key belongs to a store with `existingSecret` set still points at the chart-managed Secret. The distinct-name-per-store part matters — it also catches a ref cross-wired to the wrong store's Secret, which no current assertion would.

## Not in scope

Prose in `charts/curie/CLAUDE.md` stating the toggle + BYO invariant is not enough on its own: the invariant was written down and violated twice anyway. It should stay as documentation, with the two items above as the actual enforcement.

[...]

Issue #2052, opened 2026-08-29
Langfuse env ignores store existingSecret: a BYO-valkey install silently kills only trace ingestion


TheConnMan:
### Describe the bug

BYO store credentials never reach Langfuse. In `charts/curie/templates/_helpers.tpl`, the Langfuse env block pins `CLICKHOUSE_PASSWORD` to the chart-managed secret (`:610-614`, `secretKeyRef: name: {{ include "curie.secretName" . }}`) and pins `REDIS_AUTH` the same way (`:621-625`), while `valkey.existingSecret` IS honored by the app-services env (`:493-495`) and rustfs four lines up shows the correct `{{ .Values.rustfs.existingSecret | default (include "curie.secretName" .) }}` idiom.

Consequences:

- A supported BYO-valkey install (`valkey.deploy=false` + `valkey.host` + `valkey.existingSecret`) silently kills only Langfuse trace ingestion while every other component stays healthy. This is the exact failure class the same file's rustfs tombstone documents (`_helpers.tpl:637-641`: BYO endpoints hardcoded killed trace ingestion "with the rest of the release healthy, which is why nothing pointed at the cause").
- In `clickhouse.deploy=true` mode, setting `clickhouse.existingSecret` creates split-brain auth: the in-chart server honors it (`templates/clickhouse.yaml:101`) while Langfuse reads the chart secret.
- The chart's own invariant in `charts/curie/CLAUDE.md` (lines 18-24: flipping deploy to false "repoints consumers (Langfuse env, the collector config) at the BYO `host`/`port`/`auth`/`existingSecret` fields") is false as written, and the only CI exercise of the BYO branch asserts manifest absence, not consumer wiring (`charts/curie/ci/clickhouse-logging-assertions.sh:273`).

Adjacent, not duplicate: #1505 (langfuse.deploy=false renders a URL for a missing Service). #1801/#1802 extend existingSecret pairs on the credential side, not store consumption.

### Expected behavior

`REDIS_AUTH` and `CLICKHOUSE_PASSWORD` in the Langfuse env resolve through `<store>.existingSecret | default (include "curie.secretName" .)` exactly like rustfs, and `clickhouse-logging-assertions.sh` (plus a valkey sibling) asserts the consumer side: with `valkey.existingSecret=acme-valkey`, the rendered Langfuse container references `acme-valkey`, not the chart secret.

### Area

Helm chart and Kubernetes

### Review provenance and autonomy

[...]

Issue #1563, opened 2026-08-15
langfuse.existingSecret leaves the OTel collector authenticating with the published dev key, and the default-credential gate goes green on it


TheConnMan:
## What breaks

Setting `langfuse.existingSecret` redirects the Langfuse project credential for every consumer **except** the OTel collector, which keeps authenticating with the published development key baked into `values.yaml`. Trace export then 401s, silently.

## Root cause

`charts/curie/templates/_helpers.tpl:200-206`:

```
{{- define "curie.otlpAuthHeader" -}}
{{- if .Values.otelCollector.otlpAuthHeader -}}
{{- .Values.otelCollector.otlpAuthHeader -}}
{{- else -}}
{{- printf "Basic %s" (printf "%s:%s" .Values.langfuse.init.projectPublicKey .Values.langfuse.init.projectSecretKey | b64enc) -}}
{{- end -}}
```

A `.Values` composition structurally cannot follow a `secretKeyRef`. The collector's env is hardcoded to `curie.secretName` at `charts/curie/templates/otel-collector.yaml:62-66`, while `langfuse.yaml:56,73,82`, `api.yaml:129`, and `langfuse-model-pricing.yaml:45` all honor the override.

## Reproduction

Default render versus `--set langfuse.existingSecret=my-langfuse`:

```
checksum/config                       identical in both renders
otel collector pod template           byte-identical in both renders
langfuse LANGFUSE_INIT_PROJECT_SECRET_KEY -> secretKeyRef {name: my-langfuse}
collector auth secretKeyRef               -> {name: rel-curie-secrets, key: otlpAuthHeader}
collector auth decoded                    -> pk-lf-curie-dev:sk-lf-curie-dev
```

Only the secret half desyncs; the public key is values-plaintext everywhere.

## Why it is silent, and why the gate misses it

- The collector config sets `telemetry: metrics: level: none` and exposes only `health_check` on 13133, which is what both probes hit. A 401 is a permanent `otlphttp` error: logged and dropped. No crash, no probe failure, no retry escalation.
- No test asserts that a trace actually lands.
- `_helpers.tpl:261` guards the default-credential gate with `if not .Values.langfuse.existingSecret`, so `security.checkDefaultCredentials=true` renders **green** on exactly the path that ships the published dev key. The gate that exists to catch default credentials is disabled by the setting that causes the problem.

## Reachability

[...]

PR #2355, opened 2026-09-04
Honor store existingSecret for the last three curie.langfuse.env refs

### LINKED ISSUE #2327: curie.langfuse.env still hardcodes the chart Secret for POSTGRES_PASSWORD, SALT and ENCRYPTION_KEY, so a BYO-Postgres or BYO-langfuse Secret install gets the chart's values instead
## What happens

#2052 fixed `CLICKHOUSE_PASSWORD` and `REDIS_AUTH` in `curie.langfuse.env` to resolve through `<store>.existingSecret | default (include "curie.secretName" .)`. Three refs in the same block were left on the hardcoded chart Secret.

- **`POSTGRES_PASSWORD` ignores `postgres.existingSecret`**, while `curie.env.postgres` (api + worker) honours it. On a BYO-Postgres install the api and worker authenticate against the real instance; both Langfuse Deployments present the chart-generated `postgresPassword` and fail Prisma auth at boot, crash-looping while every other component is green.
- **`SALT` and `ENCRYPTION_KEY` ignore `langfuse.existingSecret`**, while `NEXTAUTH_SECRET` and both `LANGFUSE_INIT_*` keys in `langfuse.yaml` honour it. `values.yaml:129-133` explicitly instructs the operator that the BYO Secret must carry `langfuseEncryptionKey` and `langfuseSalt` — the template never reads either from there.

`ENCRYPTION_KEY` is the sharpest of the three: it encrypts Langfuse's stored provider keys, so the operator's key is silently unused, and a later regeneration of the chart Secret makes previously written encrypted columns undecryptable.

This is a known gap that has never been filed. `charts/curie/CLAUDE.md:50-58` enumerates exactly these three as *"the remainder, not yet tracked by an issue"* and says *"Until they're fixed the invariant is aspirational for them"*. This issue is that tracking.

## Expected

Every `secretKeyRef` in `curie.langfuse.env` resolves through the same `existingSecret | default <chart secret>` idiom its four siblings in the same block already use.

## Evidence

- `charts/curie/templates/_helpers.tpl:891` — `name: {{ include "curie.secretName" . }}` / `key: postgresPassword`
- `charts/curie/templates/_helpers.tpl:898` — `name: {{ include "curie.secretName" . }}` / `key: langfuseSalt`
- `charts/curie/templates/_helpers.tpl:903` — `name: {{ include "curie.secretName" . }}` / `key: langfuseEncryptionKey`
- Siblings that do it right, in the same helper: `:926` (`clickhouse.existingSecret | default …`), `:937` (`valkey.existingSecret | default …`), `:948` and `:970` (rustfs)
- The split pair: `charts/curie/templates/_helpers.tpl:746` — `name: {{ .Values.postgres.existingSecret | default (include "curie.secretName" .) }}` in `curie.env.postgres`
- The Langfuse-app split: `charts/curie/templates/langfuse.yaml:139`, `:156`, `:165` all use `.Values.langfuse.existingSecret | default …`
- Documentation the template contradicts: `charts/curie/values.yaml:129-133` — "set `existingSecret` to a Secret carrying keys: langfuseEncryptionKey, langfuseSalt, langfuseNextauthSecret, …"
- The standing note: `charts/curie/CLAUDE.md:50-58`
- Rendered proof, `--set postgres.existingSecret=my-pg --set langfuse.existingSecret=my-lf --set clickhouse.existingSecret=my-ch --set valkey.existingSecret=my-vk`:

  ```
  POSTGRES_PASSWORD   -> curie-secrets/postgresPassword        (api/worker -> my-pg/postgresPassword)
  SALT                -> curie-secrets/langfuseSalt
  ENCRYPTION_KEY      -> curie-secrets/langfuseEncryptionKey
  CLICKHOUSE_PASSWORD -> my-ch/clickhousePassword
  REDIS_AUTH          -> my-vk/valkeyPassword
  NEXTAUTH_SECRET     -> my-lf/langfuseNextauthSecret
  ```

- No CI coverage: `charts/curie/ci/valkey-existing-secret-assertions.sh` asserts the consumer side for valkey and `clickhouse-logging-assertions.sh` for clickhouse; there is no postgres sibling, and nothing asserts `langfuse.existingSecret` reaches `SALT` / `ENCRYPTION_KEY`.

## Suggested fix

Change the three `name:` lines to `{{ .Values.postgres.existingSecret | default (include "curie.secretName" .) }}` and `{{ .Values.langfuse.existingSecret | default (include "curie.secretName" .) }}` respectively, then add a postgres sibling to `valkey-existing-secret-assertions.sh` plus a `langfuse.existingSecret` case covering all five Langfuse app keys. Once it is fixed, drop the "Known exceptions" paragraph from `charts/curie/CLAUDE.md:50-58`.

Related: #2052 (fixed the valkey and clickhouse half of this idiom and enumerated these three as the remainder).

TheConnMan:
Closes #2327.

Fix pin: charts/curie/ci/valkey-existing-secret-assertions.sh

`POSTGRES_PASSWORD`, `SALT` and `ENCRYPTION_KEY` in the `curie.langfuse.env` helper resolved to the hardcoded chart Secret while their four siblings in the same block already honoured the store's `existingSecret`. All three now use the same `<store>.existingSecret | default (include "curie.secretName" .)` idiom — `postgres.existingSecret` for the first, `langfuse.existingSecret` for the other two.

### Render proof

Same command both ways: `helm template rel charts/curie --set postgres.existingSecret=byo-postgres --set langfuse.existingSecret=byo-langfuse`.

| env | container | before | after |
|---|---|---|---|
| `POSTGRES_PASSWORD` | langfuse-web, langfuse-worker | `rel-curie-secrets` | `byo-postgres` |
| `SALT` | langfuse-web, langfuse-worker | `rel-curie-secrets` | `byo-langfuse` |
| `ENCRYPTION_KEY` | langfuse-web, langfuse-worker | `rel-curie-secrets` | `byo-langfuse` |
| `POSTGRES_PASSWORD` | api, worker | `byo-postgres` | `byo-postgres` (unchanged) |
| `NEXTAUTH_SECRET`, both `LANGFUSE_INIT_*` | langfuse-web | `byo-langfuse` | `byo-langfuse` (unchanged) |

### Gate

`ci/valkey-existing-secret-assertions.sh` was widened in place rather than forked, because the helm-ci wiring already points at that path and its old header explicitly declared these three keys "deliberately NOT asserted" — text that would otherwise have frozen the bug into the gate. New cases:

- **postgres sibling** — `POSTGRES_PASSWORD` on both Langfuse containers *and* on api/worker in the same render, so a future edit cannot "fix" the split by breaking the app services instead.
- **`langfuse.existingSecret`** — all five Langfuse app keys: `SALT` and `ENCRYPTION_KEY` on both Deployments (shared env helper), `NEXTAUTH_SECRET` and the two `LANGFUSE_INIT_*` keys on web (written inline). Three of the five were already correct and are asserted so a regression on either half fails.

Both carry a negative control and a default-render no-regression pair. **Demonstrated non-vacuous:** with `_helpers.tpl` reverted to `origin/main` the script exits 1 with `12 of 35 python-side assertions failed` (`6a 6b 6e×2 7a 7b 7c 7d 7h×4`); with the fix, 35/35 pass.

The helm-ci step was retitled — its `name:` and comment described only the valkey/#2052 half of a gate that now covers three stores. The `run:` block is byte-identical.

### Upgrade note (please read)

The scope review surfaced a hazard worth flagging: an install **already** on `langfuse.existingSecret` has been salting and encrypting with the *chart* Secret's `langfuseSalt`/`langfuseEncryptionKey`, because those two refs ignored the BYO Secret. `values.yaml` had listed both as required BYO keys all along — the template just never read them. After this upgrade it does. A missing key fails loud (`CreateContainerConfigError`); a *differing* key fails silently and stops decrypting columns already written.

Operators must copy the live values out of the chart-managed `<release>-secrets` Secret into their BYO Secret before upgrading. Documented in `charts/curie/README.md` alongside the existing `otlpAuthHeader` note (#1563). `POSTGRES_PASSWORD` carries no equivalent hazard — Langfuse was crash-looping on that path, so switching it only heals.

### Decisions taken (conservative defaults, per the ambiguity instruction)

- **Widened the existing script instead of adding a new one**, per the issue's own suggested fix. Cost: the filename now under-describes its contents. A rename was skipped deliberately — it would mean editing CI wiring for no behavioural gain. Called out in the file's header.
- **No `values.yaml` change** — it already documented the intended contract; the template is what contradicted it.
- **Kept five separate renders** rather than combining the two new `--set` cases into one. Costs ~160ms against a minutes-scale job and preserves one-variable-per-render attribution on failure.

### Not in this PR

The altitude review flagged that both #2052 and #2327 were fixed the same way — patch the specific lines, enumerate the specific assertions — and neither makes the next instance impossible. The idiom is inlined at ~9 sites in `_helpers.tpl` and the gate hand-enumerates keys, so a tenth ref nobody lists would survive exactly as these three survived #2052. Filed as **#2354** (a named template for the idiom, plus a generic walk-every-`secretKeyRef` assertion). Deliberately out of scope here — it touches every existing site and wants its own render diff.

### Validation

- `helm lint charts/curie` — 0 failed.
- Widened gate — 35/35 pass; 12/35 fail against the unfixed template.
- Neighbouring chart gates green: `render-assertions`, `langfuse-byo`, `direct-passthrough-existing-secret`, `otel-collector-langfuse-secret`, `clickhouse-logging`, `object-store-web-identity`, `langfuse-postgres-readiness`, `connector-secrets`.
- Adversarial code + scope review via agent-router (reviewer: grok, primary excluded). Code review found no correctness defects and independently confirmed no other consumer of either `existingSecret` is newly inconsistent. Scope review found no passengers; its one finding is the upgrade note above, now addressed.

### Done-check

- [x] Delegated executor made every production edit; driver ran only git/validation — build-tier plan-doc and failing-tests-first items N/A (quick tier)
- [x] Adversarial Code Reviewer completed, file:line evidence
- [x] Adversarial Scope Reviewer completed, every hunk mapped to an AC
- [x] Guard demonstration: gate executed against a violating input (12/35 fail), not read
- [x] Sibling-path parity: api/worker postgres half asserted in the same render; residual class filed as #2354
- [x] Consolidation pass ran (4 angles); 2 cleanups applied, validation re-run and guard re-demonstrated after
- [x] Train check: v0.8.6 is a patch milestone, cut from `main`
- [x] `helm lint` clean, full affected gate suite green
- N/A Security reviewer (not flagged), E2E tester, observable UI verification

**Discovery surface:** #2327 was filed from a fresh cluster install. The defect is entirely render-level — which Secret name lands in the manifest — so a `helm template` render is the definitive observation for it, not a proxy for one. No live-cluster waiver needed.

### DIFF
--- a/.github/workflows/helm-ci.yaml
+++ b/.github/workflows/helm-ci.yaml
@@ -152,13 +152,13 @@ jobs:
           set -euo pipefail
           bash charts/curie/ci/clickhouse-logging-assertions.sh
 
-      # Prove Langfuse honors valkey.existingSecret on a BYO valkey install
-      # (#2052). A supported BYO install (valkey.deploy=false + host +
-      # existingSecret) had every component read the operator's Secret except
-      # Langfuse, which stayed pinned to the chart-managed Secret -- so trace
-      # ingestion died alone while the rest of the release stayed healthy,
-      # which is why nothing pointed at the cause.
-      - name: helm render assertions (langfuse valkey existingSecret)
+      # Prove Langfuse honors existingSecret for BYO valkey, postgres, and its
+      # own app keys (#2052, #2327). A supported BYO install has every other
+      # component read the operator's Secret except Langfuse: api and worker
+      # authenticate against the real postgres while both Langfuse Deployments
+      # present the chart-generated password and crash-loop at Prisma auth --
+      # so ingestion dies alone while the rest of the release stays healthy.
+      - name: helm render assertions (langfuse store existingSecret)
         run: |
           set -euo pipefail
           bash charts/curie/ci/valkey-existing-secret-assertions.sh
--- a/charts/curie/CLAUDE.md
+++ b/charts/curie/CLAUDE.md
@@ -59,15 +59,6 @@ component and rail detail in `charts/curie/README.md`.
   BYO `host`/`port`/`auth`/`existingSecret` fields on the same block. A new
   backing store must follow this exact pattern -- do not add a store with a
   different enable/disable shape.
-  Known exceptions: in `curie.langfuse.env` (`_helpers.tpl`), `POSTGRES_PASSWORD`
-  (`postgresPassword`) still ignores `postgres.existingSecret`, and `SALT`
-  (`langfuseSalt`) / `ENCRYPTION_KEY` (`langfuseEncryptionKey`) still ignore
-  `langfuse.existingSecret` -- #2052 fixed the valkey and clickhouse half of
-  this idiom and enumerated these three as the remainder, not yet tracked by
-  an issue. Until they're fixed the invariant is aspirational for them: an
-  operator on BYO postgres or a BYO `langfuse.existingSecret` gets the
-  chart-generated value instead of theirs. `ENCRYPTION_KEY` is the sharpest --
-  a mismatched key means previously written encrypted columns stop decrypting.
 - **Values keys are camelCase, not hyphenated.** Go templates cannot
   dot-index a hyphenated key. Keep this consistent across any new values
   additions.
--- a/charts/curie/README.md
+++ b/charts/curie/README.md
@@ -547,6 +547,18 @@ upgrading. Without it the collector pod fails to start with
 `CreateContainerConfigError`, which is the deliberate replacement for a silent
 401 on every trace export.
 
+Upgrade note for `langfuse.existingSecret` (#2327): the chart now actually reads
+`langfuseSalt` and `langfuseEncryptionKey` from that Secret. `values.yaml`
+already listed both as required keys, but the template ignored them -- an
+install already on this path was salting and encrypting with the chart-managed
+Secret's values instead, because those two keys ignored `existingSecret` while
+`NEXTAUTH_SECRET` and the `LANGFUSE_INIT_*` keys already honored it. Both
+Deployments now read the BYO Secret for real: a missing key fails loud with
+`CreateContainerConfigError`, a differing one fails silently and stops
+decrypting columns already written. Before upgrading, copy the live values out
+of the chart-managed `<release>-secrets` Secret into your BYO Secret, unless no
+encrypted data is worth preserving.
+
 Caveat: generation relies on Helm `lookup`, which is empty under client-side
 rendering. Driving this chart via `helm template | kubectl apply` or ArgoCD's
 client-side Helm (no live API lookup) regenerates these values on every sync and
--- a/charts/curie/ci/valkey-existing-secret-assertions.sh
+++ b/charts/curie/ci/valkey-existing-secret-assertions.sh
@@ -1,14 +1,24 @@
 #!/usr/bin/env bash
 #
-# Render-assertion test for `valkey.existingSecret` reaching Langfuse (#2052).
+# Render-assertion test for `<store>.existingSecret` reaching every Langfuse
+# consumer (#2052 valkey, #2327 postgres + langfuse).
+#
+# The filename is historical: this started as the valkey/REDIS_AUTH gate and
+# was widened in place rather than forked, because every case here is the same
+# split -- one consumer group honouring `<store>.existingSecret` while
+# `curie.langfuse.env` hardcoded the chart Secret -- and the helm-ci wiring
+# already points at this path.
 #
 # The chart's stated invariant (charts/curie/CLAUDE.md, "Every backing store
 # follows the same toggle + BYO idiom") is that flipping `<store>.deploy` to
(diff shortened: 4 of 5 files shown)


TheConnMan:
Note on the red `cargo audit` check: it is **not** from this PR. `Dependency Audit` is already failing on `main` (two runs today), this branch touches no `.rs`, `Cargo.toml` or `Cargo.lock` file, and the finding is RUSTSEC-2023-0071 (Marvin attack) in the transitive `rsa` crate, which reports `Solution: No fixed upgrade is available!`.

Nothing in this diff can move it, so it is left alone rather than papered over with an ignore entry here — that call belongs in its own change.

Issue #2370, opened 2026-09-04
Finish collapsing the duplicated egress ipBlock rule body and rename the misnamed shared entry validator


TheConnMan:
## What happens

The block that emits one egress rule per operator-declared CIDR entry — `- to:` / `ipBlock:` / `cidr:` / the explicit `except:` versus the `curie.metadataExcept` auto-carve-out / `{{- with .ports }}` — is written out verbatim in several places across the chart.

#2317 collapsed the five copies inside `charts/curie/templates/security-networkpolicy.yaml` behind a single `curie.egress.ipBlockRules` named template. Two copies remain in `charts/curie/templates/mail-adapter.yaml` (the adapter's provider-egress and API-egress rules), which that PR deliberately left alone to keep a patch-release security fix narrow.

This matters because the auto-carve-out is a security invariant, not formatting: *an entry with no explicit `except` must have the cloud metadata address carved back out of it*. NetworkPolicy allows are additive, so a broad CIDR that skips the carve-out silently re-permits `169.254.169.254` on a render that looks completely healthy. Every additional copy is another place a future fix to that invariant has to land, and a missed copy fails silently and in the dangerous direction.

## Expected

The two `mail-adapter.yaml` rule bodies use the same `curie.egress.ipBlockRules` helper, leaving exactly one implementation of the carve-out in the chart.

Two related cleanups belong with it:

- **`curie.objectStore.egressEntry` is now misnamed and misplaced.** It validates four unrelated keys (`rustfs.egress`, `rustfs.stsEgress`, `otelCollector.egress`, `api.egress`) and lives inside `security-networkpolicy.yaml` rather than `_helpers.tpl`. The `objectStore` name will misdirect the next author into thinking it is store-specific. Rename to something peer-neutral and move it beside the other shared defines. #2317 corrected only its doc comment, deliberately leaving the rename out of a patch-release fix. Note that `charts/curie/ci/object-store-byo-egress-assertions.sh` greps some of its failure messages, so the rename must keep the message text and its `%s` key argument intact.
- **`mail-adapter.yaml` carries its own independent copy of the entry validator.** Once the shared one is renamed and moved, the adapter should use it rather than a parallel implementation.

## Evidence

- `charts/curie/templates/mail-adapter.yaml` — the two remaining ipBlock rule bodies and the independent entry validator
[...]

Issue #2369, opened 2026-09-04
BYO runner egress is pinned only at render tier; the security probe should prove the sandbox actually reaches each declared peer


TheConnMan:
## What happens

#2213 and #2317 both closed BYO runner-egress gaps that were **found on a real cluster install** and are **pinned only at render tier**. The assertions in `charts/curie/ci/object-store-byo-egress-assertions.sh` and `charts/curie/ci/byo-endpoint-egress-assertions.sh` prove the chart renders a NetworkPolicy naming the declared CIDRs, and that it refuses the render when they are missing. Neither proves the runner can actually reach the endpoint.

That leaves the fix inert against the most likely real-world failure: the declared CIDR does not match what DNS returns at runtime. A SaaS OTLP collector or an external API behind rotating load-balancer addresses renders green, installs green, and still drops every sandbox span or boots every agent with no memory — the exact symptom #2317 describes, now with a NetworkPolicy that appears to permit it.

A CNI that does not enforce NetworkPolicy at all turns the whole rail into a silent false pass, which the chart already acknowledges in `security-networkpolicy.yaml`'s header comment.

## Expected

The security probe already builds the right primitive: a pod carrying the runner-sandbox labels running a before/after egress check (`charts/curie/templates/security-probe.yaml`). Extend it so that, when `rustfs.egress`, `rustfs.stsEgress`, `otelCollector.egress` or `api.egress` is declared, the probe issues a request to each declared endpoint from inside that pod and asserts it succeeds — and that an undeclared control address still fails.

That single observation converts all three fixes in this class from "the template renders the right YAML" to "the runner can reach the peer and nothing else", and it is the same check that catches a non-enforcing CNI.

## Evidence

- `charts/curie/templates/security-probe.yaml` — the existing runner-labelled before/after egress check
- `charts/curie/templates/security-networkpolicy.yaml` — the header comment naming the non-enforcing-CNI false-pass this probe exists to catch
- `charts/curie/ci/object-store-byo-egress-assertions.sh`, `charts/curie/ci/byo-endpoint-egress-assertions.sh` — the render-tier pins as they stand

Related: #2213, #2317, #2367, #2368.

## Bonus Drain queue alignment (2026-09-09)

Task: `curie-v088-2369`. Group: **Curie v0.8.8**. Size: **large**. Priority: **2**.

Prerequisites: `curie-v088-2367`, `curie-v088-2361`.

[...]

Issue #2368, opened 2026-09-04
The shared BYO egress entry validator claims a narrowness it does not enforce (/2 prefixes, IPv6 ULA, port-less ports)


TheConnMan:
## What happens

`curie.objectStore.egressEntry` in `charts/curie/templates/security-networkpolicy.yaml` validates every operator-declared BYO egress entry. It was written for #2213's two object-store keys; #2317 extended it to `otelCollector.egress` and `api.egress`, so it now guards **four** lists feeding additive NetworkPolicy allows on a prompt-injectable runner sandbox.

Its stated contract is that each list names ONE external peer, not a second model allowlist. Three gaps let an entry be far broader than that:

**1. Only prefix 0 and 1 are refused.** `0.0.0.0/2`, `64.0.0.0/2` and `192.0.0.0/2` all pass and together cover three quarters of IPv4. The refusal message says "use a VPC interface endpoint /32 or the narrowest prefix that reaches this one endpoint", but nothing enforces anything close to that.

**2. IPv6 containment is not checked.** The validator's own comment claims it refuses "any CIDR that contains `169.254.169.254` or `fd00:ec2::254`". That is exact and sound for IPv4, but `fd00::/8` and `fc00::/7` both contain `fd00:ec2::254` and both pass. `curie.metadataExcept` documents its IPv6 range math as partial (`::/0` and `fd00:ec2::/nn` only), and the validator inherits that limit while asserting more than it delivers.

**3. A ports entry with only `protocol`.** `ports: [{protocol: TCP}]` satisfies the "must set ports" check and renders a NetworkPolicy port rule with no `port`, which matches every TCP port on the allowed CIDR.

Separately, the auto-`except` branch inside the rendered BYO policies is unreachable: the validator already refuses any CIDR for which `curie.metadataExcept` returns non-empty, so the `{{- if $auto }}` inside the policy body can never fire. It is harmless and symmetric with the sibling branches, but it reads as live defence-in-depth that is not doing anything.

## Expected

Either the validator enforces the narrowness its message and comments claim — a maximum prefix breadth, IPv6 ULA containment, and a required `port` per entry — or the comments are corrected to state what is actually checked. The current state is worse than either, because a reviewer reading the comment believes a bound exists.

## Evidence

[...]

Issue #2367, opened 2026-09-04
The runner egress guards key on <component>.deploy, so an external endpoint reached with the in-chart pod deployed is still unpermitted


TheConnMan:
## What happens

#2317 closed the `deploy: false` half of the runner egress class: with `otelCollector.deploy: false` or `api.deploy: false`, the chart now requires an operator-declared CIDR list and renders a runner NetworkPolicy from it, or refuses the render.

An adversarial review of that fix found the **inverse shape is still open**, and it produces the same silent failure #2317 describes.

**1. `api.deploy: true` with an external `dispatcher.apiBaseUrl`.** The in-chart pod-selector policy (`security-networkpolicy.yaml`, the `api.deploy` branch) renders and looks correct, while `curie.env.apiUrl` hands the worker the external base URL and the worker mints the runner's memory/history/state refs off it (`apps/worker/src/curie_worker/binding.py:686`). The runner then dials a host no rendered policy permits. `api.egress` is not read on this path — after #2317 it fails the render if set there, which surfaces the misconfiguration but does not solve it. `security.networkPolicy.allowedEgress` is the only mechanism, and nothing tells the operator that.

**2. `worker.extraEnv` overriding `CURIE_API_URL` / `CURIE_RUNNER_API_URL`.** `dispatcher.apiBaseUrl` stays empty, so neither the #2317 guard nor anything else fires, and the runner is again pointed at an unpermitted host.

**3. `agentSandbox.runner.extraEnv` supplying `OTEL_EXPORTER_OTLP_ENDPOINT`.** Same shape on the collector side: the resolved `curie.otel.endpoint` does not reflect the override, so no guard fires.

The symptom in every case is the one #2317 documents: agents booting with no memory and no history, or sandbox spans silently dropped, with only a warning line inside the sandbox.

## Expected

The requirement should key on the **effective runner-facing endpoint**, not on `<component>.deploy`. Whenever the address the runner will actually dial is external, the chart requires a declared peer or refuses, regardless of which toggle produced it.

This was deliberately left out of #2317: `dispatcher.apiBaseUrl` with `api.deploy: true` is a supported, tested override (`charts/curie/ci/dispatcher-api-wiring-assertions.sh:261`), so making it fail closed is a behavioural change beyond that ticket's scope and needs its own decision about the upgrade path for installs relying on `allowedEgress` today.

[...]

PR #2372, opened 2026-09-04
Render runner egress for the BYO collector and API endpoints

### LINKED ISSUE #2317: The runner's collector and API egress carve-outs are gated on the in-chart pod, so both BYO paths silently lose function (the other two thirds of #2213)
## What happens

#2213 found that the runner sandbox's object-store egress carve-out is gated on the in-chart RustFS pod, so a BYO object store gets no replacement rule and every sandbox hangs. That is one of **three** carve-outs in `security-networkpolicy.yaml` built the same way. The other two have the same defect and are still open.

Rail 1 is fail-closed, `security.networkPolicy.allowedEgress` defaults to `[]` (`values.yaml:1900`), and each carve-out is gated on `<component>.deploy` with a `to:` that is a podSelector on this release's own pod. Set the component to `deploy: false`, point it at an external endpoint, and the rule disappears with nothing rendered in its place.

**1. BYO OTel collector.** With `otelCollector.deploy: false` + `otelCollector.endpoint`, the chart wires the external endpoint into every instrumented workload including the runner sandbox. The runner is the only one of those under an egress rail, and its collector allow is gated on the in-chart pod.

Rendered with `--set otelCollector.deploy=false --set otelCollector.endpoint=https://otlp.example.net:4318`:

```
sandbox env:  OTEL_EXPORTER_OTLP_ENDPOINT: "https://otlp.example.net:4318"
runner netpols: default-deny-egress, allow-dns, allow-api, allow-rustfs, ingress
                (no collector allow, and no runner-allow-egress at all)
```

The runner's exporter is bounded (`BatchSpanProcessor` with a 5 s export timeout), so nothing crashes and no turn fails. The symptom is a silent asymmetry: **API, worker and dispatcher traces reach the external collector normally while every runner / `agent.run` / `llm.generation` / `execute_tool` span is dropped**, plus up to 5 s of flush stall per turn. Nothing in the logs points at egress. The template's own comment for the in-chart case names this exact symptom — *"Without this, the default-deny blocks span export and sandbox traces vanish"* — it is simply never made true for the BYO case.

**2. BYO platform API.** `api.deploy: false` is a supported path (`NOTES.txt:51-61`, and the mail adapter refuses to render without it configured). On that path the worker's `CURIE_API_URL` derives from `dispatcher.apiBaseUrl`, and the worker mints the runner's memory/history/state refs off that same base — so the runner is handed refs on the operator's external API host. The API allow is gated on the in-chart pod.

Rendered with `--set api.deploy=false --set dispatcher.apiBaseUrl=https://api.example.net`:

```
worker env:  CURIE_API_URL: "https://api.example.net"
runner netpols: default-deny-egress, allow-dns, allow-collector, allow-rustfs, ingress
                (no API allow, and no runner-allow-egress at all)
```

The runner degrades rather than fails — it logs "booting without memory" / "booting without history" and continues — so the install looks healthy. Every turn boots with **no prior memory and no thread transcript**, `remember` writes never persist, and the auto-mounted `curie-state` MCP server cannot reach its store. The only trace is a `WARNING memory load failed` line inside the sandbox, plus roughly 30 s of added boot latency (two 15 s aiohttp total timeouts, since a NetworkPolicy drop is a timeout rather than a refusal) eating into the 90 s claim-bind window. This is the failure #1530 fixed for the in-chart case; the fix that landed covers only `api.deploy: true`.

## Expected

Each carve-out either renders an egress allow derived from the configured external endpoint on the `deploy: false` path, or refuses the render until the operator declares one — the way `mail-adapter.yaml:82-88` already refuses when `mailAdapter.apiEgress.httpsCidrs` is unset on the same BYO API path. Failing that, the README states in both the external-collector and BYO-API sections that `security.networkPolicy.allowedEgress` must carry the endpoint.

## Evidence

Collector:

- `charts/curie/templates/security-networkpolicy.yaml:107` — `{{- if .Values.otelCollector.deploy }}`
- `charts/curie/templates/security-networkpolicy.yaml:109-111` — the comment naming the symptom for the in-chart case
- `charts/curie/templates/security-networkpolicy.yaml:129-131` — the `to:` podSelector on this release's `otel-collector` component
- `charts/curie/templates/_helpers.tpl:494-500` — `curie.otel.endpoint` returns `.Values.otelCollector.endpoint` when `deploy` is false
- `charts/curie/templates/agent-sandbox.yaml:604` — the runner container includes `curie.env.otel`
- `charts/curie/README.md:920` — "Traces flow to … `otelCollector.endpoint` when that BYO field is set", with no egress caveat

API:

- `charts/curie/templates/security-networkpolicy.yaml:136` — `{{- if .Values.api.deploy }}`
- `charts/curie/templates/security-networkpolicy.yaml:138-141` — the comment states the intended contract (*"An external or overridden API destination remains governed by the operator's allowedEgress list"*), which appears nowhere in the README or values file
- `charts/curie/templates/security-networkpolicy.yaml:159-161` — the `to:` podSelector on this release's `api` component
- `charts/curie/templates/_helpers.tpl:785-796` — `curie.env.apiUrl`, `CURIE_API_URL` from `dispatcher.apiBaseUrl`
- `charts/curie/templates/worker.yaml:186` — the worker includes that helper
- `apps/worker/src/curie_worker/binding.py:686-695` — memory/history refs minted off the runner-facing API base
- `runner/src/curie_runner/__main__.py:427-434`, `:469-475` — load failure degrades to no memory / no history, never fails boot
- `runner/src/curie_runner/memory.py:192` — `aiohttp.ClientTimeout(total=15)` per load

Both:

- `charts/curie/values.yaml:1900` — `allowedEgress: []`, with the comment *"Deliberately fail-closed -- an unset allowlist must never mean 'allow everything'."*
- `charts/curie/README.md:715-718` — the fail-closed egress note frames the allowlist as "the model API and MCP endpoints" only
- `charts/curie/templates/mail-adapter.yaml:82-88` and `:322-330` — the correct BYO-egress pattern, already in this chart: a render that fails closed plus an `{{- else }}` ipBlock branch

## Suggested fix

Mirror the mail adapter for both. Add an operator-declared CIDR block per external peer (or one shared `security.networkPolicy` entry per BYO endpoint), render an `{{- else }}` ipBlock branch at `security-networkpolicy.yaml:107` and `:136`, and fail the render when the component is `deploy: false` with no declared peer. A BYO-API install with silently amnesiac agents, and a BYO-collector install with silently missing sandbox traces, are both worse outcomes than a refused render.

Then close the class: a render assertion that every `<component>.deploy: false` path either renders a replacement egress rule or fails, so the next BYO toggle added to the chart cannot re-open it. #2213, and these two, are the same defect found three times.

Related: #2213 (the object-store instance, fixed), #1530 (the in-chart half of the API case), #1529 (the paired worker `CURIE_API_URL` gap).

TheConnMan:
Rail 1 is fail-closed, and the runner's collector and API egress carve-outs each selected this release's own pod, so `otelCollector.deploy=false` and `api.deploy=false` left the sandbox with no replacement rule. #2213 fixed the object-store instance; these are the other two thirds.

Both failures were silent. With an external collector, api/worker/dispatcher traces kept flowing while every runner / `agent.run` / `llm.generation` / `execute_tool` span was dropped and each turn paid the exporter's flush stall. With an external API, every turn booted with no prior memory and no thread transcript, `remember` writes never persisted, and the auto-mounted `curie-state` MCP server could not reach its store — with only a warning line inside the sandbox to show for it.

## What changed

Both gates take an `{{- else }}` branch mirroring the shipped object-store fix: an operator-declared CIDR list (`otelCollector.egress`, `api.egress`) validated by the shared entry validator, a render that fails closed naming the exact key, and a runner NetworkPolicy built from those CIDRs.

The class guard is the point rather than a third hand-written instance. It discovers every template that renders a NetworkPolicy and requires each `<component>.deploy` carve-out to either refuse the render or render a replacement peer, or to carry a written exemption keyed by file and component. An empty `{{- else }}` does not satisfy it. The mail adapter's collector gate takes an exemption for its already-documented asymmetry.

## Conservative defaults, called out

- **The collector requirement is scoped to a non-empty resolved OTLP endpoint, and the API requirement to a non-empty `dispatcher.apiBaseUrl`.** An install with telemetry off, or with no external API named, has nothing for the runner to reach and is unaffected. The issue text implied the API case should be unconditional on `api.deploy=false`; unconditional broke `values-e2e-harness.yaml` and several sibling fixtures, which is the signal that shape names no peer.
- **Setting either key while its component is still in-chart is refused, not silently ignored.** These keys are new, so nobody can have them set yet.
- **Both `deploy: false` refusals name the migration path.** The old template comment told operators an external API destination was governed by `allowedEgress`, so some put the CIDR there; those installs now fail on upgrade, and the message says to move it.
- **Consolidation.** The rule body carrying the metadata carve-out invariant was about to be written for the fifth time in this template, so all five sites now share one helper. Rendered output is unchanged byte for byte, verified against a values set exercising all five branches at once. The two further copies in `mail-adapter.yaml` were deliberately left alone (#2370).

## Verification

`helm lint` clean. All 44 `charts/curie/ci/*-assertions.sh` pass. The two renders from the issue now emit `runner-allow-collector-endpoint` and `runner-allow-api-endpoint` respectively, and each fails the render naming its key when the list is omitted; both new in-chart guards were demonstrated rejecting live.

This is pinned at render tier only. The defect class was found on a real cluster install, and a NetworkPolicy naming the right CIDR still drops traffic if the declared CIDR is not what DNS returns at runtime — #2369 tracks extending the security probe so the sandbox proves it actually reaches each declared peer.

Adversarial review: the routed cross-engine review failed on infrastructure (the Grok reviewer's storage requires `openat2`, which this session's seccomp filter blocks, and no systemd user bus was available to work around it). The review that did run was Claude Fable — a different model family but the same provider as the executing engine, so it does not satisfy the cross-engine contract. Its findings were triaged and acted on: the silent no-op and the false comment are fixed here, and #2367 / #2368 were filed for the rest.

Follow-ups filed: #2367 (guards key on `<component>.deploy` rather than the effective endpoint), #2368 (the shared validator claims a narrowness it does not enforce), #2369 (render-tier-only pin), #2370 (finish the consolidation, rename the misnamed validator).

Fix pin: charts/curie/ci/byo-endpoint-egress-assertions.sh

Closes #2317

### DIFF
--- a/.github/workflows/helm-ci.yaml
+++ b/.github/workflows/helm-ci.yaml
@@ -499,6 +499,16 @@ jobs:
           set -euo pipefail
           bash charts/curie/ci/object-store-byo-egress-assertions.sh
 
+      # Prove the same for the other two in-chart runner carve-outs: a BYO
+      # collector or API endpoint renders its own ipBlock allow and is refused
+      # at render without one. Assertion 10 is the class guard -- a future
+      # `.deploy` carve-out added without a BYO else branch fails here rather
+      # than in a customer's silently trace-less or amnesiac install (#2317).
+      - name: helm render assertions (BYO collector/API runner egress)
+        run: |
+          set -euo pipefail
+          bash charts/curie/ci/byo-endpoint-egress-assertions.sh
+
       # Prove a nil worker.publication.githubHttpsCidrs cannot collapse the
       # tokenless publication job's GitHub egress rule to an empty to: (allow-all
       # on 443). Also rejects that empty-to shape on every rendered
--- a/charts/curie/README.md
+++ b/charts/curie/README.md
@@ -154,6 +154,19 @@ because the API's `/health` is unauthenticated, that first preflight proves only
 reachability. The following authenticated `/agents` discovery refuses startup
 when a BYO API expects a different key. Match `apiKey` to the API you point at.
 
+`api.deploy: false` also moves the runner sandbox's memory, history and state
+calls onto that external host, and the sandbox is under Rail 1's default-deny
+egress. The in-chart `runner-allow-api` policy selects this release's API pods,
+so it does not render, and NetworkPolicy has no hostname peer to derive from
+`dispatcher.apiBaseUrl`. Set `api.egress` to the endpoint's CIDRs (`{cidr,
+ports}` entries, same shape as the allowlist); the chart requires it at render
+whenever `dispatcher.apiBaseUrl` names an external API, because the failure it
+prevents is silent — agents boot with no
+prior memory and no thread transcript, `remember` writes never persist, and the
+only symptom is a warning line inside the sandbox.
+`mailAdapter.apiEgress.httpsCidrs` above is the same idea for the mail adapter
+pod: one BYO peer, declared explicitly, per pod that has an egress policy.
+
 **Cluster variants:**
 
 - **Cluster already runs the agent-sandbox controller** (cluster-scoped, one per
@@ -889,10 +902,15 @@ sandbox and renders whenever an in-chart store is deployed.
 **Fail-closed egress.** `security.networkPolicy.allowedEgress` is EMPTY by
 default: a fresh install denies all egress except DNS until the operator declares
 where the model API and MCP endpoints live (`{cidr, ports}` entries). An unset
-allowlist never means allow-all. A BYO object store is not on that list: set
+allowlist never means allow-all. The BYO in-chart peers are not on that list
+either, because each names one endpoint rather than a class of destinations: set
 `rustfs.egress` (and `rustfs.stsEgress` on the key-free path) so the sandbox
-bundle-fetch can reach S3 and STS without opening the model allowlist. See
-**Key-free object store auth** above.
+bundle-fetch can reach S3 and STS, `otelCollector.egress` when
+`otelCollector.deploy: false` points the runner at an external collector, and
+`api.egress` when `api.deploy: false` points it at an external API. Each is
+required at render on its BYO path, so the install fails loudly instead of
+shipping a sandbox holding an address it can never reach. See **Key-free object
+store auth** above.
 
 **The controller does not get a second vote on egress (#765, ADR-0067).**
 NetworkPolicy allows are additive across objects selecting the same pods -- Rail
@@ -1100,7 +1118,11 @@ to install only the control plane + backing stores without the runner substrate.
 - Traces flow to `<release>-otel-collector:4318` (HTTP) when the chart collector
   is deployed, or to `otelCollector.endpoint` when that BYO field is set. The
   env block is omitted only for explicit `telemetryDisabled` or local/offline
-  no-endpoint mode.
+  no-endpoint mode. The runner is under Rail 1, so on the BYO path the endpoint
+  alone gets it nowhere: `otelCollector.egress` must name the collector's CIDRs
+  or the default-deny drops every sandbox span while api, dispatcher and worker
+  keep exporting normally. The chart requires it at render rather than letting
+  that asymmetry ship silently.
 
 ## Deploying without inbound access
 
--- a/charts/curie/ci/byo-endpoint-egress-assertions.sh
+++ b/charts/curie/ci/byo-endpoint-egress-assertions.sh
@@ -1,0 +1,841 @@
+#!/usr/bin/env bash
+#
+# Render-assertion test for issue #2317. Rail 1 (the runner-sandbox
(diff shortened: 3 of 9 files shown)


TheConnMan:
`cargo audit` is red here for a reason unrelated to this PR: it fails the same way on `main` (RUSTSEC-2023-0071 in the transitive `rsa` crate, plus three allowed warnings), and this branch touches no Rust or Cargo files. Every other check is green, including Helm CI, both E2E parity ladders, the released-chart upgrade, and Python.

Issue #71, opened 2026-07-07
K8s: add liveness probes (worker and dispatcher have none at all)


TheConnMan:
**Severity: Medium.**

**Evidence:** grep-confirmed zero probes on worker, dispatcher, otel-collector; readiness-only (no liveness) on clickhouse, valkey, minio, ui. `livenessProbe` exists only on api, langfuse-web, postgres. The worker is the concurrency kernel driving every sandbox claim — if its event loop wedges, nothing restarts it and no readiness gate removes it; runs hang until the 90s claim timeout. The dispatcher's reconnect supervisor can wedge with the same invisibility.

**Fix:** add liveness to every long-running workload. Worker/dispatcher have no HTTP port — use an `exec` liveness (heartbeat file touched by the loop, or a process-health check) or expose `/healthz`. Add TCP/exec liveness to clickhouse (`clickhouse-client --query 'SELECT 1'`), valkey (`valkey-cli ping`), minio (`/minio/health/live`), and ui.

**Success criterion:** every Deployment/StatefulSet/DaemonSet with a persistent process has both a readiness and a liveness probe.

PR #2416, opened 2026-09-05
Source probe cadences from values so liveness outlives readiness

TheConnMan:
Four rendered containers had a liveness probe whose earliest failure cutoff came before their own readiness tolerance, so the kubelet could restart a container that was still inside the boot window readiness was sized for (postgres WAL replay, the Langfuse Prisma/ClickHouse boot migrations on a replicas-1 Recreate Deployment, api warm-up). This moves those probe cadences into values.yaml per the chart rule, raises the liveness defaults so the ordering holds, and adds a render assertion so the class cannot recur.

Cutoff formula (matches `templates/dispatcher.yaml`): `initialDelaySeconds + (failureThreshold - 1) * periodSeconds`, kubelet defaults applied to omitted keys.

| container | readiness cutoff | liveness before | liveness after | liveness timeout before -> after |
| postgres | 60 s | 40 s | 310 s | 1 s (kubelet default) -> 5 s |
| langfuse-web | 310 s | 190 s | 370 s | 1 s (kubelet default) -> 5 s |
| langfuse-worker | 310 s | 190 s | 370 s | 5 s -> 5 s |
| api | 120 s | 105 s | 195 s | 1 s (kubelet default) -> 5 s |

Readiness cadences, probe handlers, endpoints, ports, replica counts and strategies are byte-identical to before. A render diff against origin/main (default and values-dev) shows only the probe cadence lines and two comments changing. langfuse-worker is included beyond the three containers the review named because `charts/curie/CLAUDE.md` required web and worker to be lifted onto values together, never piecemeal, and the worker's own liveness (190 s) was also below its readiness (310 s).

New values keys: `postgres.readinessProbe`, `postgres.livenessProbe`, `langfuse.web.readinessProbe`, `langfuse.web.livenessProbe`, `langfuse.worker.readinessProbe`, `langfuse.worker.livenessProbe`, `api.readinessProbe`, `api.livenessProbe` (each `initialDelaySeconds` / `periodSeconds` / `timeoutSeconds` / `failureThreshold`). `values.schema.json` is untouched.

New gate: `charts/curie/ci/probe-window-assertions.sh`, wired into `helm-ci.yaml`. It renders the chart with default and values-dev values, walks every object carrying a pod spec (containers and initContainers), and fails when any container's liveness cutoff is not strictly greater than its readiness cutoff, when a livenessProbe omits `timeoutSeconds`, or when postgres liveness `timeoutSeconds` is below 5. Five red controls mutate a real render and needle-match the rejection; one override control sets all four `livenessProbe` values keys at once and requires each container to go red by name, proving every template reads its key. On the unmodified templates it reports seven problems on exactly these four containers; on this branch it passes.

Decisions taken as the conservative default: raised `failureThreshold` with an explicit `timeoutSeconds` rather than a startupProbe (one probe shape, matches every other container in the chart); the assertion covers the default and values-dev renders only, so `dispatcher`, `mail-adapter` and `inference` (deploy off in both) are outside its reach although correctly ordered today; the remaining correctly ordered hardcoded probe blocks (ui, clickhouse, valkey, rustfs, otel-collector, mail-adapter, inference, `curie.heartbeatProbes`) are left for a follow-up. Consolidation review noted two deeper changes deliberately left for a follow-up: a shared `_helpers.tpl` cadence helper that also `fail()`s at render time when liveness does not exceed readiness (the way `dispatcher.yaml` guards its startupProbe), which would extend the guard to operator overrides, and a shared pod-spec walker between this script and `render-assertions.sh`, whose kind lists have already drifted. Advisory reviewer notes not acted on here: a `livenessProbe: null` or single-key partial override in an overlay can re-invert a container at render time (the assertion catches it in CI, the templates do not `fail()`), and the assertion skips containers carrying only one probe, so the both-probes rule from #71 stays pinned only for langfuse-worker.

## Cluster proof (k8scratch, disposable namespaces, torn down)

Baseline = the origin/main chart, after = this branch, both installed with `values-dev.yaml` + `values-e2e-nogvisor.yaml`, sandbox/worker/dispatcher/ui/otel off, `--no-hooks`. Same script for both.

```
helm install <rel> charts/curie -n <ns> --create-namespace -f charts/curie/values-dev.yaml -f charts/curie/values-e2e-nogvisor.yaml \
  --set worker.deploy=false --set dispatcher.deploy=false --set ui.deploy=false --set otelCollector.deploy=false \
  --set priorityClasses.platform.create=false --set priorityClasses.sandbox.create=false --no-hooks --wait
kubectl -n <ns> exec <rel>-postgres-0 -- pgbench -i -q -s 30 -U postgres postgres
kubectl -n <ns> exec <rel>-postgres-0 -- pgbench -c 8 -j 2 -T 90 -U postgres postgres &   # write load
sleep 60; kubectl -n <ns> exec <rel>-postgres-0 -- kill -QUIT 1                            # immediate shutdown, no checkpoint
sleep 40; kubectl -n <ns> delete pod <rel>-postgres-0 --force --grace-period=0              # second variant, mid-write
helm upgrade <rel> charts/curie -n <ns> --reuse-values --set langfuse.web.nodeOptions=... --no-hooks --wait   # cold Recreate
kubectl -n <ns> get pod <pod> -o jsonpath='{.status.containerStatuses[0].restartCount}'
kubectl -n <ns> get events --field-selector reason=Killing
```

Postgres, SIGQUIT mid-write (WAL replay on restart, "database system was not properly shut down; automatic recovery in progress"):

| | redo elapsed | Ready after | restartCount | liveness kills |
| baseline | 7.49 s | 13 s | 1 | 0 |
| after | 6.53 s | 14 s | 1 | 0 |

Postgres, `delete pod --force --grace-period=0` mid-write: new pod Ready in 8 s (baseline) and 9 s (after), restartCount 0, no liveness kill either way. On this node recovery is far shorter than the old 40 s cutoff, so the baseline could not be made to crash-loop on recovery alone; the 1 s exec timeout did fire on the baseline under a 50m CPU cap (`Liveness probe failed: command timed out: "sh -c pg_isready -U postgres" timed out after 1s`), and did not on the branch (timeout 5 s).

langfuse-web cold `helm upgrade` (Recreate, boot migrations), node-default CPU: Ready in 59 s both before and after, restartCount 0, zero `Killing ... failed liveness probe` events for langfuse-web in either namespace.

langfuse-web cold boot with CPU request=limit=50m (`--set langfuse.web.resources.limits.cpu=50m --set langfuse.web.resources.requests.cpu=50m`), the case the old cutoff actually reaches:

| | outcome |
| --- | --- |
| baseline (liveness 90/20/1/6) | `Killing x3: Container langfuse-web failed liveness probe, will be restarted`; restartCount 3 at 636 s; never Ready in 772 s |
| after (liveness 90/20/5/15) | 6 slow liveness probes absorbed, Ready at 304 s, restartCount 0, no Killing |

Static checks on this branch: `helm lint charts/curie` and `helm lint charts/curie -f charts/curie/values-dev.yaml` clean; all 48 `charts/curie/ci/*.sh` scripts pass; `charts/curie/ci/probe-window-assertions.sh` red on the origin/main templates (7 problems) and green here.

## Tier classification

| tier | decision | reason |
| --- | --- | --- |
| skill | n/a | no plugin, skill, runner turn loop, ACI or eval surface is touched |
| local | n/a | compose services, dispatcher, worker, API wiring and the CLI are untouched; the change is chart-only |
| local-release | n/a | no binary, image identity, install path, version pin or release compose change |
| cluster | required | chart templates changed; proved above on k8scratch with the origin/main chart as baseline |
| live provider | n/a | no model routing, credential resolution or provider auth surface |
| external integration | n/a | no Slack, webhook, connector or third-party API shape |

## Done-check

- [x] Failing tests committed before implementation: `probe-window-assertions.sh` landed first, red on the templates (7 problems), then the template/values commit turned it green.
- [x] All production edits by delegated native executors (Test Writer, Implementer, two Fixers); the driver ran only validation, git and bookkeeping.
- [x] Broadened return types: N/A, no code returns changed (chart templates, values, bash/python CI script).
- [x] Agent-router Code Reviewer: completed (grok), verdict ACCEPT, five advisory findings; three fixed (four-container override control, walker covers Pod/SandboxTemplate, stale error string), two recorded above as follow-ups.
- [x] Agent-router Scope Reviewer: completed (grok), verdict ACCEPT; every hunk mapped to an AC or a mechanical consequence; the langfuse-worker lift ruled justified by the chart's own documented rule.
- [x] Sibling-path parity: the assertion walks every pod spec in both renders rather than a named list; the three deploy-off workloads are named as the stated gap.
- [x] Guard demonstration: the assertion was executed red on the base templates and its five negative controls plus the override control each rejected a violating render (output in the run transcript).
- [x] Consolidation pass ran (`/simplify`) and applied edits (compacted README rows and values comments, shared pod walker in the assertion); revalidated green with a byte-identical values-dev render, and the agent-router re-review of that diff (codex) returned ACCEPT.
- [x] Security Reviewer: N/A, no security scale-up flagged.
- [x] Observable verification / E2E: cluster tier proved above.
- [x] Train check: shared-line chart fix cut from origin/main, PR against main.
- [x] Discovery-surface proof: cluster; the k8scratch runs above.
- [x] Re-fix mechanism: N/A, no prior fix of this symptom exists.
- [x] Full affected validation: helm lint (both value sets) and all 48 chart CI scripts green.

### DIFF
--- a/.github/workflows/helm-ci.yaml
+++ b/.github/workflows/helm-ci.yaml
@@ -316,6 +316,21 @@ jobs:
           set -euo pipefail
           bash charts/curie/ci/langfuse-worker-probe-assertions.sh
 
+      # Prove the probe-window invariant across the WHOLE render, not a named
+      # list: for every container carrying both probes, the liveness failure
+      # cutoff -- initialDelaySeconds + (failureThreshold - 1) * periodSeconds,
+      # kubelet defaults applied to omitted keys -- must be strictly LATER than
+      # the readiness one, or the kubelet restarts a container that is still
+      # inside the boot window readiness was sized to tolerate. postgres (WAL
+      # replay), langfuse-web and langfuse-worker (Prisma/ClickHouse boot
+      # migrations) and api all inverted the two probes. Also requires an
+      # explicit livenessProbe timeoutSeconds everywhere, since the kubelet's
+      # invisible 1s default is the silent half of the postgres case.
+      - name: helm render assertions (probe windows)
+        run: |
+          set -euo pipefail
+          bash charts/curie/ci/probe-window-assertions.sh
+
       # Prove the controller NetworkPolicy RBAC split (issue #350): exactly one
       # read-only cluster grant (get/list/watch) so the informer syncs, no
       # cluster-wide mutate anywhere (#66 guarded), namespaced mutate intact, and
--- a/charts/curie/CLAUDE.md
+++ b/charts/curie/CLAUDE.md
@@ -42,34 +42,37 @@ component and rail detail in `charts/curie/README.md`.
     strategy values key**. Pinned by `ci/langfuse-recreate-assertions.sh`.
     Horizontal scale for Langfuse needs an Accepted out-of-band migrator; when
     that lands, this exception is removed rather than extended.
-  - **Third named exception: `langfuse-worker`'s probe cadences, also in
-    `templates/langfuse.yaml`.** Its `readinessProbe`/`livenessProbe` cadence
-    numbers (#2330) -- `initialDelaySeconds`, `periodSeconds`,
-    `timeoutSeconds`, `failureThreshold` -- are hardcoded to match
-    `langfuse-web`'s, three blocks above in the same file, instead of reading
-    from new `langfuse.worker.readinessProbe.*` / `.livenessProbe.*` values
-    keys. The probe types and endpoints deliberately differ from web's
-    (worker readiness is `httpGet /api/ready`, liveness is `tcpSocket`; web
-    uses `httpGet /api/public/health` for both -- see the liveness rationale
-    below) -- only the cadence numbers follow web. This is a knowing
-    deviation, not a claim that the invariant does not apply: this ticket
-    exists *because* `langfuse-web` and `langfuse-worker` had diverged (web
-    alone had the Postgres readiness gate), and adding values-driven probe
-    keys for the worker while `langfuse-web`'s probes stay hardcoded in the
-    same file would create a new asymmetry in the very file the ticket is
-    about. It is also not a new hole in the chart: every other long-running
-    Deployment already
-    hardcodes its probe numbers -- `api.yaml`, `ui.yaml`, `postgres.yaml`,
-    `clickhouse.yaml`, `valkey.yaml`, `rustfs.yaml`, `otel-collector.yaml`,
-    `mail-adapter.yaml`, `inference.yaml`, and the `curie.heartbeatProbes`
-    helper (used by `worker.yaml` and `dispatcher.yaml`) -- so `langfuse-web`
-    itself was already inside this exception before #2330. The only K8s Probes
-    genuinely driven by values in this chart are
-    `agentSandbox.runner.readinessProbe`, because its cadence couples to the
-    worker's claim-timeout budget, and `dispatcher.startupProbe`; nothing here
-    couples the same way. The correct end state is lifting `langfuse-web` and
-    `langfuse-worker` onto values-driven probes together, tracked as a separate
-    follow-up -- not extended piecemeal from this ticket.
+  - **Probe handlers stay in the template; probe cadences live in values.**
+    The probe *handler* -- `exec` / `httpGet` path / `tcpSocket` port -- is a
+    correctness choice about the image, not an environment knob, so it is
+    hardcoded in `templates/` on purpose (`langfuse-worker`'s `tcpSocket`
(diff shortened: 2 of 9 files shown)


TheConnMan:
The cargo audit failure is pre-existing on origin/main: the same check failed on the base commit 96b1a537 and the two main runs before it (RUSTSEC-2023-0071 in the rsa crate, plus unsound warnings in event-listener and lru). This branch touches no Rust; every other check is green or still running.

Issue #2519, opened 2026-09-09
A hosted connector whose Bearer placeholder expands empty still fails silently at the MCP capability probe


TheConnMan:
Follow-up to #2503. With the derived `Authorization: Bearer ${NAME}` header present, a WRONG token is user-visible (github-mcp-server returns the tool list and each call answers `GET https://api.github.com/user: 401 Bad credentials`), and a MISSING value is refused at `cluster deploy` (`connectors.yaml declares secret(s) with no value available ...`).

The remaining silent case: a secret in the `SecretRef` (`from_secret`) form never reaches the sandbox under ADR-0090, so the placeholder expands to an empty `Bearer ` and github-mcp-server answers 401 on `/mcp` itself. The runner's `probe_mcp_tool_capability` (`runner/src/curie_runner/mcp_tool_capability.py`) catches that and only logs `MCP tool-capability probe failed server=<name>`; the agent simply has no `mcp__<connector>__*` tools and answers from memory.

Ask: surface a failed capability probe for a declared connector to the message caller (or fail the boot loudly naming the connector and the credential), and decide how a SecretRef-form credential reaches the sandbox, if at all, without reimposing the deploy-identity access ADR-0090 removed.

Refs #2503, #2352, ADR-0090.

## Bonus Drain queue alignment (2026-09-09)

Task: `curie-v088-2519`. Group: **Curie v0.8.8**. Size: **large**. Priority: **2**.

Prerequisites: `curie-v088-spike-connectors`.

Execution scope: Implement only #2519 diagnostic slice after connector spike: declared-connector capability/auth failure must reach message caller or fail boot with actionable connector/credential-name diagnosis, including empty Bearer expansion. Never expose values. Keep SecretRef credential delivery and ADR-0090 authority unchanged; if that larger decision is required, retain it as a separate blocked follow-up. Coordinate #2352 to avoid two implementations of the same probe error path.

Reviewed diagnostics PR with actual failed-probe and healthy-connector caller-visible proof, negative controls and required tiers. Use Refs #2519, not Closes, while SecretRef-delivery decision remains unresolved; explicitly retain that residual in the issue.

Dependencies are enforced in the queue. A completed prerequisite job does not prove its PR is merged; verify required commits and evidence on the tested candidate. Work targets isolated environments and reviewed changes, with no implicit merge, release, permanent soak changes or immediate dispatch.

TheConnMan:
[...]

Issue #2503, opened 2026-09-09
Hosted HTTP MCP connectors never send Authorization, so github-mcp-server 401s and the agent has no tools


TheConnMan:
A hosted connector declared only in `connectors.yaml` is mounted into the sandbox on 0.8.6 (#1118's injection half), but the agent still has no `mcp__<connector>__*` tools when the image is `ghcr.io/github/github-mcp-server` over HTTP.

The GitHub MCP HTTP transport authenticates the *client*. `GET /mcp` without `Authorization` returns `401 Unauthorized` and `WWW-Authenticate: Bearer resource_metadata=".../.well-known/oauth-protected-resource/mcp"`. The same request with `Authorization: Bearer <PAT>` returns 200 and the tool list. GitHub's own HTTP docs require that header on the MCP client config.

Curie puts the PAT on the **connector pod** (`secrets:` -> `secretKeyRef`) and derives

```json
{"type": "http", "url": "http://<release>-<agent>-mcp-github.<ns>.svc.cluster.local:8000/mcp"}
```

with no headers. `plugin_format` rejects `headers:` on a hosted connector (`connectors.hosted_has_headers`). `cluster deploy --secret NAME` can put the PAT in the sandbox env, but hosted `mcp_entry` never references it, so a fresh thread still lists no `mcp__github__*` tools.

Found on shipped release-channel `curie 0.8.6` (`origin/main@0d3f6d4a`), cluster tier, live GitHub. Evidence is the 2026-09-09 stable-line usability retest of the #1118-blocked authed-MCP half.

This is not "never injected" (#1118). `connectors.yaml` is at `/bundles/current`, `CURIE_CONNECTOR_*` is set, and the runner logs `MCP tool-capability probe failed server=github`. The remaining gap is that hosted `mcp_entry` does not send the Bearer token the HTTP server requires.

## Reproduction

```yaml
# connectors.yaml -- the scaffolded hosted shape; .mcp.json has no github entry
connectors:
  github:
    image: ghcr.io/github/github-mcp-server:v1.12.0
    args: [http, --port, "8000", --read-only]
    secrets:
      - GITHUB_PERSONAL_ACCESS_TOKEN
```

```bash
export GITHUB_PERSONAL_ACCESS_TOKEN  # from the operator environment
curie cluster deploy --plugin-dir . --namespace "$NS" --release "$NS"
# connectors: applied 5 object(s)
# connector github: http://...-mcp-github....svc.cluster.local:8000/mcp
[...]

Issue #2352, opened 2026-09-04
Stable-line stopgap for #1118: cluster deploy should say loudly when a hosted connector is not mounted into the agent


TheConnMan:
On the stable line a `connectors.yaml` connector is hosted correctly and never reaches the agent (#1118), and as of 0.8.5 the workaround #1118 documents is refused by bundle validation. The result is that `curie cluster deploy` reports complete success, prints a connector URL, and ships an agent with none of the connector's tools — with no signal anywhere in the CLI that this happened.

This issue is the **stable-line stopgap** for that: make the failure visible while #1118 is fixed properly. #1118 remains the real fix.

## The gap being covered

Verified on shipped release-channel `curie 0.8.5` (`origin/main@39a57886`), cluster tier. Deploy reports success:

```
connectors: applied 5 object(s) for gh-connector-bot
connector github: http://...-gh-connector-bot-mcp-github....svc.cluster.local:8000/mcp
deployment    1892aac5-d23e-4f76-8055-c773241df9ae [dev] active
```

The connector is genuinely healthy — `1/1 Running`, `HTTP server listening addr=:8000`, a live Service endpoint, and both sandbox↔connector NetworkPolicies present. The agent still has nothing:

```
$ curie cluster message ... "Do you have any tool whose name starts with mcp__github? ..."
NO

I have the following mcp__ tools:
1. mcp__curie__request_approval
2. mcp__curie-state__append
...
None of these tools start with "mcp__github".
```

Asked to do the work, the agent degrades to answering from memory rather than reporting a broken tool surface, which is how this reaches an end user as a wrong answer rather than an error.

The escape hatch #1118 describes ("an author can copy it into `.mcp.json`") is now closed:

```
Error: uploading the bundle failed with 422 Unprocessable Entity:
{"code":"connectors.duplicate_server","message":"connectors.github: `github` is declared in both
connectors.yaml and the bundle's MCP config. Curie derives this server's URL from the Service it
creates, so remove the `github` entry from the MCP config and let connectors.yaml own it."}
```

So both configurations are dead ends: the sanctioned one mounts nothing, the workaround is rejected.

## Ask

Make `curie cluster deploy` say so. When a bundle declares hosted connectors, verify they are actually mounted into the agent's tool surface, and when they are not, emit a loud warning naming each connector that was hosted but not mounted — rather than printing its URL beside an `active` deployment.

[...]

PR #2544, opened 2026-09-09
Surface declared-connector MCP probe failures to the caller

TheConnMan:
## Summary

A declared connector whose `Authorization: Bearer ${NAME}` placeholder expands empty (or whose capability probe fails) no longer stays a log line. The runner diagnoses the connector and credential name, skips dialing `/mcp` with an empty Bearer, and short-circuits every turn before the model runs so the agent cannot answer from memory. The caller sees the diagnosis as the turn text. Secret values never appear in that text or in the new error log.

SecretRef credential delivery and ADR-0090 are unchanged. How a `from_secret` value reaches the sandbox, if at all, remains on #2519.

#2352 should reuse `tests/vectors/connector-probe-diagnosis.json` for expansion diagnosis and must not add a second HTTP MCP capability probe at deploy time. Deploy-time hosted readiness versus mounted declaration stays #2352.

Sibling: worker `kernel.py` is not edited. ErrorEvent plus a DONE final is the path that posts `Final.text` to Slack and `curie cluster message`. Classified-failure would be overwritten by the escalate copy.

## Related issue

Refs #2519
Refs #2503
Refs #2352

## Fix pin verification

Fix pin: runner/tests/test_connector_probe_diagnosis.py::test_empty_expansion_skips_http_probe_for_that_derived_server

## End-to-end verification

| skill | required | Runner ACI `/v1/event` is the diagnosis surface | fake | Commit `d259e7906`. In-process `create_app` POST `/v1/event` with empty-expansion failure: `FAIL_PROBE_OK declared connector 'github' failed MCP capability probe: credential GITHUB_TOKEN expanded empty. Connector tools are unavailable.` Fake model queries stayed `[]`. Healthy connector: `HEALTHY_OK queries= ['hi'] final_status= done`. Negative: derived server with no headers produced no failure (`NO_HEADER_NEGATIVE_OK`). |
| local | required | Worker DONE path posts `Final.text` | fake | Commit `d259e7906`. `TEST_VALKEY_PORT=36379 uv run pytest apps/worker/tests/kernel/test_kernel.py::test_connector_capability_failed_done_posts_diagnosis_not_escalate` passed. Observed: sink text is the diagnosis, no escalate, one attempt. Negative: existing `test_budget_exceeded_escalates_without_retry` still posts the human-flag copy for classified-failure. |
| local-release | n/a | No released-binary identity, install path, or release-compose pin change | | |
| cluster | n/a | Change does not reach chart templates, RBAC, sandbox claims, or init containers. Empty-expansion diagnosis runs inside the runner from headers plus env on every tier that mounts a derived server. | | |
| live provider | n/a | Does not change MCP catalog projection, tool mounting, model routing, or coding-tool capability. Diagnosis is local header-placeholder versus env inspection. | | |
| external integration | n/a | Does not change Slack, git webhooks, or connector OAuth. | | |

## Checklist

- Diagnostic slice only; SecretRef delivery left on #2519
- No frozen aci-protocol, plugin-format, schema, or ADR changes
- `kernel.py` untouched

### DIFF
--- a/apps/worker/tests/kernel/test_kernel.py
+++ b/apps/worker/tests/kernel/test_kernel.py
@@ -1182,6 +1182,34 @@ def test_budget_exceeded_escalates_without_retry(make_harness) -> None:
     asyncio.run(go())
 
 
+def test_connector_capability_failed_done_posts_diagnosis_not_escalate(
+    make_harness,
+) -> None:
+    # #2519: the runner short-circuits with ErrorEvent + DONE so the diagnosis
+    # reaches the message caller without a kernel.py change. Classified-failure
+    # would be overwritten by the escalate copy. This pins the DONE fork.
+    async def go() -> None:
+        async with make_harness() as h:
+            diagnosis = (
+                "declared connector 'github' failed MCP capability probe: "
+                "credential GITHUB_TOKEN expanded empty. Connector tools are unavailable."
+            )
+            h.runner.default_script = [
+                ErrorEvent(
+                    message=diagnosis, classification="connector-capability-failed"
+                ),
+                Final(text=diagnosis, status=DONE),
+            ]
+            await h.kernel.process_event(_qevent("go"))
+
+            assert h.runner.opened == ["go"]
+            assert h.sink.last_text == diagnosis
+            assert "Flagging for a human" not in diagnosis
+            assert "human" not in (h.sink.last_text or "").lower()
+
+    asyncio.run(go())
+
+
 def test_retries_are_bounded_then_escalate(make_harness) -> None:
     async def go() -> None:
         async with make_harness(max_attempts=3) as h:
--- a/runner/src/curie_runner/__main__.py
+++ b/runner/src/curie_runner/__main__.py
@@ -58,7 +58,12 @@ from .history import (
     resolve_history,
 )
 from .hooks import _merge_pre_tool_use_hooks, load_bundle_hooks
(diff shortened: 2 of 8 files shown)


TheConnMan:
Rebased onto current `origin/main` (no file conflicts; #2352's deploy-time hosted-but-undeclared warning is a separate path).

Python CI failed the fix-pin reverse check because reversing product files made `test_connector_probe_diagnosis.py` fail at collection (`ImportError: ConnectorCapabilityFailure`) instead of as an attributed test failure. The pin test now imports only `probe_mcp_tool_capability` at module level, so reverse yields `AssertionError: assert ['probed'] == []`.

TheConnMan:
Cluster e2e failed at `curie cluster up` with `Secret sh.helm.release.v1.curie.v1 is invalid: data: Too long: must have at most 1048576 bytes`. That is on current main (same failure on the post-merge main CI run), not this diagnostic slice: `charts/curie/ci/` was being packaged into the chart. `ci/` is now in `.helmignore`; packaged chart dropped from 518644 to 246032 bytes. Helm-ci assertion `helm-package-excludes-ci.sh` pins it.

TheConnMan:
Rebased onto current main. The two Helm-package commits are gone: #2550 already landed `ci/` in `.helmignore`. Cherry-picked only the diagnostic commit onto `origin/main` so it sits next to #2559's in-memory Bearer expansion. Probe diagnosis still runs on the unexpanded catalog in `_load_boot_fetches`, before materialize drops the env names.

Issue #2379, opened 2026-09-04
POST /channels/token never bumps generation, so a re-mint cannot be told from the token it replaces


TheConnMan:
## What happens

`POST /channels/token` returns a payload with `generation: 0` for every mint of the same binding. Three mints on 2026-09-04 for one email binding (7 d, 600 s, 7 d) all carried `generation: 0`, `scope: channel.enqueue`, the same `channel_id`.

Whatever `generation` was meant to do, rotation does not move it, so a re-mint cannot be told from the token it replaces, and the previous token stays valid until its own `exp`.

## Expected

Either bump `generation` on every mint and have the platform reject tokens below the binding's current generation (rotation revokes), or remove the field and document that only `exp` bounds a token.

## Acceptance

After a second mint for the same binding, the first token is rejected with 401 (or the field is gone and the docs say so).

PR #2560, opened 2026-09-10
Resolve the channel-token Secret before minting, not after

### LINKED ISSUE #2553: curie cluster channel-token revokes the installed token before it knows whether it can write the new one
`curie cluster channel-token` mints before it knows where to write the token.

`mint_channel_token` (`apps/api/src/curie_api/routers/channels.py:290`) is a
rotation write: it bumps `AgentChannel.generation` under `FOR UPDATE` and signs
the NEW value, so the token already installed in the adapter's Secret is
revoked the instant the mint commits. That is deliberate (#2379) -- the
generation is one of the only two things standing in for a revocation list.

The CLI verb does not respect that. In `cli/src/channel_token.rs` the mint runs
first, and only afterwards does the verb discover where the token has to go:

- `:183` mint (generation bumped, old token now dead)
- `:196` `fetch_release_computed_values` -- shells out to `helm get all`
- `:197` `live_token_secret` -- may shell out to `kubectl get deploy` to
  resolve the release fullname
- `:199-205` the `kubectl patch secret` that actually installs the new token

Every one of those steps can fail for reasons that have nothing to do with the
token: an expired kubeconfig, a Helm release the operator named wrong, a
`--namespace` typo, a transient API-server error, a `patch` refused because the
Secret does not exist. When any of them fails, the verb exits non-zero and the
install is left in a state it was not in before the command ran: the adapter is
still holding the previous token, that token is now revoked, and there is no
rollback. Inbound mail stops.

This is worse than the failure it is meant to fix. Recovery from an *expired*
token is the whole reason the verb exists; a failed recovery attempt that
revokes a still-working token converts a scheduled expiry into an outage, and
the operator gets an error about `helm` or `kubectl` with no indication that
their channel is now dead.

## Fix

Shrink the unrecoverable window to the one step that structurally cannot move,
and say so when it is hit.

1. Resolve the target Secret **before** the mint. `fetch_release_computed_values`
   and `live_token_secret` do not need the token; running them first means a
   discovery failure costs nothing.
2. The `patch` genuinely cannot precede the mint (it needs the token). When it
   fails, the error must say the previously installed token has been revoked
   and name the recovery.
3. A rollout failure *after* a successful patch is a different, milder state --
   the new token is installed, the adapter just has not picked it up -- and
   must not be reported as a revocation.

## Acceptance criteria

- [ ] `curie cluster channel-token` performs all Helm/kubectl discovery before
      it posts to `POST /channels/token`.
- [ ] A discovery failure (e.g. `helm get all` exits non-zero) aborts the verb
      with **no** mint request sent, so no generation is burned.
- [ ] A Secret-patch failure reports that the installed token was revoked by
      this mint and names re-running the verb as the recovery.
- [ ] A rollout failure after a successful patch reports the token as installed
      but not yet loaded, and does not claim a revocation.
- [ ] Regression tests pin the ordering and both failure messages; the ordering
      test fails against the pre-fix tree.

TheConnMan:
## Problem

`POST /channels/token` (`apps/api/src/curie_api/routers/channels.py:290`) is a
rotation write. It bumps the binding's `generation` under `FOR UPDATE` and signs
the NEW value, so the token the mail adapter is currently running is revoked the
instant that request returns. That is deliberate (#2379) -- the generation is one
of only two things standing in for a revocation list -- and there is no un-mint.

`curie cluster channel-token` did not respect it. The mint ran first, and only
afterwards did the verb work out where the token had to go:

- mint (generation bumped, installed token now dead)
- `fetch_release_computed_values` -- shells out to `helm get all`
- `live_token_secret` -- may shell out to `kubectl` to resolve the release fullname
- `kubectl patch secret` -- the write that actually installs the token

Every one of those fails for reasons that have nothing to do with the token: an
expired kubeconfig, a misnamed release, a `--namespace` typo, a transient
API-server error, a Secret the caller cannot patch. When one did, the verb exited
nonzero and left the install worse off than before it ran -- adapter still holding
the previous token, that token now revoked, no rollback -- and the operator read
an error about `helm` with nothing saying their channel had just gone dead.

Recovery from an expired token is the whole reason this verb exists. A failed
recovery attempt that revokes a still-working token converts a scheduled expiry
into an outage.

## Change

Shrink the unrecoverable window to the one step that structurally cannot move,
and describe the state honestly when it is hit.

- **Discovery moves ahead of the mint.** `fetch_release_computed_values` and
  `live_token_secret` do not need the token, so they now run while failing is
  still free. A discovery failure burns no generation.
- **A failed Secret write names what it left behind.** The patch genuinely cannot
  precede the mint. When it fails the error now says the mint already revoked the
  installed token and that re-running mints a fresh one, so retrying costs
  nothing.
- **A failed rollout does not claim a revocation.** Past the patch the new token
  IS installed and the adapter has merely not reloaded; that error points at
  `kubectl rollout restart` and explicitly says not to mint again -- a second
  mint there would revoke the token just installed.

Both messages reach both operator surfaces (the `--json` `{error, fix}` payload
and the terminal's `Error:`/`Fix:` pair) through the same `with_json_payload` +
`operator_context` pairing `Ui::failed_report` already uses.

- **An absent Helm release is a refusal.** The shared Helm read maps Helm's own
  `Error: release: not found` to `Ok(None)`, which is right for every other
  caller and wrong here: this verb patches a chart-rendered Secret and rolls a
  chart-owned Deployment. A `--namespace`/`--release` typo -- the case the issue
  names first -- therefore sailed past discovery, and since the mint is keyed by
  `kind:address` rather than by the release, it revoked a healthy adapter's token
  and then failed at a patch in the wrong namespace.
- **`token_exp` shares the write's message.** It sits between the mint and the
  write and used to raise a bare "retry", which is the one instruction that makes
  things worse there -- a second mint, with nothing said about the channel
  already being dead.

No API change. The rotation-at-mint semantics are correct; only the CLI's
ordering around them was not.

## Tests

`ClusterStub` gained `fail.<step>` sentinel files so a single stubbed step can be
made to exit nonzero, and the mock API appends its mint to the same `calls.log`
as the shell-outs so ordering can be asserted on one timeline. Six new tests in
`cli/tests/channel_token_verb.rs`:

- `a_discovery_failure_burns_no_generation` -- `helm` fails; the verb must abort
  with no `/channels/token` request recorded.
- `an_absent_release_refuses_before_minting` -- the typo case above.
- `discovery_runs_before_the_mint_on_the_happy_path` -- pins
  helm-read < mint < patch. (Its first version asserted only helm-before-patch,
  which held before the fix too; the review caught that.)
- `an_undecodable_token_reports_the_revoked_token`
- `a_failed_secret_write_reports_the_revoked_token`
- `a_failed_rollout_says_installed_not_revoked` -- including that the word
  "revoked" appears in neither its message nor its fix.

`curie dev verify-fix-pin` reports `PINNED` for the selector below. `cargo test`
over `channel_token_verb`, `mail_channel_health`, `json_contract`,
`cli_output_variants` and `command_surface`: 121 passed. `cargo fmt` and `cargo
clippy --all-targets` clean.

## Review

Cross-engine adversarial review (`agent-router adversarial-review --primary
claude`, routed to grok, review 439) found four real gaps in the first commit;
all are fixed in the second. Spec-vs-impl and scope-creep reviews pass.

One residual is deliberately **not** fixed here: `live_token_secret` ->
`release_fullname` (`cli/src/ops.rs:10043`) swallows a failed `kubectl get
deploy` and falls back to the chart-default Secret name, so a probe failure can
still mint against a guessed name. `release_fullname` is shared and cached
across many verbs and making it fail closed is a cross-cutting change with no
bearing on this ticket's ACs. Filed as #2561 against v0.9.0.

Closes #2553

Fix pin: cli/tests/channel_token_verb.rs::an_absent_release_refuses_before_minting

### DIFF
--- a/cli/src/channel_token.rs
+++ b/cli/src/channel_token.rs
@@ -179,6 +179,16 @@ pub async fn channel_token(opts: ChannelTokenOpts) -> Result<ChannelTokenOutput>
     }
     let ui = crate::ui::ui();
     let cl = ui.checklist();
+    // Resolve the Secret BEFORE minting. The mint is a rotation write: the API
+    // bumps the binding's generation and signs the new value, so the token the
+    // adapter is currently holding is revoked the moment that request returns
+    // (#2379). There is no un-mint. Every step that can fail for a reason
+    // unrelated to the token -- an expired kubeconfig, a misnamed release, a
+    // `--namespace` typo, an API-server blip -- therefore has to run while
+    // failing is still free, or a failed recovery attempt turns a scheduled
+    // expiry into an outage (#2553).
+    let values = fetch_release_computed_values(&opts.common).await?;
+    let (secret_name, secret_key) = live_token_secret(&opts.common, values.as_ref()).await;
     let mint_step = cl.step(&format!("minting channel token for {kind}:{address}"));
     let token = match client.mint_channel_token(&kind, &address, ttl_s).await {
         Ok(token) => {
@@ -192,30 +202,39 @@ pub async fn channel_token(opts: ChannelTokenOpts) -> Result<ChannelTokenOutput>
     };
     let exp = token_exp(&token)?;
     let expires_at = format_exp(exp);
-    let values = fetch_release_computed_values(&opts.common).await?;
-    let (secret_name, secret_key) = live_token_secret(&opts.common, values.as_ref()).await;
     let patch = serde_json::json!({ "stringData": { &secret_key: token } });
+    // The one step that structurally cannot precede the mint: it writes the
+    // token. If it fails the install IS worse off than before the command ran,
+    // and the operator has to be told that rather than left reading a bare
+    // kubectl error about a Secret.
     run_step(
         &cl,
         &format!("writing {secret_name}/{secret_key}"),
         "written",
         &patch_command(&opts.common, &secret_name, &secret_key, patch),
     )
-    .await?;
-    run_step(
-        &cl,
-        "rolling mail adapter",
-        "restarted",
-        &rollout_restart_command(&opts.common),
-    )
-    .await?;
-    run_step(
-        &cl,
-        "waiting for mail adapter",
-        "ready",
-        &rollout_status_command(&opts.common),
-    )
-    .await?;
+    .await
+    .map_err(|err| revoked_without_install(err, &opts, &kind, &address, &secret_name))?;
+    // Past the patch the new token is installed and the old one's revocation is
+    // no longer a surprise -- the adapter simply has not reloaded yet. A rollout
+    // failure is a restart away from resolved, so it must NOT be reported as a
+    // revocation.
+    for (label, ok_detail, cmd) in [
+        (
+            "rolling mail adapter",
+            "restarted",
+            rollout_restart_command(&opts.common),
+        ),
+        (
+            "waiting for mail adapter",
+            "ready",
+            rollout_status_command(&opts.common),
+        ),
+    ] {
+        run_step(&cl, label, ok_detail, &cmd)
+            .await
+            .map_err(|err| installed_but_not_loaded(err, &opts))?;
+    }
     Ok(ChannelTokenOutput::Minted {
         agent: agent.name,
         kind,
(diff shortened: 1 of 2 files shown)


TheConnMan:
Review note, carried from the spec-vs-impl pass: `live_token_secret` -> `release_fullname` (`cli/src/ops.rs:10043`) swallows a failed `kubectl get deploy` and falls back to the chart-default Secret name, so that one discovery path still does not abort pre-mint. It now runs before the mint like the rest, and the new patch-failure message covers the outcome, but a guessed Secret name can still cost a generation.

Left out of this PR deliberately: `release_fullname` is shared and cached across many verbs, and making it fail closed is a cross-cutting change with no bearing on this ticket's ACs. Filed as #2561 against v0.9.0 with three options and a recommendation.

Issue #2567, opened 2026-09-10
Inbound attachments are in ADR-0020's required core but nothing carries them


jw3329:
[ADR-0020](docs/adr/0020-message-port-rendering-free-channel-interface.md) is `Accepted` and puts attachments in the **required core** every adapter implements:

> Receive an inbound message (text, author, correlation key, **attachments**)

and `FILE_ATTACHMENTS` is one of the declared optional capabilities. The constant exists today at `packages/channel-protocol/src/channel_protocol/models.py:19`.

Nothing carries an attachment from an adapter to a runner.

| where | today |
|---|---|
| `packages/aci-protocol/src/aci_protocol/turn.py:120-143` | `QueuedTurn` is `event_id`, `conversation_id`, `author`, `text`, `reply_handle`, `received_at`, `source`. There is no file, attachment, id or url field |
| `apps/dispatcher/src/curie_dispatcher/handlers.py:355` | fills `text` from `derive_text` |
| `apps/dispatcher/src/curie_dispatcher/inbound_text.py:329-364` | `derive_text` returns a non-empty top-level `text` **unchanged** and consults `files` only when it is empty, emitting each entry's `title` or `name` and nothing else |
| any adapter | none declares `FILE_ATTACHMENTS` |

So a Slack message with a file attached **and any text at all** reaches an agent as the text alone. Not the bytes, not a url, not a file id, and not even the filename — because the sentence made `text` non-empty.

The capability name is in the tree and the plumbing that would carry it is not, which is the same shape as a port with no caller.

### Why it matters

An agent whose job involves a document a person sends it cannot be built at all, and the failure is quiet: the model receives a sentence describing a file and can answer confidently about a file it never received. Nothing in the stack reports that an attachment was dropped.

### What this needs, and why it is an issue rather than a PR

`runner/CLAUDE.md`:

> this environment shape is frozen with `aci-protocol`; a new required field is a contract change (**raise it in an issue/PR first**)

So this is raised before any change. A sketch of the shape, not a proposal to adopt as written:

- `QueuedTurn` gains an optional list of attachment references — an id and a name the adapter can later resolve, not bytes on the queue.
- The Slack adapter declares `FILE_ATTACHMENTS` and populates it from the event's `files`, independently of `derive_text`, which stays a text function.
- A capability the runner can be granted for resolving one, so an agent with no such grant sees no phantom tool.
- Absent attachments stay the ordinary case and change nothing for an adapter that declares no capability.

### Coordination

[...]

PR #2592, opened 2026-09-10
Carry inbound attachments and materialize them for the agent to read

jw3329:
## Why

A Slack message with a comment and an upload reached the worker with the comment intact and every file gone. Nothing reported the drop, so the model received a sentence describing a file and could answer confidently about a file it never received.

ADR-0020's required core names attachments alongside text, author and correlation key. This lands that half of the port and carries it through to something an agent can open.

## What it does

```
Slack event
  -> dispatcher records attachment REFERENCES (never bytes: its Slack client has a
     2s timeout against a ~3s ack budget, so downloading here would blow it)
  -> QueuedTurn.attachments   (optional, so PROTOCOL_VERSION 0.4.2 -> 0.4.3, a PATCH)
  -> worker fetches with the bot token, size-capped, parks in the private workspace
     store under its own prefix, mints a short-TTL signed reference
  -> the substrate redeems that reference, verifies the digest, and
     materializes the files: an attachments-init container on Kubernetes,
     the sandbox driver itself on Docker (no init containers there)
  -> the agent opens them as ordinary files under /attachments
```

Both substrates make the same refusals, mount at the same path, and take the same operator limits. That is deliberate and tested: the mount-path test reads the runner's own constant at runtime and cross-checks the chart value, so the two tiers cannot drift apart.

Design points worth a reviewer's attention:

- **No url on the wire.** A sandbox cannot fetch a channel URL in this release: ADR-0075's Agent Proxy is Accepted but unbuilt, and ADR-0032's release-wide CIDR egress is default-deny. A url field would be a promise the sandbox cannot keep and would invite exactly the fetch that policy refuses. The id stays adapter-scoped and the platform side resolves it.
- **No channel credential near the sandbox.** The download happens in the worker; the sandbox reaches only the object store, on the path it already has. `CURIE_ATTACHMENTS_REF` is redeemed by the substrate and kept off the runner container on both of them.
- **`/attachments` is a sibling of the workspace, not a child.** `workspace-init` clears its root on entry, so a restart would eat the files, and anything inside the checkout surfaces in `git status`, where a person's upload could ride into a publication diff.
- **Multi-file is all or nothing.** A partial set is indistinguishable to the agent from a complete one, so one failure refuses the resolve and deletes what it already uploaded.
- **The bytes have their own retention.** The workspace reaper is thread-ownership specific (keyed on `sha256(thread_key)`, TTL is the route TTL), so it is left alone; attachments get a sibling ledger with its own TTL, swept in the same orphan-reap tick under the same per-thread fence.
- **`derive_text` is byte-identical.** Its commit was "name every drop the adapter makes" and its docstring calls the non-empty-text passthrough a deliberate exception. Refs ride a parallel field rather than being folded into the prompt, so every enqueue that works today is unchanged.

## Operator action

The lane ships **off**: `worker.attachments.enabled` defaults to `false` and gates both halves, so an upgrade changes nothing until an operator turns it on. Turning it on requires the `files:read` bot scope, which means **an existing installation must reinstall the Slack app to the workspace** before attachments resolve, exactly as `channels:read` and `usergroups:read` required.

**This is not an upgrade blocker.** A missing `files:read` logs a WARNING at boot and names the scope in the preflight summary; it does **not** refuse to boot. An installation that upgrades without reinstalling boots normally and answers messages normally, and with the lane off it behaves exactly as it does today. Release notes should say "reinstall before switching attachments on", not "reinstall before upgrading".

## Verification

| Check | Result |
|---|---|
| affected suites (aci-protocol, dispatcher, worker, runner) | 2495 passed in the worker+runner sweep, 3 failed (2 pre-existing, 1 unrelated flake that passes in isolation) |
| full `uv run pytest` | 6083 passed |
| ruff / mypy / lint-imports | clean; 4 contracts kept, 0 broken |
| check-contracts / check-docs / wire-tolerance | clean |
| wire-lock base gate vs origin/main | ok, the patch bump is classified compatible |
| chart render-asserts | 56 of 57 (the one failure is a shell syntax error on main, see below) |
| cargo fmt / clippy -D warnings | clean |

The init container's refusals are proven by running the rendered program against a fixture HTTP server, not by reading it: a digest mismatch, a size overrun, an expired or non-HTTP(S) reference, a malformed digest, and a filename escaping the mount each exit non-zero. Neutering only the digest comparison makes that assertion fail with "the init container exited 0. In a pod that boots the runner and the agent silently reads whatever landed."

**What is and is not proven.** Everything from the queued turn down to the file on disk is covered on both substrates, including each refusal. The one hop still unproven is the live Slack download: it needs `files:read` granted on a real workspace, which is an operator step, and Slack is mocked in the tests. So the **external-integration tier is blocked, not passing**. Someone with a reinstalled app should upload a file and confirm the agent reads it before this is trusted in production.

Note that a fresh app installed from the updated manifest gets `files:read` on its first consent, with no reinstall friction, and `docs/slack-local-runbook.md` already recommends a separate Slack app per long-lived release. So this can be verified end to end without touching a production installation.

The 2 failures in the affected suites are `kernel/test_consumer_dead_letter` and `test_run::test_main_installs_dead_letter_alerting`. I ran the same command on a detached `origin/main` worktree and got the identical failure set, so they are pre-existing and not chased here. `test_connector_lock` also fails on main, from a `/etc` symlink on macOS. Most remaining failures in the full run are `ConnectionError`/`TimeoutError` against Langfuse and ClickHouse.

One note on how that was established, because the first run misled me: an earlier pass of the affected suites also failed three lease/TTL timing tests (`test_eval_consumer_publishes_and_renews_shared_liveness_lifecycle`, and two in `kernel/test_consumer.py`). They were not a regression. That run was contended by chart render-asserts and `cargo clippy` executing concurrently, which is why it took 11m09s against the baseline's 4m51s. Re-running the same command with nothing else on the box passes all three.

## Deliberately not in this PR

1. **Declaring `FILE_ATTACHMENTS` on the Slack adapter.** `ChannelCapabilities` and `FILE_ATTACHMENTS` have zero consumers repo-wide; nothing declares or queries a capability set anywhere. Declaring it now would add a second port with no caller, which is the defect this issue was filed about. The substantive requirement behind that bullet, an agent with no grant seeing no phantom tool, IS met: no reference in the claim means the init container materializes nothing and the runner's directory probe finds nothing.
2. **A person-facing refusal when a file exceeds the cap.** Today the error leaves the stream entry pending, so an oversize upload retries instead of telling the person. `CURIE_TURN_NOT_STARTED_TEXT` is the existing hook for it. Follow-up.
3. **Whether a resumed turn carries the attachments of the turn it continues.** `resumequeue._build_turn` replays the placeholder and endpoint, not the files; left at the empty default and commented in `cli/tests/resume_wait.rs`.
4. **`cargo test --test chart_check` cannot pass on a stock macOS dev box.** `charts/curie/ci/worker-ttl-bounds-assertions.sh` uses syntax that macOS's `/bin/bash` (3.2.57) cannot parse, so `chart_check_passes_on_the_unmodified_chart` fails locally there while passing in CI. Measured both ways: Linux `bash:5` parses it with exit 0, macOS bash 3.2 exits 2. Developer friction on a platform the release targets, not a broken gate. Follow-up.

   (An earlier revision of this description claimed that script was broken on `origin/main` and that `curie dev chart-check` fails there. That was my error -- I had run `bash -n` under bash 3.2 and read a local incompatibility as a repo defect. Corrected here so nobody chases it.)

Refs #2567

Because item 1 above declines one of the issue's four bullets on purpose, I have left the issue open rather than auto-closing it. Closing is a maintainer call.


TheConnMan:
Merged. I wanted this in 0.8.9 today, so rather than send it back I put two commits on your branch: `9ba624de` and `42b804b5`, plus `dc7ad1ff` to keep the boot-env declaration test happy. Writing up both so they are worth something beyond this PR.

**The design is right and I want to say so first.** References on the wire and never bytes; the download in the worker where the channel credential already lives; the capability redeemed by the substrate and kept off the runner; `/attachments` a sibling of the workspace rather than a child, with the `git status` reasoning spelled out. The agent opens ordinary files and nothing about them enters the prompt but the paths. That is exactly the shape I wanted, and `derive_text` being byte-identical is the part that proves you understood why. The verification section is the best I have read in this repo this month, particularly saying plainly that the live Slack download is blocked rather than passing, and the correction of your own earlier claim about the chart script on main.

**First fix: two uploads sharing a filename destroyed one of them.** The claim side is keyed by index and fine, but both substrates then wrote to `<mount>/<name>` — `os.replace` in the rendered init program, the same path join in the Docker driver — so two people attaching `report.pdf` to one message left the agent one file where two arrived, with nothing anywhere recording the loss. That is the same silent drop this issue exists to close, one layer further down. The lesson I would take: when a pipeline de-duplicates by index and then materializes by name, the index disappears at exactly the hop where it was doing the work, so the collision has to be re-answered there. `unique_attachment_leaf` now gives the first file the name the person sent and suffixes the rest before the extension, and the init program reimplements it because a rendered script cannot import the worker — with `ci/attachment-init-behavior-assertions.sh` executing that copy against the same inputs so the tiers cannot drift.

**Second fix: there was no way to turn the lane off.** `agentSandbox.runner.attachments.enabled` only stopped the sandbox half. With it false the worker still downloaded, parked bytes, wrote the retention ledger and minted a capability, and the k8s driver still emitted an Overrides entry naming an init container the chart had not rendered. For a feature whose last hop is explicitly unproven against a real workspace, the rollback path has to be one switch that turns off the whole lane, and it has to ship off. `worker.attachments.enabled` is that switch, defaulted false, gating both halves from one value; the sandbox key survives as a sub-gate that cannot turn the lane on. An install that takes 0.8.9 gets today's behaviour until an operator decides otherwise.

**One thing to correct rather than fix:** the body still said an existing installation must reinstall the Slack app before upgrading and that the preflight refuses the boot. Your own head commit had already made a missing `files:read` a warning. I edited that paragraph, because release notes get written from PR bodies and that one would have sent every operator through a reinstall they do not need.

Your three deliberate omissions all stand: the person-facing refusal text, declaring `FILE_ATTACHMENTS` with no consumer, and whether a resumed turn carries the files of the turn it continues. #2567 stays open, as you left it.

Issue #493, opened 2026-07-16
Chart: extract the hardening securityContext helpers and render-assert the sandbox lockdown


TheConnMan:
The sandbox pod template `charts/agentos/templates/agent-sandbox.yaml` writes the identical hardened securityContext block (no privilege escalation, read-only root filesystem, non-root, drop ALL capabilities) four separate times: both bundle init containers, the litellm sidecar, and the runner container. Three copies carry a comment promising they match the runner instead of factoring the block. This is the sandbox security rail; a partial edit desyncs a copy silently. Separately, the service templates repeat the same containerSecurityContext with/toYaml wrapper five times (worker, dispatcher, ui, api twice).

Extract one named hardening helper for the sandbox lockdown and one wrapper helper for the service securityContext include, switch all copies to them, and add a render assertion that every container in the sandbox template carries the hardened block.

AC:
- One helper renders the hardening block; all four sandbox containers include it; the inline copies are gone.
- The five service-template securityContext wrappers use a shared helper.
- A chart CI assertion fails if any sandbox container's securityContext deviates from the helper output.
- Rendered output before and after the refactor is identical (golden render diff, using the output-dir form rather than a stdout pipe).

Issue #492, opened 2026-07-16
Share the eval and approval wire DTOs and queue transport literals in aci-protocol


TheConnMan:
Three producer/consumer wire payloads are hand-written twice, once per lane, with no shared schema or drift gate, and two have already drifted:

- Eval stream job: `apps/api/.../evalqueue.py::EvalJobRequest` vs `apps/worker/.../eval/stream.py::EvalWorkItem`, identical seven fields, but bundle_ref is required at construction on the API side and optional on the worker side.
- Eval report: `apps/api/.../schemas.py::EvalReport` vs `apps/worker/.../eval/stream.py::EvalReport`, byte-identical classes on both ends of POST /evals/report.
- Approval request: `apps/worker/.../approvals.py::ApprovalRequest` vs `apps/api/.../schemas.py::ApprovalCreate`, matching field names but the API side carries validation constraints (min lengths, positive expiry) the worker copy lacks, so the worker can construct a payload the API rejects with 422, stranding the durable-approval path.

Alongside them, the queue transport literals are re-declared per lane: the payload envelope key in four Python modules plus `cli/src/queue.rs`; the runs-stream default in dispatcher config, worker config, and the CLI; the worker consumer-group name in worker config and the CLI, whose message verbs wait on that group's ack state, so a divergence hangs local and cluster message forever. The API additionally declares the runs stream twice itself (config field and a module constant).

The queued-turn model already proves the fix: one model in aci-protocol with stream/JSON helpers, consumed by every lane and exported to the generated Rust crate. Extract the three DTOs and the transport constants the same way; the API keeps its response-only models. Resolve the current field-optionality and constraint mismatches explicitly in the shared models (a documented choice per field, not silent loosening).

Coordination: #488 also changes aci-protocol; whichever PR merges second rebases and re-runs the contract regeneration before merge.

AC:
- One shared model each for the eval job, eval report, and approval request in aci-protocol with round-trip tests; API and worker import them; the per-lane classes are deleted.
- One declaration each for the stream name default, consumer group default, and payload envelope key in aci-protocol, exported to the generated Rust crate; all lanes import them and the API's duplicate internal declaration is collapsed.
- Grep for the former literals and class names finds only the shared declarations and env-override plumbing.
- PROTOCOL_VERSION bumped per the change-class table; schema and generated artifacts regenerated in the same change.

PR #2616, opened 2026-09-11
Summarize verification phase timing from existing run state

TheConnMan:
## Summary

Adds a lightweight, offline summarizer over the `e2e.evidence` records this repository's `/implement` run-state files already carry, so verification cost can be measured per phase instead of estimated. It reports startup, test execution, cleanup, retries and external wait separately, and a phase nobody measured stays unknown all the way to the output rather than becoming a confident zero. That is the point: conflating startup with test execution, or backfilling a plausible default, is what makes a savings claim unfalsifiable.

New `timings` block on an evidence entry is optional and backward compatible — every existing run-state file loads and reports all phases unknown. No runtime product code is touched; the tool is stdlib-only and reads only the files it is handed.

This is stage 1 of a three-stage maintainer-authorized pilot. **It is not independently merge-ready.** Full applicable repository verification is explicitly deferred to stage 3 against the assembled three-PR candidate. See "Verification debt" below.

## Related issue

No issue; this lands from an authorized pilot plan.

## End-to-end verification

This change does not alter runtime behavior because it adds a contributor-only offline analysis tool plus its tests and documentation, and one `testpaths` line. It reaches no compose service, dispatcher, worker or API wiring, no `curie` verb output or `--json` shape, no released-binary identity or install path, no chart template, RBAC or sandbox claim, no model routing or credential resolution, and no third-party API. It is deliberately not exposed as a `curie` subcommand, since adding a verb surface would pull in the local tier for a stage whose scope is explicitly lightweight.

Scoped verification, all against `54f8c82`:

- `uv run pytest tools/verification-timing/tests` — 48 passed.
- `uv run ruff check .` — all checks passed. `uv run mypy` — success, no issues in 253 source files.
- `./scripts/check-docs.sh` — OK; every command the new README names resolves.
- Real historical run-state files (no `timings` block anywhere) summarize with every phase `null` / `unknown_count` populated and `total_seconds` null — never `0`.
- Guard demonstration, executed rather than read: negative, non-numeric, non-finite and unrepresentable durations; a completion before its start; date-only timestamps; mixed timezone awareness; an unrecognised phase; a malformed `e2e` or `evidence` container; and a missing or blank `tier`/`command`/`commit` each exit 2 with empty stdout and an error naming the file and the offending field. Negative controls: an absent `e2e` key and an explicit `"evidence": []` still load as zero rows without error, and a valid timestamp pair still yields its measured duration.
- Determinism: the same record set supplied in reversed order produces a byte-identical summary (matching sha256).
- Privacy: absolute input paths and the evidence free-text fields are absent from the summary; error messages still carry the full path for diagnosis.
- Test strength: deleting the finiteness check fails 4 tests, so that assertion is not vacuous.

## Reviews

Two cross-engine adversarial reviews (read-only, reviewer excluded the authoring engine).

- Code review (#492) returned six findings — non-finite durations entering the statistics, a date-only pair becoming a measured `0.0`, mixed timezone awareness crashing instead of reporting, a malformed container silently yielding zero rows, absolute paths in the output, and empty-string identity mis-grouping unrelated candidates. All six were reproduced by execution, pinned with 15 new tests, fixed, and re-verified by execution.
- Consolidation review (#493) of the behavior-preserving refactor confirmed acceptance/rejection, unknown handling, retry attribution, timestamp precedence and output fields are all preserved. Its one noted difference — which error surfaces first when several fields are invalid at once — is non-blocking.

Four quality passes (reuse, simplification, efficiency, altitude) ran over the file; reuse and efficiency came back clean, and the simplification and altitude findings were applied in the final commit.

## Verification debt

Stages 1 and 2 of this pilot run focused test-first checks plus review and complete their handoff as draft PRs. The full applicable repository baseline and the union of required tiers are owed by **stage 3**, run against the assembled candidate containing all three diffs. This staging is a specific maintainer-authorized exception for this pilot and changes no repository-wide verification mandate; the tier rules in `AGENTS.md` remain in force and no fake result substitutes for a required real tier.

## Known limitation

Row and duplicate-group `source` values are the state file's bare name, so two identically-named run-state files from different worktrees are indistinguishable in that field. Candidate provenance itself is unaffected — duplicate identity keys on commit, tier and command, never on `source`.

## Checklist

- [x] Tests pass for the area I touched.
- [x] Docs updated if behavior changed.
- [ ] An ADR is added under `docs/adr/` if this is an architectural decision — n/a, no architectural decision.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -196,6 +196,7 @@ testpaths = [
     "tools/ci-release-target-parity/tests",
     "tools/restore-drill/tests",
     "tools/sre-contract/tests",
+    "tools/verification-timing/tests",
 ]
 
 [tool.ruff]
--- a/tools/verification-timing/README.md
+++ b/tools/verification-timing/README.md
@@ -61,10 +61,22 @@ over sparse data is honest about its own sample size: `observed_count` and
 `unknown_count` sit next to each other in every phase total, and a phase nobody
 measured reports `null` totals rather than a confident `0`.
 
-A malformed record — a negative duration, a non-numeric duration, a completion
-before its start, an unrecognised phase name — aborts the load naming the
-offending file and phase. It is never skipped silently, because a quietly
-dropped row is indistinguishable from a fast one.
+A malformed record aborts the load, naming the offending file and phase. It is
+never skipped silently, because a quietly dropped row is indistinguishable from
+a fast one. Rejected as malformed: a negative, non-numeric, non-finite
+(`NaN`/infinity) or unrepresentably large duration; a completion before its
+start; a date-only timestamp, which would make a guessed midnight look like a
+measured zero; a timestamp pair mixing a timezone-aware endpoint with a naive
+one, since the missing offset cannot be guessed; an unrecognised phase name; a
+corrupt `e2e` or `e2e.evidence` container, which would otherwise drop a whole
+file's rows from a combined summary; and an evidence entry missing any of
+`tier`, `command` or `commit`, whose absence would let unrelated candidates
+collapse into one duplicate group. An absent `e2e` key and an empty
+`"evidence": []` are not malformed — they simply contribute no rows.
+
+The `source` reported for each row and duplicate group is the state file's bare
+name, never its local absolute path; error messages keep the full path, because
+a diagnosis needs it.
 
 ## Running it
 
--- a/tools/verification-timing/README.md
+++ b/tools/verification-timing/README.md
@@ -1,0 +1,107 @@
+# verification-timing
+
+A lightweight, offline summary of how much wall-clock time the verification
+phases of an `/implement` run actually consumed. It reads the run-state files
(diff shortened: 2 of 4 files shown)


TheConnMan:
Closing unmerged per maintainer decision (2026-09-11): this stage is measurement/evidence scaffolding with no runtime surface, so it does not stand on its own as a PR. The branch `task/interface-pilot-1-measure` stays pushed at `54f8c825`; its commits will be folded into the stage 3 verification PR, where the timing evidence is actually consumed. No work is lost.

TheConnMan:
Reopened. Maintainer decision (2026-09-11) supersedes the earlier close: keep all three pilot PRs stacked, and review/decide on the assembled stack as a whole at stage 3 rather than on any single PR. This PR remains a draft and is not independently merge-ready.

TheConnMan:
Closing in favor of #2702.

PR #2623, opened 2026-09-11
Compose the real approval journey as a development-only test

TheConnMan:
Stage 2 of the three-PR interface pilot stack. **Stacked on `task/interface-pilot-1-measure` (PR #2616), not on `next`.** Draft, and **not independently merge-ready** — Brian reviews and decides on the assembled stack as a whole at stage 3.

## What this adds

One approval driven end to end across three apps, with no scripted stand-in anywhere between them:

real `approval_card(allow_free_text=True)` render → real Bolt `SocketModeHandler` → real ownership probe → real `views_open` note modal → real `view_submission` → real `ApprovalResolveClient` over loopback HTTP → the real `curie_api` app **with its lifespan entered** → real principal verifier + real authorizer → real Postgres → real `ResumeQueue.enqueue` → a real Valkey stream → the real worker `Consumer` and `Kernel`.

No test in this repo previously drove a Slack interactivity payload through the dispatcher handlers into the API's resume enqueue and on into a live worker kernel. The existing lifecycle test fakes the approval API at the `ApprovalCreator` seam and hand-builds resume turns.

**Ships zero production code.** `git diff --name-only` against the base matches nothing under any `src/`.

## Three findings from review worth calling out

- **Production posts *note* cards.** `kernel.py:3816` passes `allow_free_text=True` unconditionally, so the immediate `APPROVE_ACTION_ID` buttons are the older migration path. Testing only those would have made "the real approval journey" a false claim, so the note path is the journey under test and the immediate render is kept as an explicitly labelled variant.
- **A route-name denylist was the wrong control.** It collided with the real production route `POST /approvals/principals/operator`, and it only defends against surfaces named the way we would name them. The exclusion test instead freezes an inventory off the live app — exactly one mint handler and one resolve handler, asserted by object — so a new `POST /approvals/principals/chat` fails it by construction.
- **The first version of the sweeper/reconciler guard did not bind.** It set `CURIE_`-prefixed names that `Settings` (no env prefix, `extra="ignore"`) silently drops. Fixed to the real names, plus an assertion on the *resolved* settings so a future rename cannot quietly re-break it.

## Why the assertions can actually fail

Four mutations of real production code, each run by the driver and each caught:

| Mutation | Tests failing |
|---|---|
| `kernel.py:3816` note flag flipped | 1 |
| Authorizer always allows | 16 |
| `ResumeQueue.enqueue` never xadds | 3 |
| Verifier ignores signature tampering | 3 |

Chart assertions were mutation-tested separately (13 injected violations, each failing).

Every refusal pins *why* it failed: an authorizer refusal asserts the audit row and its authorizer name and that it is **not** `UnboundRouteBinding`; each forged principal re-sends the identical envelope correctly signed to prove it would otherwise succeed; the replay loser asserts the 409 body names the winner.

## Honest labelling

The runner is the existing in-process `FakeRunner` and Slack is the existing socket/`WebClient` fakes, so this proves **API, worker and dispatcher behavior only** — not real-runner or real-Slack behavior. The module docstring says so and also notes that "the card settled" is a `chat_update` call on a mock, and that ledger entries are fake-runner frames. Stage 3 connects the real runner.

## Verification

Run against real isolated Postgres, Valkey and RustFS in a private compose project. `CI_REQUIRE_VALKEY_TESTS=1` is required: without it 13 of 27 dispatcher approval tests silently skip, which is how a green run can mean nothing here.

- 33 new tests pass (17 journey + 16 exclusion); 36 compose tests; chart assertions OK
- Regression baselines green with no skips: dispatcher approval 27, API approval principals 13, action ledger, prod-gate
- Full kernel suite 480 passed / 2 skipped (both pre-existing opt-in long-run harnesses) in random order
- `ruff check` clean, `mypy` clean on 253 source files, docs gate OK

Reviews: adversarial plan review (13 blocking findings, all addressed), then code, scope and security reviews via `agent-router` on an independent engine. Scope review returned no unjustified creep. Security review found no harness reachability into production.

## Deferred to stage 3

Reject/expiry terminal outcomes, resume-enqueue failure and reconciliation, the note-modal *failure* half (validation, cancellation), the real runner, `SlackUserGroupMembers`, operator/console as journeys, and **the full assembled repository baseline and union of required tiers** — owed against the assembled three-PR candidate.

Known limits, stated rather than papered over: the concurrent-resolve overlap check skips (rather than fails) if the scheduler serialises it, so it never claims concurrency it did not observe; the live-app issuance inventory is static and package-scoped, so a mint reached by dynamic dispatch would evade it; and bearer replay inside the 60s TTL and `dependency_overrides` remain undefended.

## Observation, not a finding

The generated release compose inherits `${CURIE_APPROVAL_CHAT_ATTESTER_SECRET:-curie-dev-approval-chat-attester}` from `compose.dev.yaml`. This is the established dev-compose posture — it sits alongside `curie-dev-key` and `curie-dev-worker-token`, and is in fact stricter than both since it is overridable. Production is gated by `_refuse_dev_defaults_in_prod` under `ENVIRONMENT=prod` and by the chart's generated managed secret, and the sealed chart render is asserted clean. Flagging for awareness only; no change made here.

### DIFF
--- a/apps/api/tests/test_approval_dev_surface_exclusion.py
+++ b/apps/api/tests/test_approval_dev_surface_exclusion.py
@@ -1,0 +1,734 @@
+"""No development-only approval capability ships, and none is reachable in prod.
+
+The capability defended here is *"resolve an approval without a genuine chat
+principal"*, not "a route named /dev exists".  A route-name denylist is
+deliberately absent: ``POST /approvals/principals/operator`` is a real
+production issuance surface whose path contains "principal", so a denylist
+either fails on the real route or is amended until it no longer names the
+actual issuance surface.  Either way it is not the control, and it only ever
+defends against a surface named the way we would have named it.  Every
+assertion below is therefore positive and behavioral, driven against the real
+app in ordinary production wiring with a real approval row.
+
+The one structural assertion is the surface inventory: the live router table is
+asked which handlers can reach ``approval_principal.mint`` or
+``crud.claim_approval_resolution`` at all -- by object identity, not by path --
+so a NEW issuance or resolution surface fails loudly instead of going unnoticed
+by a suite that only exercises the surfaces it already knows about.
+
+Bypasses this set does NOT test, stated plainly rather than claimed complete:
+- **Replay of a valid chat bearer inside its 60s TTL.** Inherent to a bearer with no
+  nonce; untested here and not defended by this PR.
+- **In-process `dependency_overrides`** on the app object (already used by
+  `apps/api/tests/test_approvals.py:2117-2128`). Anyone with in-process access to the app
+  can bypass everything; that is not a boundary this test can defend.
+- **A stolen valid operator token used against an `ExplicitUsers` route** -- that is the
+  intended operator capability (#1531), not a bypass, but it means "platform key holder
+  cannot resolve" is **false in general** and this plan does not claim it. The claim is
+  narrowed to: *a platform key holder cannot resolve a channel-bound approval.*
+- **A handler that reaches mint/claim through a non-``curie_api`` indirection** (a
+  third-party dispatcher, a dynamically resolved attribute). The inventory walk
+  follows names to objects through this package only.
+- **``approval_auth``'s ``compare_digest(attester, api_key)`` equal-secret guard.** It is
+  unreachable through ``Settings``, which refuses equal keys at construction; the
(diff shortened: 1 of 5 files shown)


TheConnMan:
Closing in favor of #2702.

PR #2627, opened 2026-09-11
Make the approval journey drivable and prove the harness cannot ship

TheConnMan:
Stage 3 of the three-PR approval pilot (`.projects/local-readiness/2026-09-10-interface-test-plan/review.md`, item 3). Stacked on #2623, which is stacked on #2616. **Do not merge this PR on its own** — per the maintainer decision of 2026-09-11 the three PRs are reviewed and decided as one stack.

## What this adds

**A reusable, agent-facing interaction harness** (`packages/test-support/src/curie_test_support/interaction/`) exposing `send`, `messages`, `act`, `await_outcome`, `reset` and `inject_fault`, each returning a frozen JSON-serialisable result, plus a `python -m curie_test_support.interaction` NDJSON CLI. Every blocking call is bounded by an explicit deadline and raises `HarnessTimeout` naming the verb and the segment — including nested HTTP reads and every Valkey command. `act` accepts only an action captured off a rendered card, and over the CLI only via an unforgeable per-action handle minted at capture.

**The real runner with a fake model.** The harness builds `create_app(build_runner(..., fake_model=True))` and passes it through the existing `kernel_harness(runner_app=)` seam, so the runner state machine is genuinely in the path for the harness verbs, not only in one test.

**The failure half stage 2 deferred:** rejected terminal outcome; expiry driven by invoking the production sweeper once; a lost resume-enqueue healed by the production reconciler; modal cancel; tampered `private_metadata` refused rather than declined; and a failed resolve transport surfaced with no phantom resume.

**Production-exclusion proof** for the whole capability: an instruction-level Dockerfile posture check across six production recipes, the dependency closure each image installs, the single dev-group declaration, the release compose, the sealed chart render, and a prod-mode probe that must be refused and must leave approval state unchanged.

## A stage 2 defect this fixes

`test_approval_journey_dev.py` pinned `TEST_VALKEY_PORT` to the literal `36379` — stage 2's own private stack — so the module failed on CI and on every machine but the one that wrote it. The guard now asserts the property it wanted (frozen Valkey constant agrees with the environment, `VALKEY_URL` absent, `CI_REQUIRE_VALKEY_TESTS` set, run namespace unique and empty), with two tests proving the guard can still fail.

## Known gap, stated rather than claimed

**There is no server-side note validation in the product.** The modal's `max_length=2000` is client-side only, `resolve_note_submission` neither validates nor truncates, and `ApprovalResolve.note` is unconstrained. The plan's "note-modal validation failure" case is therefore carried as a *characterisation* test of current behaviour plus an `UNMET PROOF` note, not as a covered case. Closing it is a production change this stage does not own.

## Tier classification (assembled diff `9e3d8b62` → this head)

No production `src/` file is changed; the diff is test, dev-test-package and CI-assertion code only. All six tiers not applicable: **skill** — no skill surface or bundle; **local** — no compose service, image or `curie` verb behaviour (the compose edit is an added assertion); **local-release** — no released-binary surface and deliberately no new `curie` verb; **cluster** — no chart template, the chart edits are `ci/*.sh` render assertions; **live-provider** — `fake_model=True`, no provider credential or model call; **external-integration** — Slack stays faked at the edge.

Deliberately no `curie` subcommand: a CLI surface would pull the local and local-release tiers into a stage whose diff otherwise bears no runtime behaviour. Stage 1 recorded the same decision for the same reason.

## Review

Four independent adversarial reviews via `agent-router` (reviewer: grok; primary claude excluded), three rounds of fixes. Findings files are under `.projects/plans/`. Notable catches, all fixed: the bounded-wait claim was false (a predicate's 10s HTTP read outran a 1s deadline); `act`'s provenance rule accepted a hand-built action, and later accepted a guessable one over the CLI; `send` double-produced one turn; the expiry/reconciler "production object" assertions were import tautologies; the new chart script was not wired into helm-ci while `chart_check.rs` requires a 1:1 match, so that gate was red; and four separate ways to get the harness into a shipped `/app/.venv` that every exclusion control passed.

## CI

**This PR reports no checks.** `ci.yaml` triggers only on `pull_request: branches: [main, next]`, and this PR targets `task/interface-pilot-2-approval`. That is structural, not a bypass or a failure. Do not read the absence of red as green — the local assembled baseline in the handoff is the evidence for this stage.

### DIFF
--- a/.github/workflows/helm-ci.yaml
+++ b/.github/workflows/helm-ci.yaml
@@ -39,6 +39,19 @@ on:
       # (#2190) both read compose.dev.yaml to check third-party image tags,
       # so a compose-only edit that unpins one must run this workflow too.
       - "compose.dev.yaml"
+      # The dev-harness exclusion gate EXECUTES every production image recipe
+      # and the runner pin exporter. A PR that adds the harness to one of them
+      # touches no charts/ path, so without these the gate that exists to catch
+      # exactly that is skipped. `adapters/discord/Dockerfile` is a production
+      # recipe of the same shape (COPY . . then a --package uv sync) and was
+      # missing here, so a Dockerfile-only path-install there ran no gate.
+      - "runner/Dockerfile"
+      - "runner/export_dependency_pins.py"
+      - "apps/api/Dockerfile"
+      - "apps/dispatcher/Dockerfile"
+      - "apps/worker/Dockerfile"
+      - "apps/mail-adapter/Dockerfile"
+      - "adapters/discord/Dockerfile"
 
 permissions:
   contents: read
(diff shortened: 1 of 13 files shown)


TheConnMan:
Closing in favor of #2702.

